PR #29, opened 2016-08-05
Remove coverage testing from the scaffold.

DanielBrookRoberge:
Coverage testing breaks tests on some React components. For the moment, then, we need to remove it.

## Changes
- Remove the coverage packages and integration.
- Switch CircleCI testing to just use the plain tests

## How to test-drive this PR
- `npm prune`
- Run `npm test` and `npm run lint` and see that they pass
- See that the build passes on CircleCI with test results.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,9 +8,6 @@
     "transform-object-assign"
   ],
   "env": {
-    "test": {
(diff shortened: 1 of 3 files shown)


mikenikles:
👍

PR #32, opened 2016-08-10
Update devDeps to better match SDK

marlowpayne:
## Changes
- New release of SDK means dep version bump
- devDeps in scaffold were a bit mismatched when compared against SDK (for packages used by both)
  - Currently placed importance on the SDK versions

## How to test-drive this PR
- Checkout code
- Refresh node modules with `rm -rf node_modules && npm i`
- Test and check things still work


DanielBrookRoberge:
Works for me here :+1:

PR #35, opened 2016-08-10
Add tasks to upload and save-credentials

MikeKlemarewski:
Add the necessary npm scripts to upload to cloud.

 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/83

## Changes
- Add npm scripts to save API credentials and upload to cloud
- Change output of loader.js to be called adaptive.min.js so that cloud accepts it

## How to test-drive this PR
- Point the progressive-web-sdk dependency in package.json to the branch in https://github.com/mobify/progressive-web-sdk/pull/83
- `npm install` the progressive-web-sdk
- Get your API key from https://cloud.mobify.com/account/
- Run `npm run save-credentials -- -u youremail@domain.com -k your-api-key`
- Run `npm run push -- -m "upload message"`
- Visit the [progressive web project](https://cloud.mobify.com/projects/progressive-web-scaffold/publishing/production/) in cloud and make sure your bundle is there, and that the preview link works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "minimist": "1.2.0",
     "mobify-code-style": "2.5.3",
     "node-sass": "3.8.0",
-    "progressive-web-sdk": "0.0.5",
+    "progressive-web-sdk": "0.0.6",
     "prompt": "1.0.0",
     "raw-loader": "0.5.1",
     "react-addons-test-utils": "15.2.1",
--- a/package.json
+++ b/package.json
@@ -75,11 +75,13 @@
     "build": "webpack --config webpack/production.js -p --display-error-details --optimize-dedupe",
     "dev": "node ./dev-server/index.js",
     "lint": "eslint '**/*.{js,jsx}'",
-    "sasslint": "sass-lint -c node_modules/mobify-code-style/css/.sass-lint.yml 'app/**/*.scss' -v",
     "lint:circleci": "npm run lint && npm run sasslint",
+    "push": "cross-env NODE_ENV=production npm run build && sdk-upload",
+    "sasslint": "sass-lint -c node_modules/mobify-code-style/css/.sass-lint.yml 'app/**/*.scss' -v",
+    "styleguide": "styleguidist server",
+    "save-credentials": "sdk-save-credentials",
     "test": "ava",
-    "test:watch": "npm test -- --watch",
-    "styleguide": "styleguidist server"
+    "test:watch": "npm test -- --watch"
   },
   "ava": {
     "require": [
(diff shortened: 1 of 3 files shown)


webpack/base.loader.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     entry: './app/loader.js',
     output: {
         path: path.resolve(process.cwd(), 'build'),
-        filename: 'loader.js'
+        filename: 'adaptive.min.js'
```
jansepar: Hmm - I actually updated the Lancome tags to use `loader.js`. We will need a way to override this on a per-project basis.

In order for this to work properly with `loader.js`, we will just need to override the `filename` field on Cloud for that project.

MikeKlemarewski: Think we want to leave out this change then, and make that a task when a user sets up a progressive site? Changing the name in pbjt that is...

MikeKlemarewski: After talking with Shawn, we decided to revert this change. If a customer puts the tag in, and it's pointing to adaptive.min.js, we'll just have to manually update this for that one project. If we keep running into this issue, it will forces us to update cloud sooner.

package.json
```diff
@@ -75,11 +75,13 @@
     "build": "webpack --config webpack/production.js -p --display-error-details --optimize-dedupe",
     "dev": "node ./dev-server/index.js",
     "lint": "eslint '**/*.{js,jsx}'",
-    "sasslint": "sass-lint -c node_modules/mobify-code-style/css/.sass-lint.yml 'app/**/*.scss' -v",
     "lint:circleci": "npm run lint && npm run sasslint",
+    "push": "cross-env NODE_ENV=production npm run build && sdk-upload",
+    "sasslint": "sass-lint -c node_modules/mobify-code-style/css/.sass-lint.yml 'app/**/*.scss' -v",
+    "styleguide": "styleguidist server",
+    "save-credentials": "sdk-save-credentials",
```
johnboxall: @MikeKlemarewski not a huge deal... but it would be cool at some point to take a look at the whole "Save Credentials" model we've been using... and maybe also see if we could support environment variables ... `MOBIFY_API_KEY=xyz mobify push`

MikeKlemarewski: I feel like that would be fairly easy to support. My only question would be: How do environment variables behave on windows?

I'm sure Node has some cross-platform ways of getting/setting env vars.

mikenikles: https://www.npmjs.com/package/cross-env is the most common tool I'm aware of for this situation.

mikenikles:
👍 Tested on @jnurse's machine and it worked as expected.

PR #36, opened 2016-08-11
Use jQuery passed from jqueryResponse function

MikeKlemarewski:
Sometimes jQuery would load slow. If our component tried to import jQuery before it loaded, an exception would be thrown becuase webpack builds a module that looks like this:

```
module.exports = jQuery;
```

That throws an exception if jQuery isn't defined. If we pull it from window, jQuery will just be undefined until it loads.

## Changes
- Grab jQuery from window

## TODO
- [ ] Release a new version of the SDK and bump the version in this PR

## How to test-drive this PR
- Download the scaffold
- Point the SDK to develop and `npm i && npm run dev`
- In Charles, enable SSH proxying for your local ip address
- Then open up `throttle settings` and throttle `https://<your-ip>:8443/static/js/jquery.min.js`
- Set the bandwidth to something ridiculous like 1kbps and the roundtrip time to 15000 ms
- Now preview merlins potions against your local dev server
- The page should load, and the title on the page shouldn't appear until jQuery loads, but it should eventually load.


webpack/base.main.js
```diff
@@ -28,7 +25,7 @@ module.exports = {
     plugins: [
         new webpack.ProvidePlugin({
             React: 'react',
-            $: 'jquery'
+            fetch: 'imports?this=>global!exports?global.fetch!whatwg-fetch'
```
MikeKlemarewski: I'm not sure why this got added in here....

app/containers/home/actions.js
```diff
@@ -8,8 +8,8 @@ export const fetchHomeContents = () => {
     return (dispatch) => {
         fetch('/')
             .then((response) => jqueryResponse(response))
-            .then(($responseText) => {
-                dispatch(receiveHomeContents(homeParser($responseText)))
+            .then(([$, $loool]) => {
+                dispatch(receiveHomeContents(homeParser($, $loool)))
```
DanielBrookRoberge: Did you mean to include this variable rename?

webpack/base.main.js
DanielBrookRoberge: We don't need this with `whatwg-fetch` in the entry section, which is the polyfill developers' preferred method.

DanielBrookRoberge:
This looks like it works, though your test drive instructions don't work now that we're now loading jQuery from our CDN.

app/containers/home/actions.js
MikeKlemarewski: hah, nope :P

DanielBrookRoberge:
:+1: let's get this out

PR #44, opened 2016-08-12
Add Themes.scss

jeffkamo:
Compile and add `themes.scss` to the scaffold.

## Changes
- Add example theme file
- Add a central stylesheet for themes to be pulled in
- Tweak Webpack configuration to extract two separate CSS files: one for the main stylesheet, one for the themes.

## How to test-drive this PR
- Add some styles to `app/components/example/theme.scss`, like `body { background: red; }`
- Run `npm run dev`
- [Preview](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Confirm that the theme styles you added apply


mlworks:
+1 theme styles applied correctly

app/main.jsx
```diff
@@ -14,9 +14,10 @@ import App from './containers/app/container'
 import Home from './containers/home/container'
 import PLP from './containers/plp/container'
 
-// Stylesheet: importing it here compiles the SCSS into CSS. The CSS is actually
-// added to the markup in `loader.js`
+// Stylesheets: importing it here compiles the SCSS into CSS. The CSS is
+// actually added to the markup in `loader.js`
 import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
+import Themes from './themes.scss' // eslint-disable-line no-unused-vars
```
MikeKlemarewski: Instead of adding the eslist-disable to every line here, you can add it to the top of the file like this: https://github.com/mobify/progressive-web-sdk/blob/develop/scripts/build.js#L3

jeffkamo: Hmm, fair enough. However, at this point it's a temporary issue, since I'll eventually be converting our CSS to our old/traditional way. That will will including deleting this line at that time.

MikeKlemarewski:
After playing around with this, I don't think there is any benefit of having two separate css files. We can just import all the theme styles at the bottom of the main stylesheet. As long as the stylesheet import comes after all of the component imports in our main.jsx entry point, then the themes will be written at the end of the stylesheet, and properly override the component styles.

jeffkamo:
@MikeKlemarewski thanks for catching that. I've made the tweak so theme styles are imported into the main SCSS file now.

styleguide.config.js
```diff
@@ -30,11 +30,15 @@ module.exports = {
...
+                include: [
+                    /node_modules/,
+                    /app/
+                ]
             },
             {
                 test: /\.svg$/,
-                loader: 'text'
+                loader: 'text',
+                include: /node_modules/
```
MikeKlemarewski: Are we loading svg's from our sdk? If so, I think we want to be very specific with the path we give it: `node_modules/progressive-web-sdk`

webpack/base.main.js
```diff
@@ -65,6 +65,10 @@ module.exports = {
             {
                 test: /\.scss$/,
                 loader: ExtractTextPlugin.extract(['css', 'sass']),
+                include: [
+                    /node_modules/,
```
MikeKlemarewski: Same here, I think we want to only include the node_modules that we need to, not the entire folder.

styleguide.config.js
jeffkamo: done

webpack/base.main.js
jeffkamo: done

MikeKlemarewski:
To finish this off, I think it would be best to have an example of adding a theme style to an SDK component. So, maybe bringing in a button to the home page, then adding a theme.scss file for it.

Then we could remove the example component that was added in this PR. I'd rather not add code that we know people will remove. The button theme may actually be used on a project.

PR #45, opened 2016-08-16
Use SDK router

DanielBrookRoberge:
The SDK will now wrap the entire react-router API and export the useful parts from there. This changes the scaffold to use it and removes the direct dependencies on `react-router` and `react-router-scroll`.

 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/93 (merged), https://github.com/mobify/progressive-web-sdk/pull/95

## Changes
- Convert router components to the SDK versions
- Remove now-unneeded deps

## How to test-drive this PR
- Connect to a version of the SDK with the above fixes.
- Build and preview Merlin's Potions
- Ensure that the page renders and the links work.


app/main.jsx
```diff
@@ -18,15 +16,15 @@ import PLP from './containers/plp/container'
 // added to the markup in `loader.js`
 import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
 
+import {Router, Route, IndexRoute} from 'progressive-web-sdk/dist/routing'
+
```
MikeKlemarewski: It probably won't matter with these components, but in general, any components who import their own styles will have their styles applied at the end of the stylesheet **if** they are imported after the main stylesheet. We generally don't want this to happen because we will import extra theme styles in the stylesheet.

MikeKlemarewski:
@DanielBrookRoberge I'm a 👍 , as long as we move those router imports above the stylesheet import

PR #51, opened 2016-08-18
Revert to old SCSS methodology

jeffkamo:
Tweak the scaffold to be compatible with the changes from [here](https://github.com/mobify/progressive-web-sdk/pull/108)

 **Linked PRs**: [progressive-web-sdk#108](https://github.com/mobify/progressive-web-sdk/pull/108)

## Changes
- Make all the SCSS import into a single monolithic `stylesheet.scss`
- No more relying on `import styles from 'style.scss'`
- Remove `/override` and `/vendor`, but keep their SCSS files and update with examples

## Todos
- [x] Merge [progressive-web-sdk#108](https://github.com/mobify/progressive-web-sdk/pull/108)
- [x] Update to the latest SDK version
- [x] Update any dependencies to point to the correct files (i.e. base SCSS files in the SDK)

## How to test-drive this PR
- `npm i`
- From the SDK styleguide, pick a component and copy and paste it's import JS code into a container file, such as `app/containers/home/container.js`
- Add one of the usage examples of the component to the container.
- From the SDK styleguide, pick a component and copy and paste it's import SCSS code into the scaffold's `app/styles/_components.scss` file, under the "SDK Components' Main Styles" heading
- `npm run dev`
- [Preview the scaffold](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.min.js&disabled=0&domain=&scope=1)
- Confirm that the component displays as intended in the container.


app/styles/_components.scss
```diff
@@ -0,0 +1,19 @@
+// Components
+// ===
+
+// SDK Components' Main Styles
+// ---
+
+@import 'node_modules/progressive-web-sdk/dist/components/button/button';
```
fractaltheory: With the latest changes in https://github.com/mobify/progressive-web-sdk/pull/108 the base file is now called `base` instead of the component name, right?

jeffkamo: that's correct. I will fix this.

jeffkamo: Oh, well... that'll break the branch as it is, since it's not using the updated version of the SDK yet, which hasn't had these files renamed.

fractaltheory: Gotcha. We'll have to update that separately once the SDK releases with your changes.

fractaltheory:
👍

PR #52, opened 2016-08-18
Rename z-index variables

jeffkamo:
Due to popular demand, rename the Z-Index variables into something easier to use.

## Changes
- Rename `z-index` away from the theatre stage analogy, and more toward a numbered unit.

## How to test-drive this PR
- Confirm that you feel like this is a good change.


mlworks:
I like it +1

yourpalsonja:
I like it too!

PR #53, opened 2016-08-18
Update utilities to work with Spline 2.2.0

cole-sanderson:
Update utilities to work with Spline 2.2.0

## Changes
- Add `u-text-capitalize` 
- Add `u-width-full-block`
- Update `u-width-full`

## How to test-drive this PR
- Apply `u-text-capitalize` class and should lowercase all text in string. 
- Apply `u-width-full-block` class and should display full width. 
- Apply `u-width-full` class and should display full width.


jeffkamo:
👍

PR #54, opened 2016-08-18
Fall back to desktop if the provided regexes don't match the path.

DanielBrookRoberge:
Since automatically extracting the path regexes from the router to the loader is taking a long time, this PR provides a means for controlling the React app loading based on manually-entered regexes. 

## Changes
- Add a basic `isReactRoute` function to the loader
- Fall back to desktop when the function returns false

## How to test-drive this PR
- Preview Merlin's Potions
- Observe that the React app continues to work correctly loading from the home page.
- Observe the desktop site at http://www.merlinspotions.com/books.html
- See that the React site reappears on navigating to the Potions section.

### DIFF
--- a/app/loader.js
+++ b/app/loader.js
@@ -3,6 +3,10 @@ import {displayPreloader} from 'progressive-web-sdk/dist/preloader'
 
 window.Progressive = {}
 
+// THESE ARE EXAMPLES
+// They work for Merlin's Potions as currently set up.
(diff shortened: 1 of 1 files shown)


cdegit:
Seems to be working well on C&E!

fractaltheory:
👍  apart from that SDK versioning problem which doesn't really block this.

PR #56, opened 2016-08-19
Split out an AppProvider component from main.jsx

DanielBrookRoberge:
The current builds of Crabtree and Lancome have a split in their root-level scripts between `main.jsx` which contains mainly boilerplate for react-hot-loader and Redux, and `app-provider.jsx` which contains the actual Provider and Router components. 

This is very useful for automatic route regex extraction, since it will need to actually render the Router in a different environment. With the splitting, the routes can be imported for this purpose.

## Changes
- Split the `main.jsx` into `main.jsx` and `app-provider.jsx`, along the lines of the Crabtree and Lancome builds
- Explicitly import `React` in all files with JSX, so that it can find it outside Webpack.

## How to test-drive this PR
- npm run dev
- see that Merlin's Potions previews well.


fractaltheory:
If this change makes the explicit import of `React` inside component files a requirement, do you think we should remove it from our Webpack config's `ProvidePlugin` setting?

mikenikles:
👍 For explicit imports vs. webpack magic!

samplefrequency:
:+1:

DanielBrookRoberge:
Thanks people! I'll merge after removing React from the webpack plugin.

PR #57, opened 2016-08-20
Add Nightwatch boilerplate

marlowpayne:
Add boilerplate code for writing end-to-end Nightwatch tests

## Changes
- Adds an es5 Nightwatch config file to hook on to Babel so we can write our tests in ES2015
- Adds a standard `nightwatch.json` config which leverages our existing `nightwatch-commands` & selenium runner
- Adds a `site.js` config (es5 for backwards compatibility with `nightwatch-commands`' `.preview()` method) which is auto-populated on project generation
- Adds a sample `home` page object and associated example workflow test
- Adds `npm run nightwatch` command to kick off the party

## How to test-drive this PR
- Clone the Progressive generator [here](https://github.com/mobify/generator-progressive-web)
- Hop into the generator's `generate.sh` file and change line 5 to `SCAFFOLD_VERSION_OR_BRANCH="add-nightwatch"`
- Run the generator in the directory where your test project will be generated: `./path/to/generator/generate.sh`
- Accept the license and provide a test project slug
- Enter `http://www.merlinspotions.com` as the project URL (the URL must include the "http" protocol part for `.preview()`)
- Once the new testing project is generated, navigate into the directory
- There's a [bug with setting the site URL](https://github.com/mobify/generator-progressive-web/issues/2) in the generator, so run the command manually: 

`egrep -lR "siteUrl" . | tr '\n' '\0' | xargs -0 -n1 sed -i '' "s/\"siteUrl\": \"\"/\"siteUrl\": \"http:\/\/www.merlinspotions.com\"/g" 2>/dev/null` 
- Start up the dev server with `npm run dev`
- Once the dev server is up and running, in a new window/tab navigate to the project directory and run `npm run nightwatch`
- 💥 🎉


mikenikles:
👍 :shipit: 
Works as expected, thanks a lot for that contribution, Marlow!
image

tests/system/nightwatch.json
```diff
@@ -0,0 +1,53 @@
+{
```
ellenmobify: We've mostly switched to putting all the `nightwatch.json` config inside `nightwatch-config.js` instead. See https://github.com/mobify/carnival-cruises-mobile/blob/develop/tests/system/nightwatch-config.js Can we do that here?

marlowpayne: @ellenmobify I did try that, however I encountered issues with it regarding use of ES2015 in the test files. The JSON version didn't seem to have these issues, so I went with that instead.

PR #62, opened 2016-08-23
Use cache manifest

fractaltheory:
**JIRA**: [WEB-726](https://mobify.atlassian.net/browse/WEB-726)
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/128

## Changes
- Update `.gitignore` to ignore `tmp/` directory
- Adds required, user-editable `config/cache-hash-config.json` configuration file as well as some sane defaults (for the scaffold, anyway)
- Updates loader and main webpack configuration to provide an alias for `cacheHashManifest`
- Update `dev` and `build` npm scripts to call cache manifest binary before starting dev server or building app files, respectively

## Todos
- [x] Add a high-level description of your changes to the CHANGELOG.md.
- [x] Investigate whether it's worthwhile to remove `loader.js` dependencies on `progressive-web-sdk/asset-utils` since it will now include the hash manifest (and possibly bloat file size)

## How to test-drive this PR
- Check out [progressive-web-sdk#use-cache-manifest](https://github.com/mobify/progressive-web-sdk/pull/128) and run `npm link`
- Check out this branch
- run `npm run dev`

### Verify:
- `tmp/cache-hash-manifest.json` was created and lists `static/js/jquery.min.js`
- `tmp/loader-cache-hash-manifest.json` was created and has an empty `hashes` object and a buildDate string
- project builds and can be previewed against `http://www.merlinspotions.com`
- rendered HTML applies cachebreakers to `jquery.min.js`, `main.js` and `main.css` that match manifest `buildDate` entry
- modifying `hashLength` in `app/config/cache-hash-config.json` modifies generated hashes

### Bonus:
- Add `import {getAssetUrl} from 'progressive-web-sdk/src/asset-utils'` to a container file
- Add a new image to the project somewhere in `static/`
- Reference that image using `getAssetUrl` e.g. ``
- Run `npm run dev`
- Rendered HTML shows the image you imported into the project, with a md5 hash applied to the src path that matches what's in the cache manifest


fractaltheory:
One big thing I'd like to discuss is whether we should modify the project `loader.js` to not import asset-utils, for things like `loadAsset` and `getAssetUrl`. By bringing in this module, we also bring in the hash manifest, which can grow large on big projects. One example of what we did was simply copy pasting those relevant functions into `loader.js` -- not ideal but certainly one approach: https://github.com/mobify/roadrunner-sports-mobile/commit/f3c2126da201a22c6c4d427810f13e90f1d35312#diff-85044fad63ed16d573b29b4da020c90b

DanielBrookRoberge:
Normally I'm a little fanatical about DRY, but I think that copying some of the utility functions into the loader to avoid pulling in the manifest is a good idea. On webpush we definitely duplicated some functionality in our loader so that it could be as small as possible.

DanielBrookRoberge:
Or, we could have a `loader-utils` module in the SDK that the `asset-utils` could wrap with additional functionality.

DanielBrookRoberge:
:+1: on the functionality of this, though.

MikeKlemarewski:
Since jquery has an entry in the hashes object, shouldn't `getAssetUrl` use the hash for the jquery entry rather than the build date?

fractaltheory:
So, this is something I was talking to Shawn about, when we figured two separate files would be the best option. One negative though is that it could be confusing, since you might expect entries from one file. 

In this case, since jquery.min.js is loaded by the loader, it uses the loader cache manifest. Which only has the build date. That's why you see build date instead.

MikeKlemarewski:
That all makes sense. 👍

PR #64, opened 2016-08-23
Automatically generate loader route regexes

DanielBrookRoberge:
This completes the automation of isReactRoute

 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/100

## Changes
- Modify the loader to import the regexes from `loader-routes.js`. 
- Add a placeholder `loader-routes.js` so that things will always compile
- Add an NPM script for running the route extraction. (This is a slow process that we probably can't do every build)

## How to test-drive this PR
- Install an SDK on the linked branch
- npm run dev and preview
- see that it loads the React app on all routes (/potions.html and /books.html, for example)
- npm run update-loader-routes (this will produce a lot of irrelevant warnings)
- see that it only loads the React app on the correct routes, and not missing routes

### DIFF
--- a/app/loader-routes.js
+++ b/app/loader-routes.js
@@ -1,0 +1,3 @@
+// GENERATED FILE DO NOT EDIT
+const routes = [/.*/]
+export default routes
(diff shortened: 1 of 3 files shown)


fractaltheory:
This works pretty good 👍

PR #74, opened 2016-08-31
Move changelog note to correct location

jeffkamo:
Hotfix: A changelog item got added to the wrong spot. Moving it to the correct spot here.

## Changes
- Move changelog item to correct spot ("To Be Release")


MikeKlemarewski:
👍

PR #75, opened 2016-09-01
Use getasseturl from sdk

mikenikles:
Upgrade to version `0.3.0` of the Progressive Web SDK.

## Changes
- Add a `Logo` component to test the `getAssetUrl` function.
- Remove the `cacheHashManifest` webpack alias.
- Use the new `initCacheManifest` function to initialize the cache manifest at app startup.

## How to test-drive this PR
- Make sure `npm run test:all` passes
- Start the dev server with `npm run dev` and preview Merlin's Potions


app/components/logo/index.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import {getAssetUrl} from 'progressive-web-sdk/dist/asset-utils'
+import Image from 'progressive-web-sdk/dist/components/image'
+
+class Logo extends React.Component {
+    constructor(props) {
+        super(props)
+
+        this.logoURL = getAssetUrl('static/img/logo.png')
```
fractaltheory: This usage (declaring the url in the constructor) isn't strictly necessary now that we have the manifest.

app/loader.js
```diff
@@ -1,5 +1,6 @@
-import {getAssetUrl, loadAsset} from 'progressive-web-sdk/dist/asset-utils'
+import {getAssetUrl, loadAsset, initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
 import {displayPreloader} from 'progressive-web-sdk/dist/preloader'
+import cacheHashManifest from '../tmp/cacheHashManifest'
```
fractaltheory: This should be `../tmp/loader-cache-hash-manifest.json`

app/main.jsx
```diff
@@ -1,4 +1,6 @@
 import {polyfill} from 'es6-promise'
+import {initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
+import cacheHashManifest from '../tmp/cacheHashManifest'
```
fractaltheory: This should be `../tmp/cache-hash-manifest.json`

app/components/logo/index.jsx
mikenikles: I get a lint error, @fractaltheory:

image

app/main.jsx
```diff
@@ -1,4 +1,6 @@
 import {polyfill} from 'es6-promise'
+import {initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
+import cacheHashManifest from '../tmp/cache-hash-manifest.json'
```
MikeKlemarewski: We should probably put the cache-hash-manifest.json in a different folder. It seems very odd that we import something from a `tmp/` folder

fractaltheory:
It would make sense for it to be in `build/`. But everything in `build/` is uploaded to Cloud, and we don't need to upload these files. Since these are only used once to be included in our build files, I figured it was reasonable to be in `tmp/`. Also, that's the behavior from Adaptive :P

app/main.jsx
fractaltheory: It would make sense for it to be in `build/`. But everything in `build/` is uploaded to Cloud, and we don't need to upload these files. Since these are only used once to be included in our build files, I figured it was reasonable to be in `tmp/`. Also, that's the behavior from Adaptive :P

MikeKlemarewski: @fractaltheory ok, I'm cool with that. Is there anything else we need to do to get this merged?

fractaltheory: Don't think so.

PR #76, opened 2016-09-07
Add login form using FormFields component

MikeKlemarewski:
**JIRA**: https://mobify.atlassian.net/browse/WEB-723
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/183

## Changes
- Uses the FormFields component to build the form.

## How to test-drive this PR
- `npm run dev`
- Browse to the login page
- Make sure the login form works, and has proper validation


app/containers/login/container.js
```diff
@@ -10,29 +12,41 @@ import * as loginActions from './actions'
...
     }
 
     render() {
         return (
             <div>
                 <Link href="/">Go Home</Link>
-                <LoginForm onSubmit={this.onSubmitLoginForm.bind(this)} />
+                <LoginForm
+                    formFieldDescriptor={this.props.loginForm.fields}
+                    onSubmit={this.onSubmitLoginForm.bind(this)} />
```
fractaltheory: I think React recommends you bind class methods in the constructor.

fractaltheory:
There's a lint error:

screen shot 2016-09-16 at 1 34 06 pm

and I get these prop type warnings in the console when I preview the app:

screen shot 2016-09-16 at 1 30 32 pm

app/components/login-form/auth.js
```diff
@@ -0,0 +1,32 @@
+import {SubmissionError} from 'redux-form'
+
+const loginUrl = 'http://www.merlinspotions.com/customer/account/loginPost/'
+
+export const attemptLogin = (formData) => {
+    const postBody = ''
+        .concat(`form_key=${formData.form_key}`)
```
fractaltheory: Is this the preferred method for dealing with form submissions? I know we've previously placed the form markup in the page, or in memory, used jQuery to modify the fields, and then serialized it for POST.

app/components/login-form/index.jsx
```diff
@@ -1,46 +1,76 @@
...
+import {connect} from 'react-redux'
 import {Field as ReduxFormField, reduxForm} from 'redux-form'
 
-import Field from 'progressive-web-sdk/dist/components/field'
+import classNames from 'classnames'
+
+import FormFields from 'progressive-web-sdk/dist/components/form-fields'
+import SkeletonBlock from 'progressive-web-sdk/dist/components/skeleton-block'
 
 const validate = (values) => {
```
fractaltheory: There might be some value in commenting this a bit, or maybe linking to the relevant redux-form doc?

app/components/login-form/index.jsx
```diff
@@ -1,46 +1,76 @@
...
-            <ReduxFormField name="password" label="Password" component={Field}>
-                <input type="password" />
-            </ReduxFormField>
-            <button type="submit" disabled={pristine || submitting || invalid}>Login</button>
+        <form onSubmit={handleSubmit} className={classes}>
+            {formFields ?
+                <FormFields items={formFields} />
+            :
+                <div>
+                    <SkeletonBlock className="short" width="20%" height="14px" />
```
fractaltheory: What's the effort look like having the FormFields component provide built-in skeletons?

app/components/login-form/index.jsx
```diff
@@ -1,46 +1,76 @@
...
-    submitting: PropTypes.bool,
+    submitting: PropTypes.bool
 }
 
-LoginForm = reduxForm({
-    form: 'login',
-    validate
-})(LoginForm)
+LoginForm = connect(
+    (state) => {
```
fractaltheory: Is this a `mapStateToProps` argument? Maybe we could stick to the convention of declaring that as a variable.

app/containers/login/reducer.test.js
```diff
@@ -16,11 +15,3 @@ test('unknown action type leaves state unchanged', () => {
 
     expect(reducer(inputState, action)).toBe(inputState)
 })
-
-test('reducer implements all defined actions', () => {
```
fractaltheory: Why did you need to remove this test?

app/components/login-form/index.jsx
MikeKlemarewski: I purposely didn't want to do this. Form fields will likely be used when the developer needs to parse the Desktop form to get the form structure. That means making a network request once the page mounts.

To make a convincing skeleton, we need to know what the form is going to look like. We can't know that until the network request for the desktop markup gets back, which is too long.

app/components/login-form/index.jsx
MikeKlemarewski: Good call

app/components/login-form/auth.js
MikeKlemarewski: I didn't really try any other methods. I can give jquery serialization a shot. Ben C had issues on Crabtree getting form submission to work, and this was his end approach.

app/containers/login/reducer.test.js
MikeKlemarewski: It expected the reducer to have a case for all actions. In `login/actions.js` we have an action for `fetchLoginContents` which dispatches the `receiveLoginContents` action. There doesn't need to be a `fetchLoginContents` case in the reducer, because the store doesn't get updates by it, so I nuked the test as I didn't think it was helpful.

fractaltheory:
👍

app/containers/login/parsers/login.js
```diff
@@ -0,0 +1,32 @@
+// import parseInput from 'progressive-web-sdk/dist/parsing/input-parser'
```
fractaltheory: Hm, can remove this I think right?

MikeKlemarewski: Yes...that's for a subsequent PR

PR #77, opened 2016-09-08
Script loading example

fractaltheory:
Currently, there is no standardized or easy way to parse the original document, extract information about third-party scripts such as their src attribute or innerHTML, and insert those scripts into the new document.

This PR introduces LabJS as well as a helper method, for both specifying search terms and injecting scripts into the document.

 **JIRA**: [WEB-815](https://mobify.atlassian.net/browse/WEB-815)
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/193

## Changes
- Adds example of searching for a few scripts contained in the desktop website

## How to test-drive this PR
- First, follow the directions in https://github.com/mobify/progressive-web-sdk/pull/193
- `npm link progressive-web-sdk`
- `npm run dev`

## Verify
- A `requirejs/require.js` script inserted into `<head>`
- A network request for `collect?` initiated by `analytics.js` in your Network panel


app/containers/home/actions.js
```diff
@@ -1,15 +1,43 @@
...
+                    // This only needs to be done once in the application
+                    .init($)
+                    /**
+                     * url: current url you want to extract scripts from
+                     * $response: jQuery response document
+                     * {searchTypes}: string or RegExp search terms
+                     *   `contains` - inline scripts
+                     *   `src` - external scripts
+                     */
+                    .inject(url, $response, {
```
jansepar: A bit confused about this line - why do you need to supply it with both a `url` and a `$response`? I feel like it's either one or the other but maybe I'm missing something!

fractaltheory: People seem to use `$response` in their projects to name the jQuery-wrapped HTML that is returned by `jqueryResponse`, but it's not really a `response` object (i.e. we don't have access to anything that relates to the original URL). So, maybe I want to change that variable name.

As for why we need the URL, it's part of an optimization to prevent having to re-parse the HTML for scripts if we've already parsed it once. But, there's a comment on the SDK PR regarding cases where this will prevent new scripts from being extracted/injected so I need to fix that.

fractaltheory:
Hm, CircleCI is going to choke on the tests until we can point to a version of the SDK with the necessary scripts...

PR #78, opened 2016-09-12
Take out webpack dashboard

MikeKlemarewski:
## Changes
- Remove webpack dashboard from the webpack dev.js config

## How to test-drive this PR
- checkout this branch
- `npm run dev`
- Ensure the console outputs the original webpack build information, not the webpack dashboard.


mikenikles:
I think there's also a setting in the dev server called "silent". Let's remove that so we get to see the output in the console.

PR #79, opened 2016-09-12
Run get-routes on dev

MikeKlemarewski:
Previously we had to manually run `sdk-get-routes` to update the loader-routes. This ensures that it runs when the dev server starts so that we don't have to remember.

## Changes
- Runs the sdk-get-routes script when the dev script starts up
- Updates the loader routes

## How to test-drive this PR
- Update one of the routes in the `app/app-provider.jsx` file
- `npm run dev`
- Kill the dev server and ensure that the loader-routes file changed to reflect your updated route.

### DIFF
--- a/app/loader-routes.js
+++ b/app/loader-routes.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 // GENERATED FILE DO NOT EDIT
-const routes = [/.*/]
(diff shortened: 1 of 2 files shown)


fractaltheory:
@MikeKlemarewski should we also run this during the build step?

fractaltheory:
Apart from comment above, looks good 👍

MikeKlemarewski:
Yeah, added it to the build step as well

PR #80, opened 2016-09-13
Adds triggering page views by default to the scaffold using the latest analytics lib from the SDK

jansepar:
Adds triggering page views by default to the scaffold using the latest analytics lib from the SDK

 **JIRA**: https://mobify.atlassian.net/browse/WEB-723
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/195

## Changes
- Triggers page views when moving between new pages

## How to test-drive this PR
- Clone the progressive-web-sdk and checkout the `WEB-723-add-analytics` branch
- Run `npm link` in progressive-web-sdk and `npm link progressive-web-sdk` in the scaffold
- Navigate to different pages, ensure that the page view analytics requests look correct (I recommend using the GA chrome extension to ensure that's the case)
- Clicked on the themed component button to ensure events are firing correctly


MikeKlemarewski:
This doesn't seem to work with the SelectorRouter. It looks like the SelectorRoute's onEnter function never gets called. I'm looking into it a bit.

MikeKlemarewski:
Alright, I've got a PR that ensures `onEnter` gets called for SelectorRoutes: https://github.com/mobify/progressive-web-sdk/pull/200

app/containers/home/container.js
```diff
@@ -1,16 +1,27 @@
...
+        super(props, context)
+        this.triggerTapEvent = this.triggerTapEvent.bind(this)
+    }
+
     componentDidMount() {
         this.props.fetchHomeContents()
     }
 
+    triggerTapEvent() {
+        triggerMobifyEvent('ui', 'tap', 'sample component')
```
MikeKlemarewski: What do you think about passing an options object instead of individual arguments? I like options objects because they are more clear as to what each argument is, and you don't have to remember the correct argument order. Also, the `triggerMobifyEvent` function can then be extended to accept extra options without breaking the API:

```
triggerMobifyEvent({
    category: 'ui',
    eventAction: 'tap',
    eventLabel: 'sample component',
})
```

app/main.jsx
```diff
@@ -18,8 +18,12 @@ import AppProvider from './app-provider'
 // added to the markup in `loader.js`
 import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
 
+import {initMobifyAnalytics} from 'progressive-web-sdk/dist/analytics'
+
 polyfill()
 
+// TODO: replace slug with grabbing something from package.json
```
jansepar: This needs to happen before merging!

MikeKlemarewski: Fixed :)

MikeKlemarewski:
👍

PR #81, opened 2016-09-13
Update Scaffold for SDK 0.4.2

fractaltheory:
Updates the scaffold from 0.3.0 to 0.4.2, accounting for breaking changes in 0.4.0.

## Changes
- Removes AVA as testing infrastructure
- Replaces AVA with Jest, including configuration based on https://github.com/mobify/progressive-web-sdk/pull/182
- Adds Jest configuration files
- Updates existing boilerplate tests to work with Jest
- Adds `makeRequest` method to utils.js

## TODO
- ~~Add test coverage?~~
- [ ] Use `global.$` in parser test, or add eslint ignore comment?

## How to test-drive this PR
- `npm install`

## Verify
- `npm run dev` runs and you can preview http://www.merlinspotions.com
- `npm run test:all` runs tests and linting and it all passes


app/containers/home/parsers/home.test.js
```diff
@@ -1,10 +1,9 @@
...
 import {jquerifyHtmlFile} from 'progressive-web-sdk/dist/test-utils'
 import homeParser from './home'
 
-test('the title is parsed', (t) => {
-    const home = homeParser(window.$, jquerifyHtmlFile('home.test.html'))
-    t.truthy(home)
-    t.truthy(home.title)
-    t.is(home.title, 'Home page')
+test('the title is parsed', () => {
+    const home = homeParser(global.$, jquerifyHtmlFile('app/containers/home/parsers/home.test.html'))
```
fractaltheory: Unsure if having people need to use `global.$` is the best approach here.

package.json
```diff
@@ -89,25 +89,25 @@
...
     "test:all": "npm run lint && npm test",
     "test:watch": "npm test -- --watch",
     "update-loader-routes": "sdk-get-routes"
   },
-  "ava": {
-    "files": [
-      "**/*.test.js",
-      "**/test.js",
-      "!node_modules"
+  "jest": {
```
fractaltheory: Something I didn't add in this PR is test coverage. Is that something we want to add?

MikeKlemarewski: I'm going to say that SDK unit test coverage is more important that scaffold unit test coverage. I think we should get some systems tests for the scaffold eventually, and those would be more important.

mikenikles: I agree. We should document unit tests as a best practice and explain how coverage can be enabled, but shouldn't try to enforce it for all projects.

app/containers/home/parsers/home.test.js
MikeKlemarewski: Might be able to specify `jquery` as an environment in the .eslintrc.yml

http://eslint.org/docs/user-guide/configuring#specifying-environments

circle.yml
```diff
@@ -3,7 +3,8 @@ machine:
         version: 4.4.7
 test:
     override:
-        - npm run test:all
+        - npm run lint
+        - npm test -- --runInBand
```
MikeKlemarewski: do you know what `runInBand` does? My 2 minutes of googling failed me

MikeKlemarewski:
Tested this out and it's all working. Maybe try including that jQuery environment in the eslint config and see if that gets around the linting errors.

👍

app/containers/home/reducer.test.js
```diff
@@ -12,5 +11,5 @@ test('unknown action type leaves state unchanged', (t) => {
         item: false,
     })
 
-    t.is(reducer(inputState, action), inputState)
+    expect(reducer(inputState, action)).toBe(inputState)
```
mikenikles: @fractaltheory That should be `toEqual` since it's comparing objects.

fractaltheory: Good catch! https://github.com/mobify/progressive-web-scaffold/pull/81/commits/74aa8c8a6fbb33688d7d0e348c2ff224fba1122b

circle.yml
mikenikles: It runs tests in series vs. parallel.

app/containers/home/parsers/home.test.js
fractaltheory: Nice, that worked: https://github.com/mobify/progressive-web-scaffold/pull/81/commits/fdf611887b88b92e5c5ada24ae1344b70b3ff582

PR #89, opened 2016-09-17
Fix hotloader issues, refactor analytics in router

jansepar:
## Changes
- Use latest version of react-hot-loader to fix debugging issues
- Clean up router analytics code

### To test Analytics:
- Install the [GA debugger chrome extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en)
- Load a page, ensure a GA page view is triggered
- Navigate to a different page, ensure page view is triggered

### To test hotloader fix:
- Add `throw Error('test')` inside the `triggerPageView` function in main.jsx, ensure you can see the error in the browser


app/app-provider.jsx
```diff
@@ -9,18 +9,27 @@ import Home from './containers/home/container'
...
 
 const AppProvider = ({store}) => {
     return (
         <Provider store={store}>
             <Router>
-                <Route path="/" component={App}>
-                    <IndexRoute component={Home} routeName="home" onEnter={triggerPageView} />
-                    <Route component={Login} path="customer/account/login/" routeName="login" onEnter={triggerPageView} />
-                    <Route component={PLP} path="*.html" routeName="productListPage" onEnter={triggerPageView} />
+                <Route path="/" component={App} onEnter={handleEnter} onChange={handleChange}>
```
fractaltheory: Couldn't `onEnter` just call `triggerPageView` directly?

jansepar: Ah, yeah they totally could, when I was doing this in Crabtree they were doing other stuff in `onChange`, hence why I wanted a separate function for it. But yeah sounds good.

fractaltheory:
Seems to work! 👍

PR #90, opened 2016-09-18
✨ Add SDK components to the project styleguide

marlowpayne:
Adds the Progressive Web SDK components (as `src/`) to the project styleguide

## Changes
- Add SDK components to project styleguide
- Since SDK components now visible in project styleguide, update project's stylesheet to pull in their base and theme styles
- Clean up README, include styleguide mention

## How to test-drive this PR
- Checkout code
- `npm i`
- `npm run styleguide`
- Visit `http://localhost:4000` to see the new project styleguide
- Add a new project component with `npm run add:component`
- Verify the new project component appears in the project styleguide


marlowpayne:
The SDK `dist/` directory isn't being cleaned up correctly on releases, leading to ambiguous Sass imports for SDK component themes. This is fixed by https://github.com/mobify/progressive-web-sdk/pull/215. This PR will be blocked until this is resolved.

marlowpayne:
See it in action on an in-flight project with this Crabtree PR: https://github.com/mobify/crabtree-evelyn-progressive/pull/150

marlowpayne:
This will be unblocked as of the next SDK release.

app/styles/_components.scss
```diff
@@ -1,22 +1,73 @@
...
+@import 'node_modules/progressive-web-sdk/dist/components/pagination/base';
+@import 'node_modules/progressive-web-sdk/dist/components/rating/base';
+@import 'node_modules/progressive-web-sdk/dist/components/scroll-to/base';
+@import 'node_modules/progressive-web-sdk/dist/components/select/base';
+@import 'node_modules/progressive-web-sdk/dist/components/sheet/base';
 @import 'node_modules/progressive-web-sdk/dist/components/skeleton-block/base';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-inline/base';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-text/base';
+@import 'node_modules/progressive-web-sdk/dist/components/stepper/base';
+@import 'node_modules/progressive-web-sdk/dist/components/tabs/base';
```
jeffkamo: As mentioned in the other PR, I would say leave of these uncommented except for the components we do include by default (button).

My preference is to only opt-in to a component when we know it's going to be used, to avoid unused code from slipping into the production release.

app/styles/_components.scss
```diff
@@ -1,22 +1,73 @@
...
+@import 'node_modules/progressive-web-sdk/dist/components/nested-navigation/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/pagination/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/rating/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/scroll-to/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/select/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/sheet/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-block/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-inline/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-text/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/tabs/theme';
```
jeffkamo: Same idea with the theme files as my above comment.

app/styles/_components.scss
```diff
@@ -1,22 +1,73 @@
...
 
 
-// SDK Components' Theme Styles
+// Local Components
 // ---
 
+@import '../components/login-form/base';
```
jeffkamo: Canwe keep this alphabetical?

app/styles/_components.scss
marlowpayne: I've been a bit conflicted about that.

On one hand, including all base and theme styles from the SDK means that a designer/developer can use the [generator](https://github.com/mobify/generator-progressive-web) to scaffold out a new Progressive Web project, boot up the styleguide and immediately be able to start intelligently building with the SDK components. They won't need to look at an SDK component in the styleguide and first have to uncomment some styles to see how it's expected to function. On the other hand, like you mentioned, this carries the risk that unnecessary code can make it through.

What about leveraging `styleguide.scss`? It can import all SDK styles, and it won't be bundled up with the project. Duplicate SDK style imports in both the project stylesheet and `styleguide.scss` should be fine.

jeffkamo: (I can't help but lament the fact that this probably would've been a cinch with Themr... oh well)

Hmm, well leveraging `styleguide.scss` does sound like it could be a reasonable middle ground, but the developer still has to remember to import the styles into `stylesheet.scss` or the final bundle will be missing styles. I could see that happening very easily.

marlowpayne: I agree. It would be nice if we had some automated "include-sdk-component" NPM script that would set everything up nicely for you. Combined with `npm run add:component` for local project components, `_components.scss` could be completely untouched by humans.

jeffkamo: actually on second thought, maybe this is a none issue. Assuming you're working in both the style guide *and* the project preview, then you 100% have a chance to see whether base styles are being applied. Also QA would notice too.

Okay maybe this is a good call then 👍

marlowpayne:
Now that v0.5.1 of the SDK is out, a refresh of your node modules (`rm -r node_modules/progressive-web-sdk && npm i`) will let you see these changes' full effect locally.

styleguide/styleguide.scss
```diff
@@ -1 +1,77 @@
...
+@import 'node_modules/progressive-web-sdk/dist/components/nested-navigation/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/pagination/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/rating/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/scroll-to/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/select/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/sheet/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-block/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-inline/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/skeleton-text/theme';
+@import 'node_modules/progressive-web-sdk/dist/components/tabs/theme';
```
jeffkamo: maybe let's not include the theme styles. This actually can create problems. It could result in the developer overriding styles that don't need to be overridden (these theme styles won't necessarily be included in prod), or these will add additional styles that aren't wanted for the prod site.

I feel like these would do more harm than good. The base styles are definitely necessary, but not the themes.

jeffkamo: Ugh, I just deleted my comment intending to copy and paste it, but lost the it from my clip board some how...

Long story short: I feel like adding the SDK theme styles will do more harm than good here and should not be included. The base styles should be enough.

jeffkamo: Ugh, I just deleted my comment intending to copy and paste it, but lost the it from my clip board some how...

Long story short: I feel like adding the SDK theme styles will do more harm than good here and should not be included. The base styles should be enough.

jeffkamo: It'll either force developers to override styles they don't need to override, or they see styles that shouldn't be there in the style guide but not on the prod site.

jeffkamo:
👍

PR #94, opened 2016-09-20
Add script charset to main.js

fractaltheory:
If a desktop document does not specify a character set using this HTML tag:
```html
<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
```
then there are instances where certain characters (i.e. Japanese text) are improperly encoded in our `main.js` file by the browser. This leads to invalid rendered text.

Specifying a `charset` attribute on the script element when we add it to the page fixes the issue.

 **JIRA**: [WEB-787](https://mobify.atlassian.net/browse/WEB-787)

## Changes
- `loader.js` adds a `charset=utf-8` attribute to `main.js` script element
- Removed unnecessary test setup file

## How to test-drive this PR
Unfortunately, Merlins' Potions has the correct `<meta>` tag described above in the PR summary, so there's some hoops to jump through to test this properly.
- Import this Charles rewrite rule: [merlinspotions.xml.zip](https://github.com/mobify/progressive-web-scaffold/files/481386/merlinspotions.xml.zip)
  - Or, rewrite both these strings to blank:
    - `<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>`
    - `<meta charset="utf-8"/>`
- Comment out [L47 of loader.js](https://github.com/mobify/progressive-web-scaffold/compare/add-script-charset?expand=1#diff-85044fad63ed16d573b29b4da020c90bR47)
- `npm run dev`
- Note that the Japanese text is incorrect
- Revert the above changes
- See that the text renders correctly


MikeKlemarewski:
👍  other than my comment

PR #99, opened 2016-09-22
Add SkipLinks

jeffkamo:
SkipLinks are an accessibility feature that should be included on all of Mobify's project *by default*. This way, screen reader users, assistive technology and keyboard users are able to skip to relevant content much more easily.

## Changes
- Added SkipLink component to scaffold
- Added nightwatch tests that check to ensure the SkipLinks and their targets are valid (at least on the homepage)
- Make Merlin's Potions the default initial URL for local Nightwatch tests

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlins Potions](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.min.js&disabled=0&domain=&scope=1)
- (if testing in desktop Chrome) Ensure that dev tools is closed, and then...
    - Click the browser's URL bar
    - On your keyboard, press `tab` until you arrive on the page itself
    - Ensure that you are able to tab to the "Skip to x" links
    - Ensure pressing `enter` on any of the skip links sends you to that link's destination
- (if testing on device) Ensure VoiceOver is activated, and then...
    - Tap Safari's url bar
    - Swipe right ("tabbing") until you arrive on the page itself
    - Ensure that you are able to tab to the "Skip to x" links
    - Ensure double tapping the screen ("clicking") on any of the skip links sends you to that link's destination

### DIFF
--- a/app/components/skip-links/index.jsx
+++ b/app/components/skip-links/index.jsx
@@ -1,7 +1,6 @@
 import React, {PropTypes} from 'react'
 
-const SkipLinks = ({fragments}) => {
-
+const SkipLinks = () => {
     return (
         <div className="c-skip-links">
             <a href="#app-main" className="c-skip-links__anchor">
@@ -15,7 +14,6 @@ const SkipLinks = ({fragments}) => {
             <a href="#app-footer" className="c-skip-links__anchor">
                 Skip to footer
             </a>
-
         </div>
     )
 }
--- a/app/components/skip-links/index.jsx
+++ b/app/components/skip-links/index.jsx
@@ -1,8 +1,11 @@
 import React, {PropTypes} from 'react'
+import classNames from 'classnames'
 
-const SkipLinks = () => {
+const SkipLinks = ({className}) => {
+    const classes = classNames('c-skip-links', className)
+
     return (
-        <div className="c-skip-links">
+        <div className={classes}>
             <a href="#app-main" className="c-skip-links__anchor">
                 Skip to content
             </a>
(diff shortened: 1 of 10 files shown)


app/containers/app/container.js
```diff
@@ -15,22 +16,25 @@ class App extends React.Component {
         let currentTemplate = `t-${this.props.children.props.route.routeName}`
 
         return (
-            <div id="outer-container" className="t-app">
+            <div id="app" className="t-app">
```
jeffkamo: I wouldn't mind confirmation that these IDs weren't named anything specific for a reason, and if these are safe to change?

app/components/skip-links/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+            <a href="#app-navigation" className="c-skip-links__anchor">
+                Skip to main navigation
+            </a>
+
+            <a href="#app-footer" className="c-skip-links__anchor">
+                Skip to footer
+            </a>
+        </div>
+    )
+}
```
DanielBrookRoberge: I guess by being in the scaffold it isn't meant to be a highly reusable component, but do we really want to remove the possibility of the user passing in additional `className` props?

app/containers/app/container.js
```diff
@@ -15,22 +16,25 @@ class App extends React.Component {
...
                 <IconSprite />
+                <SkipLinks />
 
-                <main id="page-wrap" className={currentTemplate}>
```
jeffkamo: The `main` element is often misused as an overall wrapper when it *should* in fact be used as a wrapper around only the relevant main content.

...In fact, I meant to re add this element in the the right place. I'll do that as well as include some other landmark roles for more a11y goodness.

DanielBrookRoberge:
Also, there are lint errors.

app/components/skip-links/index.jsx
jeffkamo: I was actually just being lazy when I accidentally removed and and didn't add it back in. I'll re-include it now.

PR #110, opened 2016-10-04
Ignore SVG imports in Jest tests

fractaltheory:
With no loader to handle SVG imports, Jest tests fail with a SyntaxError:

screen shot 2016-10-04 at 2 10 32 pm

## Changes
- Add a [moduleNameMapper](https://facebook.github.io/jest/docs/tutorial-webpack.html) alias for anything that has `.svg` extension
- Create a mock file to load instead of SVGs
- Add an example test which imports the `<Icon>` SDK component, which itself imports an SVG via `<IconSprite>`
- Simplify the regEx used for finding test files in the project
- Remove `ignore-styles` imports in test files (as they serve no purpose here - aligns with change to component generator in SDK)

## How to test-drive this PR
- Remove the `moduleNameMapper` object from `package.json`
- Run `npm run test -- --no-cache`
- Verify test fails
- Revert change to `package.json` and repeat steps
- Verify test passes


package.json
```diff
@@ -102,6 +102,9 @@
       "jsx",
       "json"
     ],
+    "moduleNameMapper": {
```
DanielBrookRoberge: This looks cleaner than the search-and-replace approach we're using in the SDK at the moment, do you think you could port it back?

PR #119, opened 2016-10-19
Basic footer for the Merlin's site

olibrook:
**JIRA**: 
- https://mobify.atlassian.net/browse/WEB-930
- https://mobify.atlassian.net/browse/WEB-922

## Changes
- A few bits of global setup (fonts, app-wide action for page loads).
- New footer.

## How to test-drive this PR
- Run locally and review against designs.

## Todos
- Pass on footer CSS/HTML


app/containers/footer/_footer.scss
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+}
+
+.t-footer__panel {
+    padding: 15px;
+}
+
+.t-footer__panel--newsletter {
+    position: relative;
```
jeffkamo: @olibrook what is this `position: relative` used for?

olibrook: Fix display of the border

app/containers/footer/parsers/parser.test.js
```diff
@@ -0,0 +1,41 @@
...
+                title: 'Orders and Returns'
+            },
+            {
+                url: 'http://www.merlinspotions.com/catalogsearch/advanced/',
+                title: 'Advanced Search'
+            }
+        ]
+        expect(parser.parseNavigation($content)).toEqual(expected)
+    })
+})
```
DanielBrookRoberge: I sure do love seeing these tests in the project!

app/utils/utils.js
```diff
@@ -24,3 +24,17 @@ export const createAction = (description, ...argNames) => {
...
+/**
+ * Form-encode an arbitrary JS object.
+ */
+export const formEncode = (data) => {
+    const pairs = []
+    for (const k in data) {
+        if (data.hasOwnProperty(k)) {
+            pairs.push(`${encodeURIComponent(k)}=${encodeURIComponent(data[k])}`)
+        }
+    }
```
DanielBrookRoberge: We could replace this loop with `Object.keys(data).forEach()` which will only loop over the own properties. (ES5 so it needs no polyfill)

app/containers/footer/container.js
```diff
@@ -0,0 +1,153 @@
...
+            <div className="t-footer__social u-padding-md">
+                <div className="u-flexbox u-justify-center u-padding-md">
+                    {
+                        social.map((item) => {
+                            const [url, icon, title] = item
+                            return (
+                                <a href={url} className="t-footer__social-link" key={url}
+                                    style={{backgroundImage: `url(${getAssetUrl(icon)})`}}>
+                                    <span className="u-visually-hidden">{title}</span>
+                                </a>
```
DanielBrookRoberge: Are these icons not available through the Icon component?

jeffkamo: Unfortunately not at the moment. Shoppicon, the SDK's default icon set, does not include any social icons.

Whenever I get around to tackling the icons properly, I'll be sure to include the social icons too.

jeffkamo:
I think I've applied everything I needed to add to this PR. 👍

jeffkamo:
Except for perhaps that breaking test... looking into it.

app/containers/footer/actions.js
```diff
@@ -0,0 +1,31 @@
...
+export const signUpToNewsletter = (action, method, data) => {
+    return (dispatch) => {
+
+        const onSuccess = () => {
+            dispatch(newsletterSignupComplete(constants.SIGNUP_SUCCESSFUL))
+        }
+
+        const onFail = () => {
+            dispatch(newsletterSignupComplete(constants.SIGNUP_FAILED))
+        }
```
olibrook: I don't have grand plans, but we *are* supposed to do a notification thing on success, which is why I set it up the way it is. Gonna leave it, if that's cool.

DanielBrookRoberge: Yeah that's absolutely fine, though I realize we usually structure this as 
```javascript
return utils.makeRequest(action, opts)
    .then(() => stuff)
    .catch(() => otherStuff)
```
which is why it looked odd to me.

app/utils/utils.js
olibrook: Fixed

app/containers/footer/actions.js
olibrook: Oh right. I thought there was a preference generally for assigning names to these callbacks, for debugging purposes?

DanielBrookRoberge: IIRC we still get a reasonable stack trace without that for something simple. But I could be wrong.

PR #126, opened 2016-10-21
Add parser utils from the delivery builds with tests

DanielBrookRoberge:
We should parse elements like links and images in a uniform way, so I looked in the Crabtree and Lancome codebases for the parser utilities they found useful. This imports them to the scaffold and adds tests

## Changes
- Add generic parsers for `a`, `button`, `img`, `option`, and `select` elements
- Add tests

## How to test-drive this PR
- run the tests!

### DIFF
--- a/app/utils/parser-utils.js
+++ b/app/utils/parser-utils.js
@@ -1,0 +1,42 @@
+export const parseLink = ($link) => {
+    return {
+        href: $link.attr('href'),
+        text: $link.text(),
+        title: $link.attr('title')
+    }
+}
+
+export const parseButton = ($button) => {
+    return {
+        text: $button.text(),
+        type: $button.attr('type'),
+        name: $button.attr('name'),
+        value: $button.attr('value'),
+        disabled: !!$button.attr('disabled')
+    }
+}
+
+export const parseImage = ($img) => {
+    return {
+        title: $img.attr('title'),
+        alt: $img.attr('alt'),
+        src: $img.attr('x-src') ? $img.attr('x-src') : $img.attr('src')
+    }
+}
+
+export const parseOption = ($option) => {
+    const value = $option.attr('value')
+    return {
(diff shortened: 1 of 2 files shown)


app/utils/parser-utils.js
```diff
@@ -0,0 +1,42 @@
...
+    return {
+        href: $link.attr('href'),
+        text: $link.text(),
+        title: $link.attr('title')
+    }
+}
+
+export const parseButton = ($button) => {
+    return {
+        text: $button.text(),
```
jeffkamo: should the key be `text` or `children` and the value target the node's content instead of just `.text()`?

app/utils/parser-utils.js
```diff
@@ -0,0 +1,42 @@
+export const parseLink = ($link) => {
+    return {
+        href: $link.attr('href'),
+        text: $link.text(),
```
jeffkamo: Same here, should we do more like `children: $link.html()`? Or would outputting markup in this way actually cause problems?

app/utils/parser-utils.js
DanielBrookRoberge: Maybe? I just copy-pasted this from Crabtree and am not especially familiar with what we'll need for this.

app/utils/parser-utils.js
DanielBrookRoberge: I don't think we want to be outputting markup from the parsers. If we do that then the React side needs to either parse the HTML or DangerouslySet it.

jeffkamo: gotcha, fair point

app/utils/parser-utils.js
mikenikles: Given that a `<button>` element can have arbitrary children, I vote for `children` as the key.

olibrook:
👍

PR #127, opened 2016-10-21
Merlin's Potions PLP

fractaltheory:
**JIRA**: [WEB-917](https://mobify.atlassian.net/browse/WEB-917)

## Notes
- The category images weren't added to this branch, as they exist in `develop` - so they'll appear as broken

## Changes
- Adds Merlins Potions PLP
- Adds ProductTile component
- Adds `getComponentName` utility method
- Adds `onRouteChanged` global action (fired when route changes)
- Adds styles to `c-grid`
- Adds app-level exported constant `FETCH_IN_PROGRESS` as well as adds it to the app store object as `fetchInProgress`

## How to test-drive this PR
- `npm run dev`
- Click the `View potions` link on the homepage


app/components/grid/_base.scss
```diff
@@ -0,0 +1,41 @@
...
+//
+.c-grid {
+    @include clearfix();
+}
+
+// Grid Item
+// ---
+
+.c-grid__item {
+    min-height: 160px;
```
jeffkamo: For all the box values for this component, can they be in multiples of the `$unit` variable?

app/components/grid/_base.scss
```diff
@@ -0,0 +1,41 @@
+// Grid
```
jeffkamo: In fact, the Grid component doesn't appear to have any jsx associated to it. Should this actually be a utility? Or should this be abstracted to indeed have a JSX component?

app/components/grid/_base.scss
```diff
@@ -0,0 +1,41 @@
...
+
+.c-grid__item-name {
+    padding: 16px 0 24px;
+
+    color: black;
+    font-weight: 300;
+    font-size: 28px;
+    line-height: 1.24;
+    word-break: break-word;
+    text-transform: uppercase;
```
jeffkamo: There appears to be more to this grid than just layout. Is this really only a grid component?

jeffkamo: Instead of adding typographic CSS here, you could make use of heading utility classes instead.

app/components/grid/_base.scss
jeffkamo: As a general rule of thumb, if a set of styles don't have an associated React component, it's no longer considered a component.

app/containers/plp/_plp.scss
```diff
@@ -1,7 +1,86 @@
...
 // ===
 
+// [1] - Display box-shadow properly at bottom of container
+
 .t-plp {
-    // Root Class. Contains all core styles, typically for the template's
-    // wrapper element.
+    font-family: 'Oswald';
+
+    .container {
```
jeffkamo: If a class is added by us, be sure to name space it to the file it's created in. In this case, and for many of the below classes, should be namespaced with `t-plp__*` and no longer need to be nested inside of the `t-plp` block

app/containers/plp/_plp.scss
```diff
@@ -1,7 +1,86 @@
...
+    .no-results__text {
+        margin: 20px 50px 32px;
+
+        color: #3a3a48;
+        font-family: $font-family;
+        font-weight: 300;
+        font-size: 14px;
+        line-height: 20px;
+        letter-spacing: 0;
+        text-align: center;
```
jeffkamo: A lot of the typographic and spacing styles being added via CSS could instead be added using utility classes added to the markup. Can these be adjusted to use those utility classes?

app/containers/plp/_plp.scss
```diff
@@ -1,7 +1,73 @@
...
+.t-plp__heading {
+    position: relative;
+
+    .c-link {
+        font-size: $unit * 1.5;
+    }
+}
+
+.t-plp__heading-logo {
+    position: absolute;
```
jeffkamo: Would good to have some comments that provide some context for what this absolute position is relative too.

app/containers/plp/_plp.scss
```diff
@@ -1,7 +1,73 @@
...
+}
+
+.t-plp__grid-item-price {
+    color: $light-error-color;
+    font-weight: 600;
+    font-size: $unit * 2;
+}
+
+.t-plp__heading {
+    position: relative;
```
jeffkamo: Would good to have some comments that provide some context for why this relative positioning is needed.

app/containers/plp/_plp.scss
```diff
@@ -1,7 +1,73 @@
...
+
+    .c-link {
+        font-size: $unit * 1.5;
```
jeffkamo: Minor: this is probably fine, although generally for fonts I start with `$font-size` and add/subtract to achieve the desired size.

fractaltheory: Ah cool yes that makes more sense: https://github.com/mobify/progressive-web-scaffold/pull/127/commits/526a91ec67e4f91791d31ea2b7de823227998b01

app/containers/plp/container.js
```diff
@@ -1,41 +1,104 @@
...
+                        <h1 className="u-text-lighter u-text-uppercase">{title}</h1>
+                    }
+                </div>
+
+                <Image
+                    alt="Heading logo"
+                    className="t-plp__heading-logo"
+                    height="51px"
+                    src={isPlaceholder ? '' : getAssetUrl(`static/img/${title.trim().toLowerCase()}.png`)}
+                    width="61px"
```
DanielBrookRoberge: Could this be a `SkeletonBlock` when `isPlaceholder` is true?

app/containers/plp/container.js
```diff
@@ -1,41 +1,104 @@
...
+     */
+    title: PropTypes.string.isRequired
+}
+
 const mapStateToProps = (state) => {
+    const selector = state.plp.get(SELECTOR)
+
     return {
-        ...state.PLP,
+        ...state.plp.get(selector).toJS(),
```
DanielBrookRoberge: Should you use the `selector` var here?

DanielBrookRoberge: Also, if you pass in the Immutable object (either instead of or in addition to the `toJS` one), you can easily write a shouldComponentUpdate function which we probably want.

app/containers/plp/reducer.test.js
```diff
@@ -1,15 +1,17 @@
...
-        test: true,
-        item: false,
+        expect(newState).not.toBe(initialState)
     })
 
-    expect(reducer(inputState, action)).toBe(inputState)
+    test('does nothing for unknown action types', () => {
+        const action = {type: 'qwertyuiop'}
+        expect(reducer(initialState, action)).toBe(initialState)
+    })
```
DanielBrookRoberge: Would probably be good to test the storage of different page information in here. I think with jsdom we can do whatever we like with window.location.href without causing problems.

app/containers/plp/container.js
fractaltheory: I need to use `selector` since it allows me to choose which entry in the PLP store to pull data from (i.e. which href we're on).

One thing I'm unsure how to resolve when not using `toJS` is defining `propTypes`. Any ideas there? Or have we solved that problem in other branches I'm not aware of?

DanielBrookRoberge: Yeah in the first comment I misread this line as `...state.plp.get(SELECTOR).toJS()`.

Yeah I don't think we have a good solution for propTypes yet. Oli set something up in https://github.com/mobify/progressive-web-sdk/pull/330 but I'm not sure if it's a good enough solution. One thing I've done as kind of a transitional solution is pass the toJS'd object and the Immutable one, rendering with the toJS props and implementing shouldComponentUpdate with the Immutable one.

fractaltheory: Okay, I've added a `shouldComponentUpdate` check: https://github.com/mobify/progressive-web-scaffold/pull/127/commits/1dbe79976076e4ecf9dc341a8bf2e5709c0c845e

app/containers/plp/reducer.test.js
DanielBrookRoberge: Yes it is what I meant. Glad the workaround worked.

DanielBrookRoberge:
:+1: from me!

PR #128, opened 2016-10-21
Add Global Styles

jeffkamo:
Add some of the necessary global styles to the scaffold for use with Merlin's Potions.

## Changes
- Add form styles, such as radio buttons, checkboxes and buttons
- Add some styles for common components, like Icon, List and ListTile
- Tweak some base styles, such as typography
- Add base colors: brand, accent, success, error, neutrals, etc.

## How to test-drive this PR
- Run `npm run styleguide`
- Visit [http://localhost:4000/](http://localhost:4000/)
- Confirm that the above mentioned changes are reflected in the styles


app/components/list-tile/_theme.scss
```diff
@@ -0,0 +1,19 @@
+// List Tile
+// ===
+
+.c-list-tile__primary {
+    font-size: $font-size + 1; // 15px
```
Magitron: I just saw that you added a comment for the one above, but not this one. I think it would be good to have the same commenting method for consistency :)

app/containers/footer/container.js
```diff
@@ -84,7 +84,7 @@ const Footer = (props) => {
             <div className="t-footer__newsletter u-padding-md u-padding-top-lg u-padding-bottom-lg">
                 {newsletter ? (
                     <div>
-                        <h2 className="u-h3 u-margin-bottom-md">
+                        <h2 className="u-h2 u-margin-bottom-md">
                             Subscribe to Merlin's Newsletter
```
Magitron: Just wondering, does anything need to be escaped here?

jeffkamo: I don't think so. I assume this weird syntax highlight is just because JSX isn't a recognized syntax in Github yet (or maybe it's because the file is `.js` instead of `.jsx`)

app/components/list-tile/_theme.scss
jeffkamo: Yup, you're right. I was being lazy :( thanks for keeping me honest! :D

jeffkamo:
Verbal +1 from MQ

PR #136, opened 2016-10-25
Working navigation for the Merlin's site

olibrook:
**JIRA**: https://mobify.atlassian.net/browse/WEB-922

## How to test-drive this PR
- Code review
- Test in the browser (click the menu button to reveal)


app/components/nav-item/index.jsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react'
+import * as assetUtils from 'progressive-web-sdk/dist/asset-utils'
+import * as DefaultNavItem from 'progressive-web-sdk/dist/components/nav-item'
```
DanielBrookRoberge: Since the nav-item/index.js only has a default export, this can just be `import DefaultNavItem from '...'`

app/components/nav-item/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+        title,
+        className,
+        childIcon,
+        beforeContent: beforeContentProp,
+        content: contentProp,
+        hasChild,
+        path,
+    } = props
+
+    const before = beforeContentProp
```
DanielBrookRoberge: Is this necessary, since there's no change?

app/components/nav-item/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+    const after = hasChild ? childIcon : null
+    const classes = classNames('c-nav-item', className,
+        {
+            'c-nav-item--has-child': hasChild,
+            'c-nav-item--selected': selected,
+        }
+    )
+
+    return (
+        <Link className={classes} href={path} onClick={navigate}>
```
DanielBrookRoberge: Do you think we should change the default NavItem to use a Link? Connecting to the router seems like something we'll commonly need to do.

app/components/nav-item/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Project-specific NavItem which displays an icon on the left.
+ */
+export const NavItemWithIcon = (props) => {
+    return (
+        <NavItem {...props} className="c-nav-item--with-icon" />
+    )
+}
+
+NavItemWithIcon.propTypes = NavItem.propTypes
```
DanielBrookRoberge: If we already have the NavItemWithIcon component, does it make sense for it to manage the `beforeContent`? It could take an `iconSrc` prop or the like.

olibrook:
Ah, good catch about the `path`. I forgot to push my change to the SDK, which also needs a trivial review - https://github.com/mobify/progressive-web-sdk/pull/319

app/components/nav-item/index.jsx
olibrook: Fixed

app/components/nav-item/index.jsx
olibrook: I would rather not to be honest - my reasoning is that (1) react-router on its own doesn't really fit the redux model particularly well and (2) doing so makes it awkward to do some of the things we discussed doing - ie. expanding an item in the menu rather than selecting it and actually triggering a URL change. Swapping them out for a `Link` triggers an actual URL change for every item unless we are much more prescriptive with what people do with that `Nav` component.

app/components/nav-item/index.jsx
olibrook: I am shooting for a component that takes no additional props over what the default `NavItem` in the SDK provides so that this can trivially be thrown into a factory function. Either way, I don't think this is a blocker, is it?

app/components/nav-item/index.jsx
olibrook: Fixed

app/components/nav-item/index.jsx
olibrook: The copypasta `NavItem` is now gone, no need to use `Link` in the nav.

app/components/nav-item/_theme.scss
```diff
@@ -0,0 +1,94 @@
...
+    height: $unit * 2;
+
+    transform: translate(-50%, -50%);
+}
+
+
+.c-nav-item--with-icon {
+    padding: 0;
+
+    background-color: #f8f8f8;
```
jeffkamo: Can we use `$neutral-20` instead (`#f7f7f7`)?

app/containers/navigation/_navigation.scss
```diff
@@ -0,0 +1,25 @@
+// Navigation
```
jeffkamo: This navigation container could potentially use an instance of the headerBar component form the SDK, which I am styling in a different branch. Might refactor these after they get merged together.

app/static/img/icon-account.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="26px" height="26px" viewBox="-1 -1 26 26" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jeffkamo: Can this file be added to `static/svg` instead?

app/static/img/icon-close.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="16px" height="16px" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jeffkamo: Same here: add the file to `static/svg`

app/static/img/logo.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="144px" height="21px" viewBox="0 11 144 21" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jeffkamo: Same here: add the file to `static/svg`

jeffkamo: actually we'll want to have a different name for this logo perhaps, which is different in appearance than the logo that appears in the normal site header.

app/components/icon-text-button/_base.scss
```diff
@@ -0,0 +1,41 @@
+// IconTextButton
+
+.c-icon-text-button,
+.c-icon-text-button:active,
+.c-icon-text-button:focus {
```
jeffkamo: Do all those pseudo selectors need to be selected?

app/components/nav-item/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+import NavItem from 'progressive-web-sdk/dist/components/nav-item'
+
+/**
+ * Icon used on project-specific nav items
+ */
+const NavItemIcon = (props) => {
+    const {src} = props
+    return (
+        <div className="c-nav-item__icon">
+            <img role="presentation" className="c-nav-item__icon-content" src={src} />
```
jeffkamo: Should we be using the Image component here?

jeffkamo: Or, should this really be using Icon?

app/components/nav-item/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+    src: React.PropTypes.string,
+}
+
+
+/**
+ * Project-specific NavItem which displays an icon on the left.
+ */
+export const NavItemWithIcon = (props) => {
+    return (
+        <NavItem {...props} className="c-nav-item--with-icon" />
```
jeffkamo: Our convention for modifier classes is to have two. So normally, for this particular example, we would have: `c-nav-item c--with-icon` and the CSS would be written `.c-nav-item.c--with-icon {...}`

app/components/nav-item/index.jsx
DanielBrookRoberge: Yeah that looks good. I see your point about the Link in the NavItem. I do like the new solution.

app/components/nav-item/index.jsx
DanielBrookRoberge: Definitely not a blocker, just a thought.

app/components/nav/_theme.scss
```diff
@@ -0,0 +1,7 @@
+.c-nav {
+    position: absolute;
```
olibrook: Documented.

app/components/nav-item/_theme.scss
```diff
@@ -0,0 +1,94 @@
+// NavItem Theme
+// ===
+
+$navitemBlack: #3a3a48;
```
olibrook: Fixed

app/components/icon-text-button/index.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import classNames from 'classnames'
+import Button from 'progressive-web-sdk/dist/components/button'
+
+
+const IconTextButton = (props) => {
```
olibrook: I don't think that has been merged yet, so I can't use it. If you want to merge your changes I can swap my component for yours.

app/components/icon-text-button/_base.scss
```diff
@@ -0,0 +1,41 @@
...
+    padding: 0;
+    border: 0;
+    background: none;
+
+    background-color: $brand-color;
+    color: white;
+}
+
+.c-icon-text-button__icon {
+    position: absolute;
```
olibrook: Documented

app/components/icon-text-button/_base.scss
```diff
@@ -0,0 +1,41 @@
...
+    top: $unit;
+    left: 50%;
+    width: ($unit * 2);
+    height: ($unit * 2);
+    transform: translateX(-50%);
+}
+
+.c-icon-text-button__text {
+    position: absolute;
```
olibrook: Documented

app/static/img/logo.svg
olibrook: Renamed to `nav-logo.svg`

app/static/img/icon-account.svg
olibrook: Moved

app/static/img/logo.svg
olibrook: Moved into static/svg

app/components/icon-text-button/index.jsx
jeffkamo: Hmm, should be there. It was part of the [0.8.1 SDK release](https://github.com/mobify/progressive-web-sdk/releases/tag/v0.8.1)

olibrook: Ooooh, I thought you meant a Merlin's component.

app/components/nav-item/index.jsx
olibrook: Swapped for `Icon`

app/components/nav-item/index.jsx
olibrook: Fixed

olibrook: Swapped the selector to `.c-nav-item.c--with-icon` in the CSS and made the corresponding change in the SDK (this doesn't depend on the SDK change).

See https://github.com/mobify/progressive-web-sdk/pull/320

app/components/icon-label-button/_base.scss
```diff
@@ -0,0 +1,17 @@
+// IconLabelButton
+
+.c-icon-label-button,
+.c-icon-label-button:active,
+.c-icon-label-button:focus {
+    position: relative;
```
jeffkamo: Last comment: a comment to describe why this position relative is needed

PR #139, opened 2016-10-26
Remove props argument and gratuitous exports from mapDispatchToProps

DanielBrookRoberge:
Redux will cache the results of `mapDispatchToProps` for a given component instance, keeping the functions' identities through multiple render calls, but only if the `mapDispatchToProps` function has no second `props` argument. This is a performance problem when `props` is unused, so this PR removes it from the currently merged components.

Mainly, the problem is that the functions would be created anew each render, which makes `shouldComponentUpdate` on child components think that things have changed.

It also removes the `export` on the mapStateToProps and mapDispatchToProps functions since that's not actually useful anywhere.

 **JIRA**: N/A
## Changes
- Fix the Login container to have the correct arguments and exports.


mikenikles:
👍

PR #140, opened 2016-10-26
Add a utility wrapper for submitting forms.

DanielBrookRoberge:
We make many requests with form-encoded data and a corresponding Content-Type. This PR adds a utility function that can be passed a JS object and makes sure to form-encode it and set the Content-Type before making the request. 

 **JIRA**: N/A
## Changes
- Add the described utility


mikenikles:
👍

PR #143, opened 2016-10-26
PDP Styling

Magitron:
Styling the PDP
## Changes
- [x] Style the heading text
- [x] Style the carousel
- [x] Style the accordion
- [x] Style the stepper
- [x] Create indicator for stock
- [x] Add 'add to cart' button
- [x] Add sheet
- [x] Style "Added to Cart" sheet
- [x] Fix broken tests

## Things to tackle soon
**Not required for this PR**

- [ ] @jeffkamo: Fix SVG icons to `fill` based on `currentColor`
- [ ] @jeffkamo: Replace instances of `u-heading-family` with `u-text-header-font-family`

## How to test-drive this PR
- Run Merlins
- click on PDP


app/containers/pdp/container.js
```diff
@@ -18,26 +20,48 @@ class PDP extends React.Component {
...
 
         const {
-            contentsLoaded = false,
             product = {},
-            itemQuantity = 1
+            itemQuantity = 1,
+            quantityAdded = 0,
+            formInfo = {},
+            itemAddedModalOpen = false,
+            contentsLoaded = false
```
jeffkamo: Should these default properties be added via `PDP.defaultProps` method?

jeffkamo: Oh nevermind these aren't coming from `this.props`

jeffkamo: Since these are coming from the `pdp` state, aren't these defaulting to the values as per the `initialState` object in the reducer?

app/containers/pdp/container.js
```diff
@@ -18,26 +20,48 @@ class PDP extends React.Component {
...
+                            onSubmit={addToCart} />
+                        <PDPItemAddedModal
+                            open={itemAddedModalOpen}
+                            onDismiss={closeItemAddedModal}
+                            product={product}
+                            quantity={quantityAdded} />
+                    </div>
+                :
+                    <div className="u-padding-md">
+                        <SkeletonText lines={5} width="100%" size="24px" lineClassName="u-margin-bottom" />
```
jeffkamo: Since the content in this area isn't plain text (accordions and form fields) are these skeletons really all that helpful?

app/containers/pdp/partials/pdp-add-to-cart.jsx
```diff
@@ -0,0 +1,34 @@
+import React, {PropTypes} from 'react'
+import Button from 'progressive-web-sdk/dist/components/button'
+import Stepper from 'progressive-web-sdk/dist/components/stepper'
+
+const PDPAddToCart = ({quantity, setQuantity, onSubmit}) => {
+
+    return (
+        <form className="u-padding-start-md u-padding-end-md">
+            <div className="t-pdp__stepper u-margin-top-lg">
```
jeffkamo: This `t-pdp__stepper` doesn't appear to be used. Can it be removed?

app/containers/pdp/partials/pdp-add-to-cart.jsx
```diff
@@ -0,0 +1,34 @@
...
+                <label htmlFor="quantity">Quantity</label>
+                <div className="u-flexbox u-margin-bottom-lg u-margin-top">
+                    <Stepper decrementIcon="minus" incrementIcon="plus" initialValue={quantity} minimumValue={1} onChange={setQuantity} />
+                    <div className="t-pdp__indicator u-border">In stock</div>
```
jeffkamo: I think this element is missing the checkmark icon?

screen shot 2016-11-03 at 6 13 17 pm

app/containers/pdp/partials/pdp-add-to-cart.jsx
```diff
@@ -0,0 +1,34 @@
...
+                </div>
+            </div>
```
jeffkamo: It may be worth replacing these `div` containers with the field components (`Field` and `FieldRow`).

Actually... not sure how that would work with the stepper... might be missing functionality from the SDK (there's a PR that's in stasis to fix Field's ability to use arbitrary components)...

Thoughts @guisers ?

If we can't use the `Field` components, we should at least use `FieldRow`

app/containers/pdp/partials/pdp-carousel.jsx
```diff
@@ -2,20 +2,24 @@ import React, {PropTypes} from 'react'
...
+                    return (
+                        <CarouselItem key={idx}>
+                            <Image src={item.img} hidePlaceholder={true} loadingIndicator={<SkeletonBlock height="100vw" />} />
+                        </CarouselItem>
+                    )
+                })}
+            </Carousel>
+            : <SkeletonBlock height="100vw" width="100%" />
+    )
+}
```
jeffkamo: To get that clean ternary operator formatting, you could do...

```
return (items.length > 0 ?
    <Carousel>...</Carousel>
:
    <SkeletonBlock height="100vw" width="100%" />
)
```

app/containers/pdp/partials/pdp-description.jsx
```diff
@@ -2,13 +2,11 @@ import React, {PropTypes} from 'react'
...
-    <div className="c-pdp-description">
-        <Accordion>
-            <AccordionItem header="Product Description" closeIconName="x">
-                <p>{description}</p>
-            </AccordionItem>
-        </Accordion>
-    </div>
+    <Accordion>
+        <AccordionItem header="Product Description" closeIconName="x">
+            <p>{description}</p>
```
jeffkamo: Are we confident this will alway be a single paragraph of text? Desktop content could be unpredictable (I know this is Merlins which we control, but still want to think about best practices)

app/components/carousel/_theme.scss
```diff
@@ -1,6 +1,49 @@
...
+// ===
+//
 
-// [1] - Hide the carousel slide numbers from being displayed
+.pw-carousel {
+    position: relative;
+
+    padding: $unit*1.5;
+
+    background: $neutral-20;
```
jeffkamo: Something that's missing is the inset box-shadow from the mockup:

screen shot 2016-11-03 at 6 23 11 pm

app/containers/pdp/partials/pdp-description.jsx
```diff
@@ -2,13 +2,11 @@ import React, {PropTypes} from 'react'
...
 const PDPDescription = ({description}) => (
-    <div className="c-pdp-description">
-        <Accordion>
-            <AccordionItem header="Product Description" closeIconName="x">
-                <p>{description}</p>
-            </AccordionItem>
-        </Accordion>
-    </div>
+    <Accordion>
+        <AccordionItem header="Product Description" closeIconName="x">
```
jeffkamo: The accordion icons don't seem to be taking in custom colors. Let's take a look at that tomorrow.

app/containers/pdp/partials/pdp-item-added-modal.jsx
```diff
@@ -0,0 +1,34 @@
...
+            <button type="button" onClick={onDismiss}>
+                close
+            </button>
+        </div>
+        <div className="c-sheet__content">
+            <img role="presentation" src={carouselItems[0].img} />
+            <div className="c-sheet__details">
+                <p>category</p>
+                <h1>{title}</h1>
+                <h1>{price}</h1>
```
jeffkamo: I assume this isn't finished and that's why there are multiple H1 elements like this? 😅

app/containers/pdp/reducer.js
```diff
@@ -18,7 +27,21 @@ export default createReducer({
             ...pdpParser($, $response)
         })
     },
+    [onRouteChanged]: (state) => {
+        return Immutable.fromJS(initialState)
```
jeffkamo: I see `Immutable.fromJS(initialState)` happening twice, could this be done above when `initialState` is first assigned?

app/app-provider.jsx
```diff
@@ -57,6 +57,7 @@ const AppProvider = ({store}) => {
                     <IndexRoute component={Home} routeName="home" />
                     <Route component={Login} path="customer/account/login/" routeName="login" />
                     <Route component={PDP} path="eye-of-newt.html" routeName="productDetailsPage" />
+                    <Route component={PDP} path="unicorn-blood.html" routeName="productDetailsPage" />
```
jeffkamo: Oh man, no detectable pattern for PDP routes, huh? :feelsbadman:

app/styles/utilities/_text.scss
```diff
@@ -97,6 +97,35 @@
     @include text-quiet(true);
 }
 
+
+// Size
+// ---
+
+.u-text-medium {
+    font-size: 24px;
```
jeffkamo: Instead of making these text utilities, could the heading utility classes be used instead?

app/styles/utilities/_text.scss
```diff
@@ -97,6 +97,35 @@
...
+
+.u-text-accent {
+    color: $accent-color;
+}
+
+
+// Font-Family
+// ---
+
+.u-text-header-font {
```
jeffkamo: This exists already as `u-heading-family`, except the letter-spacing part. I'm not sure all headings necessarily need the letter spacing.

app/styles/base/_typography.scss
```diff
@@ -23,7 +22,9 @@ h6 {
     margin: 0;
 
     font-family: $header-font-family;
+    font-weight: normal;
     line-height: 1.25;
+    letter-spacing: 1px;
```
jeffkamo: Not sure we want all headings to have 1px letter spacing. I see some headings in the mocks that have 0 letter spacing.

app/containers/pdp/container.js
guisers: I just thought it looked nice... the space down there is kinda empty otherwise and this sort of gives off the signal that there will be content there. (Also the description, which is the only accordion item, is text content anyway)

app/containers/pdp/partials/pdp-add-to-cart.jsx
guisers: Isn't `Field` for inputs? I don't think the "In stock" `div` is an input, and the `Stepper` comes with its own input... not sure what is being suggested here

app/containers/pdp/partials/pdp-carousel.jsx
guisers: perfect haha I think i used some inconsistent ternary operator formatting throughout cause I wasn't sure what was best. is this something we can add to the linter?

app/components/carousel/_theme.scss
```diff
@@ -1,6 +1,49 @@
-// Carousel theme
```
Magitron: Ah, that's much more helpful to see now that all the templates are merged.

app/containers/pdp/partials/pdp-item-added-modal.jsx
guisers: Modal isn't done being styled yet
@Magitron

app/containers/pdp/reducer.js
guisers: Good catch 👍 That's leftover from a time when i needed the `initialState` in its JS form somewhere in that file, but that part has since been cut out

app/containers/pdp/partials/pdp-item-added-modal.jsx
Magitron: yes :)

app/app-provider.jsx
guisers: Yeah. 😭 

I heard we could implement a `SelectorRoute` but probably the least friction solution is to fix up Merlin's desktop (unless we _never_ get access to it)...

app/components/accordion/_theme.scss
```diff
@@ -0,0 +1,37 @@
...
+.pw-accordion__inner-header {
+    align-items: center;
+}
+
+.pw-accordion__title {
+    font-size: $font-size+2;
+    text-align: left;
+}
+
+.pw-accordion__content {
```
Magitron: 🤔 So the idea was to use `pw-` for SDK styles and `c-` for project styles?

I think that makes sense. And I think it will be much easier to remember for someone who's not working on both at the same time or in quick succession. My only concern would be that during later maintenance or during partner work the namespaces will get mixed up.

jeffkamo: That's exactly my worry. On a project, whether or not it's an SDK component or local component it gets styled in the same place. That seems problematic :\

app/containers/pdp/partials/pdp-add-to-cart.jsx
jeffkamo: Sort of, yes. Technically, the way Field works now is it only works with normal inputs. But we want to be able to eventually provide "custom" inputs as well (date fields, steppers, credit card, etc.)

Field (and FieldRow) is really just a helper for laying out common field elements (label, captions, errors, inputs, and related content like the "In Stock" notice)

app/containers/pdp/partials/pdp-description.jsx
jeffkamo: ya, I think this would be a common example of where we would use the `DangerousHTML` component

app/components/accordion/_theme.scss
mlworks: Yea I like the idea of `c-` for project styles and `pw-` for base SDK styles but same concern as you guys regarding partners mixing up the namespaces in their work which is almost guaranteed bound to happen. 

@jeffkamo  I thought we were gonna phase out `c-` eventually in the SDK components, `c-` is only in there for current backwards compatibility right?

app/containers/pdp/partials/pdp-item-added-modal.jsx
```diff
@@ -0,0 +1,34 @@
+import React, {PropTypes} from 'react'
+import Link from 'progressive-web-sdk/dist/components/link'
+import Sheet from 'progressive-web-sdk/dist/components/sheet'
+
+const PDPItemAddedModal = ({open, onDismiss, quantity, product: {title, price, carouselItems}}) => (
+    <Sheet open={open} onDismiss={onDismiss} effect="slide-bottom">
```
jeffkamo: @guisers Heads up: if you add sheets to a project, make sure you give them some kind of unique identifier (generally in the form of a classname) that will make it easy to find vs. other sheets.

I'll add one to this file, so just keep in mind for future reference.

app/containers/pdp/_pdp.scss
```diff
@@ -5,3 +5,16 @@
     // Root Class. Contains all core styles, typically for the template's
     // wrapper element.
 }
+
+.t-pdp__indicator {
+    flex: 1;
```
mlworks: ^ Yea what Jeff said. `flex: 1` used to work reliably but lately I stick with `flex: 1 1 auto` just to be safe :)

app/components/accordion/_theme.scss
jeffkamo: Correct, the `c-` are still there for backwards compatibility reasons. v1.0.0 of the SDK will see them fully phased out

app/containers/pdp/partials/pdp-item-added-modal.jsx
```diff
@@ -0,0 +1,34 @@
...
+        <div className="c-sheet__heading">
+            <h1>Product Added to Cart</h1>
+            <button type="button" onClick={onDismiss}>
```
jeffkamo: @guisers also make sure you default to using the `Button` component. The general styles that we'll want come from this component rather than just a plain `button` tag.

jeffkamo: cc @Magitron

app/containers/pdp/partials/pdp-item-added-modal.jsx
```diff
@@ -0,0 +1,34 @@
...
+            </button>
+        </div>
+        <div className="c-sheet__content">
+            <img role="presentation" src={carouselItems[0].img} />
+            <div className="c-sheet__details">
+                <p>category</p>
+                <h1>{title}</h1>
+                <h1>{price}</h1>
+                <p>{quantity}</p>
+                <Link href="/checkout/cart"><button type="button">go to checkout</button></Link>
```
jeffkamo: This is the other reason we use the `Button` component, because the component decides to be a Link or button based on whether it is passed an `href` prop AND gives our desired button styles.

app/styles/utilities/_text.scss
Magitron: Right. I made this one because that size wasn't covered in the headings. I changed it to `u-text-size-medium` for now. Do you have better suggestions for font sizing not covered by headings?

app/components/accordion/_theme.scss
cole-sanderson: I think it's okay to use `pw-` since Accordion component is from SDK. If its new project component and it will make sense to use `c-`.

jeffkamo: Here's my main question those: should the SCSS files that style `pw-` classes be in the same folder you'll find styles for `c-` styles?

```
/app
    /components
        /button
            /_theme.scs    <<<<    uses pw- namespace
        /product-item
            /_base.scss    <<<<    uses c- namespace
```

app/containers/pdp/partials/pdp-carousel.jsx
jeffkamo: I don't know, but I've added that question to the next sync meeting's agenda!

app/styles/utilities/_text.scss
jeffkamo: I would say that if a situation like this occurs (where an unidentified type style is being used) it should be raised to the designer to confirm whether or not it was intentional or just an accident.

If intentional, then determining where it fits in the design system. If it's missing from the list of headings, should it be added? Do we need to make a global change?

I'll a note to discuss with Rose on Monday.

In the meantime, I'm going to tweak this to be a one-off instance of `font-size: 24px`. My reasoning here is because: what exactly does `text-medium` mean from a project-by-project basis? If you were to use the same class on another project, how would its value be influenced? Is there some variable that it was use that would make it easy to configure? What part of the design system dictates this value?

app/styles/utilities/_text.scss
jeffkamo: For posterity: we decided to replace `u-heading` classes with `u-text` equivilents

jeffkamo:
I think this is looking pretty good! @guisers if we can just get those tests passing, then we can get this merged!

app/containers/pdp/container.js
DanielBrookRoberge: This probably also better situated in the partials (since there it can have a closer relationship to the actual components)

DanielBrookRoberge:
The test failures are because the tests expect each partial to be wrapped in a `div`. We can probably get rid of a lot of them since the description and carousel now render their SDK component at the top level. Also, IIRC `hasClass` doesn't work correctly on components so that'll need to be replaced by string operations on the `className` prop.

PR #147, opened 2016-10-27
Footer refactoring

DanielBrookRoberge:
The Footer container wasn't using all of the current best practices for containers. This PR refactors it to match current best practice to serve as an example

 **JIRA**: N/A
## Changes
- Split the footer into partials
- Convert to a class with a shouldComponentUpdate method to avoid re-rendering
- Use mapDispatchToProps rather than passing `dispatch` as a prop.
## How to test-drive this PR
- run the tests
- preview Merlins
- see that the footer appears as before.


jeffkamo:
This looks good to me 👍

PR #149, opened 2016-10-27
Site Header

jeffkamo:
Style the Header, and add some other tweaks

 **JIRA**: [WEB-922](https://mobify.atlassian.net/browse/WEB-922)
## Changes
- Style the Site Header
- Add the correct SVG icons from the mockups
- Add a new build script for the SVG sprite (`npm run build-sprite`)
- Tweaks to various components: Button, HeaderBar, IconLabelButton, IconLabel, Icon, Image, NavItem, Sheet and SkipLinks
- Tweaks to various containers: App, Header and Navigation
## How to test-drive this PR
- `npm i`
- `npm run dev`
- Visit [here](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1) to preview Merlin's Potions
- Observe the site header and navigation


app/containers/header/container.js
```diff
@@ -0,0 +1,114 @@
...
+        const headerHeight = 52
+
+        if (window.pageYOffset > headerHeight && !isCollapsed) {
+            this.props.shrinkHeader()
+        }
+
+        if (window.pageYOffset < headerHeight && isCollapsed) {
+            this.props.expandHeader()
+        }
+    }
```
olibrook: Not a blocker, but two comments:

1) Maybe this is worth throttling?
2) We could save ourselves an action, a reducer and a test if we did this:

```
handleScroll() {
    const headerHeight = 52
    const isCollapsed = window.pageYOffset > headerHeight
    this.props.setHeaderState(isCollapsed)  // Single new action to handle shrink/expand
}
```

app/containers/header/container.js
```diff
@@ -0,0 +1,114 @@
...
+    isCollapsed: PropTypes.bool,
+    shrinkHeader: PropTypes.func,
+    onMenuClick: PropTypes.func,
+}
+
+export const mapStateToProps = ({header}) => {
+    return {
+        header: header.toJS()
+    }
+}
```
olibrook: I am trying to convince people to stop using this pattern. It's not a problem right now, since the only state is a boolean, but in order that future maintainers fall into the pit of success I would pull the values out of the header state in your `render` function and do this instead:

```
export const mapStateToProps = (state) => {  // Don't destructure immutablejs objects
    return {
        header: state.header
    }
}
```

olibrook:
Those fixes are nice-to-haves, if you're interested in doing them. Otherwise it's a 👍  from me.

app/containers/header/actions.js
```diff
@@ -1,5 +1,4 @@
 // import * as constants from './constants'
 import * as utils from '../../utils/utils'
 
-export const shrinkHeader = utils.createAction('Shrunk the header')
-export const expandHeader = utils.createAction('Expanded the header')
+export const toggleHeader = utils.createAction('Toggled the header')
```
olibrook: Change this to 

```
export const setHeaderState = utils.createAction('Set header state', 'isCollapsed')
```

app/containers/header/container.js
```diff
@@ -12,6 +13,12 @@ import DangerousHTML from 'progressive-web-sdk/dist/components/dangerous-html'
 
 
 class Header extends React.Component {
+    constructor() {
+        super()
+
+        this.handleScroll = throttle(this.handleScroll.bind(this), 200)
```
olibrook: Nice

app/containers/header/container.js
```diff
@@ -21,21 +28,19 @@ class Header extends React.Component {
...
+        const hasScrolledDown = window.pageYOffset > headerHeight && !isCollapsed
 
-        if (window.pageYOffset > headerHeight && !isCollapsed) {
-            this.props.shrinkHeader()
-        }
-
-        if (window.pageYOffset < headerHeight && isCollapsed) {
-            this.props.expandHeader()
+        if (isCloseToTop || hasScrolledDown) {
+            this.props.toggleHeader()
```
olibrook: Oh, that's not what I meant at all! The previous version was much, much better! :D 

```
handleScroll() {
    const headerHeight = 52
    const isCollapsed = window.pageYOffset > headerHeight  // This gives you the correct new value
    this.props.setHeaderState(isCollapsed)
}
```

That version gives you consistent results - don't depend on doing a no-args toggle like that.

app/containers/header/actions.js
olibrook: Maybe our `utils.createAction` thing isn't obvious enough. You can pass in argument names which you can use in your action, eg `setHeaderState(true)` above.

app/containers/header/container.js
jeffkamo: hmm, interesting. I suppose that makes sense, it will be more predictable. Kk, will fix.

app/containers/header/container.js
```diff
@@ -0,0 +1,112 @@
...
+import Button from 'progressive-web-sdk/dist/components/button'
+import IconLabel from 'progressive-web-sdk/dist/components/icon-label'
+import {HeaderBar, HeaderBarActions, HeaderBarTitle} from 'progressive-web-sdk/dist/components/header-bar'
+import Link from 'progressive-web-sdk/dist/components/link'
+import logo from '../../static/svg/logo.svg'
+import DangerousHTML from 'progressive-web-sdk/dist/components/dangerous-html'
+
+
+class Header extends React.Component {
+    constructor() {
```
olibrook: Oh, also, that needs to be:

```
   constructor(props) {
       super(props)
   }
```

app/containers/header/container.js
```diff
@@ -0,0 +1,112 @@
...
+    handleScroll() {
+        const {isCollapsed} = this.props.header.toJS()
+        const headerHeight = 52
+        const isCloseToTop = window.pageYOffset < headerHeight && isCollapsed
+        const hasScrolledDown = window.pageYOffset > headerHeight && !isCollapsed
+
+        if (isCloseToTop || hasScrolledDown) {
+            this.props.toggleHeader(!isCollapsed)
+        }
+    }
```
olibrook: This still looks a little funky to me, but I can let it slide. Is it to prevent triggering the action needlessly? This would do it:

```
handleScroll() {
    const {isCollapsed} = this.props.header.toJS()
    const headerHeight = 52
    const newIsCollapsed = window.pageYOffset > headerHeight

    if (newIsCollapsed !== isCollapsed) {  // Saves triggering the action
        this.props.toggleHeader(newIsCollapsed)
    }
}
```

PR #150, opened 2016-10-28
set sass linter to inherit from modify-code-style

jeffkamo:
Before, it was difficult to modify the Sass linting rules if there was a need. All the rules were coming directly form the mobify-code-style repo, which meant making tweaks required submitting pull requests to that other repo. Now, however, there is a project specific lint ruleset we can use: `.sass-lint.yml`! 🎉 
## Changes
- Instead of directly using the sass-linter from mobify-code-style directly, create a yml file in the scaffold which inherits from the code style version.
## How to test-drive this PR
- Pull down this branch
- Run `npm run lint:sass`, all the tests should pass!
- Add or tweak the rules in `/.sass-lint.yml` and show that the base ruleset can be overridden ([see list of rules](https://github.com/sasstools/sass-lint/tree/master/docs/rules))


.sass-lint.yml
```diff
@@ -0,0 +1,6 @@
+options:
+    config-file: node_modules/mobify-code-style/css/.sass-lint.yml
+rules:
+    # class-name-format:
```
marlowpayne: My only suggestion would be to add another comment here describing what is going on in this example, to document to developers what they can do in this file (override or extend the code-style's sass-lint ruleset).

PR #151, opened 2016-10-31
Global Component: Mini-Cart

jeffkamo:
Exploratory work on the global minicart component.

 **JIRA**: [WEB-922](https://mobify.atlassian.net/browse/WEB-922)

## Changes
- [x] Styles and markup
- ~~Engineering~~ (will be separate PR)

## How to test-drive this PR
- Run `npm run dev`
- Preview Merlin's Potions and open the Mini-Cart
- Confirm that the mini-cart is styled like the mockups


app/components/product-item/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+    alt,
+    category,
+    children,
+    className,
+    imageHeight,
+    imageWidth,
+    price,
+    src,
+    title,
+}) => {
```
DanielBrookRoberge: Feels like it might be clearer to have `imageAlt` and `imageSrc` rather than `alt` and `src`, but it might make other code clearer to leave them.

app/components/product-item/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+    const classes = classNames('c-product-item', 'u-flexbox', 'u-row-reverse', className)
+
+    return (
+        <article className={classes}>
+            <div className="u-flex">
+                {!!category &&
```
DanielBrookRoberge: In React 15, do we have any falsy values that render text other than 0? (i.e. is the `!!` necessary?)

app/components/product-item/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+                {!!children &&
+                    <div>
+                        {children}
+                    </div>
+                }
+            </div>
+            <div className="u-padding-start-lg u-padding-end-lg u-flex-none">
+                <Image src={src} alt={alt} width={imageWidth} height={imageHeight} />
+            </div>
```
DanielBrookRoberge: Should we render this even if no src is provided?

jeffkamo: a placeholder skeleton should render if the src is either invalid or missing

kschmidtdev:
I'm getting the following:
screen shot 2016-11-04 at 2 52 09 pm

jeffkamo:
@kschmidtdev fixed

app/components/product-item/index.jsx
jeffkamo: Would be better?

```
{category.length &&
    ...
}
```

DanielBrookRoberge: You should be able to do just `{category && ...}`. `category.length` will render a 0 to the page, while the only falsy things that pass proptype validation are `''` and `undefined`, both of are benign in rendering.

app/components/product-item/index.jsx
jeffkamo: I'll make that change after merging, since I need to do that to make this component work in other containers anyway.

DanielBrookRoberge: :+1:

PR #153, opened 2016-11-02
Allow components to restrict when they render

fractaltheory:
The PLP has multiple URLs that match it, so if you were to receive multiple `onPageReceived` actions, each one would update the store and render --- but we only want to render the last one. This allows for us to compare the `onPageReceived` request URL to the current URL, and only render if it matches.

 **JIRA**: [WEB-917](https://mobify.atlassian.net/browse/WEB-917)

## Changes
- Provide the request url when dispatching `onPageReceived`
- Provide the current url when dispatching `onRouteChanged`
- Update the `app` store with the current URL, in the `app` reducer for the `onRouteChanged` action
- Add a new condition inside the PLP reducer to only update the store selector when the `onRouteChanged` url matches the current url

## How to test-drive this PR
- `npm run dev`
- Quickly click multiple PLP links by opening the navigation menu and clicking any of the list items, e.g. Potions, Books, Ingredients, etc.
- Verify that this change works properly by seeing that as the responses come in, nothing renders until we receive the response for the last link clicked

### DIFF
--- a/app/app-provider.jsx
+++ b/app/app-provider.jsx
@@ -29,7 +29,7 @@ const AppProvider = ({store}) => {
 
     const getPageType = (routerState) => getComponentName(routerState.routes[1].component)
 
-    const dispatchRouteChanged = (nextState) => store.dispatch(appActions.onRouteChanged(getPageType(nextState)))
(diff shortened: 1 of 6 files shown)


DanielBrookRoberge:
Good plan and works for me here! :+1:

PR #156, opened 2016-11-03
Merlins login

Magitron:
Login page for Merlin's Potions.

## Todo
- [x] style login 
- [x] remove temporary util for form encoding in `app/utils/utils.js`

## How to test-drive this PR
- Run Merlins
- Make an account on desktop if you haven't already
- Go to login and sign in


app/containers/login/partials/form.js
```diff
@@ -0,0 +1,55 @@
...
+        <form
+            onSubmit={handleSubmit((values) => {
+                return new Promise((resolve, reject) => {
+                    submitForm(values, resolve, reject)
+                })
+            })}
+            noValidate={true}
+        >
+            {error && <div className="u-margin-bottom-md u-color-error">{error}</div>}
+            <FormFields items={items} />
```
jeffkamo: @DanielBrookRoberge are we still recommending this be used? Are we going to move away from generating components in this manner vs. being declarative? i.e...

```
<Form>
    <FieldRow><Field {...} /></FieldRow>
    <FieldRow><Field {...} /></FieldRow>
</Form>
```

app/styles/lib/_color.scss
```diff
@@ -0,0 +1,17 @@
...
+
+// Base Color
+// ---
+//
+// These are just a helper mixins to aid in creating all the below mixins
+
+
+// Neutral Colors
+// ---
+@mixin bg-color-neutral-20(
```
jeffkamo: I suppose it's okay if you want to flesh out these helpers here, but maybe add a `@TODO` to that this file needs to (and should) be migrated over to Spline pronto

app/styles/utilities/_box-shadow.scss
```diff
@@ -0,0 +1,6 @@
+// Box Shadow
+// ===
+
+.u-box-shadow {
+    box-shadow: 0 0 2px 0 rgba(0, 0, 0, 0.3);
```
jeffkamo: Good call, this is a fairly common utiliy. Please add a `@TODO` note to remind me to migrate this to Spline at a later date.

app/styles/utilities/_color.scss
```diff
@@ -44,6 +44,10 @@
 // Background Colors
 // ---
 
+.u-bg-color-neutral-20 {
+    @include bg-color-neutral-20(true);
```
jeffkamo: Same as the other utility, please add a `@TODO` reminder so I come back and migrate this to spline at a later date.

app/styles/utilities/_text.scss
```diff
@@ -108,3 +108,10 @@
 .u-text-small {
     font-size: 12px !important; // @TODO: Add to spline
 }
+
+// Font Family
+// ---
+
+.u-text-font-family {
+    font-family: $font-family;
```
jeffkamo: please add ` // @TODO: Add to spline`

app/containers/login/partials/form.js
DanielBrookRoberge: I think we're aiming for the latter; @mikenikles would know for sure.

jeffkamo: @Magitron it looks like the "Forgot your password?" and "What's this?" links are missing.

jeffkamo: Also the `*` charcters at the end of the label, and the `* Required Fields` message are missing too

mikenikles: Yes we're looking to use the declarative approach where possible.

jeffkamo: @Magitron Also missing is the "!" icons from the caption.

Make sure the error caption and field borders inherit the correct error color. Right now they are set to a value of `red` which is coming from the SDK base styles, but they should use the local `$error-color`

app/containers/login/partials/form.js
```diff
@@ -0,0 +1,55 @@
...
+    return (
+        <form
+            onSubmit={handleSubmit((values) => {
+                return new Promise((resolve, reject) => {
+                    submitForm(values, resolve, reject)
+                })
+            })}
+            noValidate={true}
+        >
+            {error && <div className="u-margin-bottom-md u-color-error">{error}</div>}
```
Magitron: @guisers

app/containers/login/partials/form.js
Magitron: @guisers

app/components/field/_theme.scss
```diff
@@ -1,17 +1,33 @@
...
+    &.c--error {
+        .c-field__label {
+            color: $font-color; // 1
+        }
+    }
+}
+
+.c-field + .c-field {
+    margin-top: $unit*2;
+}
```
Magitron: Makes sense, except that we aren't using field rows in this case. @guisers

guisers: We aren't explicitly using `FieldRow`. I'm actually using our `FormFields` component which blackboxes all the field stuff. I just looked at the source code though and this component _does_ create `FieldRow` for me so it should be there

app/containers/login/partials/form.js
guisers: @jeffkamo that's so weird, this is what it does on mine:
image

app/components/field/_theme.scss
guisers: Oh I see this is talked about later in the PR. I was reading this chronologically. Will refactor this the declarative way now 😅

app/containers/login/partials/form.js
```diff
@@ -0,0 +1,55 @@
...
+            <FormFields items={items} />
+            <button className="c-button c--primary u-width-full u-margin-top-lg" type="submit" disabled={submitting}>{submitText}</button>
```
guisers: @Magitron wanna do this w/css?

app/containers/login/partials/form.js
guisers: Sounds good. Might be worth adding a note in the docs if this is the case

app/components/field/_theme.scss
jeffkamo: Thanks @guisers!

@Magitron now that the FieldRow component is being used, this `field + field` style can probably be removed.

app/containers/login/partials/form.js
jeffkamo: there's a utility class you can use: `u-text-uppercase` I believe is the one

guisers:
@jeffkamo Could you please check if the PR is meeting your requested changes? I _think_ we have addressed everything. :)

PR #158, opened 2016-11-04
Mini-cart functionality

kschmidtdev:
Implement populating the mini cart components with cart data. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-922
 **Linked PRs**: #151

## Changes
- Fix up and use the `cart` actions and parser to get cart contents
- Hook up mini-cart rendering

## How to test-drive this PR
- Open the normal Merlins website and run as a mobile website - add something to your cart
- Preview Merlins with this PR
- Observe the contents of the mini cart are accurate

Remaining TODOs:
- [x] Figure out what to do about a cart button counter badge (if we want one or not)
- [x] Figure out what to do about rendering when the cart is empty
- [x] Figure out what to do about cart image size
- [x] Figure out what to do about cart item subtotal
- [x] Test when you have multiple items of the same type in your cart


jeffkamo:
Heads up @kschmidtdev: I just posted a PR that adds styles/layout for the mini-cart's empty state.

app/containers/cart/actions.js
```diff
@@ -15,10 +15,17 @@ const baseHeaders = {
  * Get the contents of the users cart
  */
 export const getCart = () => {
-    const opts = {headers: baseHeaders}
+    const opts = {
+        headers: baseHeaders,
+        credentials: 'same-origin'
+    }
     return fetch('http://www.merlinspotions.com/customer/section/load/?sections=cart', opts)
```
kschmidtdev: See if we can grab this URL dynamically

app/containers/mini-cart/reducer.js
```diff
@@ -7,16 +7,10 @@ const initialState = Immutable.fromJS({
...
-    // },
-    // [miniCartActions.receiveResponse]: (state) => {
-    //     return state
-    // },
-    // [miniCartActions.fetchContents]: (state) => {
-    //     return state
-    // }
+    [miniCartActions.receiveContents]: (state, payload) => {
+        return state.set('contentsLoaded', true)
+                    .mergeDeep(payload)
```
kschmidtdev: can pass multiple parameters to mergeDeep

app/containers/cart/actions.js
```diff
@@ -15,10 +15,17 @@ const baseHeaders = {
...
-        .text()
-        .then(parse)
+        .then((response) => {
```
kschmidtdev: `.then(parse).then(dispatch)` - dispatch a function with the result, and shorten chain

app/containers/mini-cart/actions.js
```diff
@@ -1,24 +1,16 @@
...
-//     }
-// }
-//
-// export const fetchContents = () => {
-//     return (dispatch) => {
-//         makeRequest(window.location.href)
-//             .then((response) => dispatch(receiveResponse(response)))
-//     }
-// }
+export const fetchContents = () => {
```
kschmidtdev: this could take the action to dispatch as a param

app/containers/mini-cart/container.js
```diff
@@ -12,12 +12,20 @@ import {HeaderBar, HeaderBarActions, HeaderBarTitle} from 'progressive-web-sdk/d
...
     }
 
     render() {
-        const {miniCart, closeMiniCart} = this.props
-        const isOpen = miniCart.get('isOpen')
+        const {contentsLoaded} = this.props
+
+        if (!contentsLoaded) {
+            return (
+                <div />
```
kschmidtdev: can `return false` here if content is meant to be empty

app/containers/mini-cart/container.js
```diff
@@ -12,12 +12,20 @@ import {HeaderBar, HeaderBarActions, HeaderBarTitle} from 'progressive-web-sdk/d
...
 class MiniCart extends React.Component {
     componentDidMount() {
-        // this.props.fetchContents()
+        this.props.fetchContents()
     }
 
     render() {
-        const {miniCart, closeMiniCart} = this.props
-        const isOpen = miniCart.get('isOpen')
+        const {contentsLoaded} = this.props
```
kschmidtdev: might as well object destructure all at once here

app/containers/mini-cart/container.js
```diff
@@ -113,25 +97,32 @@ class MiniCart extends React.Component {
...
 
 MiniCart.defaultProps = {
-    // contentsLoaded: false,
+    contentsLoaded: false,
 }
 
 const mapStateToProps = (state) => {
     return {
-        miniCart: state.miniCart
+        ...state.miniCart.toJS()
```
kschmidtdev: use immutableJS here and call toJS() in render

app/containers/mini-cart/container.js
```diff
@@ -31,61 +37,37 @@ class MiniCart extends React.Component {
...
+                    {cart.items.map((product, idx) => {
+                        return (
+                            <ProductItem
+                                className="u-padding-top-lg u-padding-bottom-lg"
+                                category="Potions"
+                                title={<h2 className="u-h3">{product.product_name}</h2>}
+                                price={product.product_price}
+                                src={product.product_image.src}
+                                alt={product.product_image.alt}
+                                imageWidth="64px"
```
kschmidtdev: @jeffkamo Should this still be here? The cart payload gives me image size info but since this is in the mini cart we probably want them to all be the same size - thoughts?

jeffkamo: I'd say we should control the size of the images on our side. The size they use on desktop may or may not fit in our design.

kschmidtdev: Should this value (`64px`) be in a stylesheet file instead?

jeffkamo: It's generally preferred to apply width dimensions directly to image tags rather than CSS. Makes things look better when internet connections are slow (and therefor CSS not yet loaded)

kschmidtdev: Makes sense! Thanks!

app/containers/mini-cart/container.js
```diff
@@ -10,14 +11,27 @@ import List from 'progressive-web-sdk/dist/components/list'
...
+export const productSubtotal = (price, quantity) => {
+    const priceInCents = price.replace(/[$,. ]/g, '')
+    const priceNumber = parseFloat(priceInCents) / 100
+    return (priceNumber * quantity).toFixed(2)
+}
+
 class MiniCart extends React.Component {
     componentDidMount() {
-        // this.props.fetchContents()
+        this.props.fetchContents()
```
fractaltheory: This will fetch the cart contents only when `MiniCart` component is first mounted, which means this only occurs once per application start (i.e. when the parent component is mounted, in this case, `App`). I expect you'll want to also dispatch this action after the user adds to cart (or perhaps when the user opens the mini-cart).

app/containers/mini-cart/actions.js
```diff
@@ -1,24 +1,4 @@
 import {createAction} from '../../utils/utils'
-// import {jqueryResponse} from 'progressive-web-sdk/dist/jquery-response'
-// import miniCartParser from './mini-cart-parser'
-
-// export const receiveContents = createAction('Received MiniCart Contents')
 
 export const openMiniCart = createAction('Open navigation')
```
fractaltheory: You might want to change these descriptor strings.

app/containers/mini-cart/container.js
kschmidtdev: I am planning to add this in an upcoming PR :)

fractaltheory:
👍

app/containers/mini-cart/container.js
```diff
@@ -110,8 +99,14 @@ class MiniCart extends React.Component {
 
     render() {
         const {miniCart, closeMiniCart} = this.props
-        const isOpen = miniCart.get('isOpen')
-        const isEmpty = false // @TODO replace with actual functionality
+
+        const {cart, contentsLoaded, isOpen} = miniCart.toJS()
+
+        if (!contentsLoaded) {
```
fractaltheory: I'm not sure how I feel about this - I think the mini-cart should display something even while the contents are being loaded.

fractaltheory:
👍

PR #160, opened 2016-11-12
Readme updates

mikenikles:
A few updates to the readme to sync it with how we develop on the Progressive Web team.

## Changes
- See commit messages.

## How to test-drive this PR
- Review the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 - [Git](https://git-scm.com/)
 - We recommend you use [nvm](https://github.com/creationix/nvm#installation) to
 manage node and npm versions.
-- node 4.x or greater
-- npm 2.x or greater
+- node 6.x LTS
+- npm 3.x
 
 ## Setup
 
--- a/README.md
+++ b/README.md
@@ -19,12 +19,12 @@ npm run dev
 npm run add:page
 ```
 
-## Styleguide
+## Docs with Styleguide
 
-To run the project styleguide use:
+To run the project documentation, including a live styleguide, use:
 
 ```
-npm run styleguide
+npm run docs:dev
 ```
 
 ## Tests and test coverage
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,8 +3,8 @@
 - [Git](https://git-scm.com/)
 - We recommend you use [nvm](https://github.com/creationix/nvm#installation) to
 manage node and npm versions.
-- node 4.x or greater
-- npm 2.x or greater
+- node 6.x LTS
```
mikenikles: @jansepar you mentioned this may cause issues when we build Apaptive and Progressive Web projects at the same time.

Since the Node community moved to version 6.9.x as the long-term support, can we deal with a workaround internally?

Our documentation no longer works with node 4 / npm 2. I prefer to deal with Adaptive as a special case rather than holding Progressive back with a node upgrade.

jansepar: So this doesn't affect Lancôme if this is a scaffold change, but in Lancôme I've changed 'npm run dev' to do the adaptive build first. We could easily switch the instructions to not build it every time you run that command, but instead run it beforehand. Shouldn't be a big deal - Adaptive is out, Progressive is in.

README.md
```diff
@@ -19,20 +19,20 @@ npm run dev
...
 
-To run the project styleguide use:
+To run the project documentation, including a live styleguide, use:
 
 ```
-npm run styleguide
+npm run docs:dev
 ```
 
 ## Tests and test coverage
```
DanielBrookRoberge: IIRC the scaffold doesn't run coverage in the `test:all` script.

mikenikles: You're right, I'll fix that.

README.md
mikenikles: That makes sense, thanks for the clarification.

DanielBrookRoberge:
:+1: :shipit:

PR #168, opened 2016-11-16
Add a flag to opt you into auto fetching the HTML

jansepar:
**JIRA**: N/A

## Changes
- Add a `autoFetchHTMLFromOriginPage` flag to opt you into auto fetching the HTML, instead of opting you in by default

## How to test-drive this PR
- Run the scaffold, make sure the page is still fetched.
- Change `autoFetchHTMLFromOriginPage` from `true` to `false`, and make sure the page _isn't_ fetched.

## Background info

When I initially jumped into the scaffold when writing some documentation, it took me a very long time to understand where the actual page was being fetched for each of our routes. I eventually realized that this was happening "automagically" in app-provider.

To me, this goes against something in our developer values, ["explicit is better than implicit"](https://github.com/mobify/developer-values#readability-counts).

To me, this should be self documenting, hence why I've made the flag name very explicit, and also made it so that you have to set it on each route - it will force developers to take notice when adding new routes.

Personally, I don't believe this PR goes far enough - I think people should be able to go into their container and immediately understand where the data is coming from. At least with this change, people will see it pretty quickly when heading to the router.

If we want to provide easier mechanisms to auto-fetch the HTML, I think we could do that in a way that makes it much more clear in the containers. The question is, should we even bother with this intermediate step, or should we go straight to the containers?


app/app-provider.jsx
```diff
@@ -37,31 +37,34 @@ const AppProvider = ({store}) => {
     const onEnter = (nextState) => {
         triggerMobifyPageView(nextState.routes[1].routeName)
         dispatchRouteChanged(nextState)
-        dispatchFetchPage(nextState)
+        if (nextState.routes[1].autoFetchHTMLFromOriginPage) {
```
mikenikles: I'd rather not add that as it feels like a step back to where we'd like to go. On every page change, we want to fetch page content. Currently, it's an XHR request to the current URL, later we want to support REST requests.

The change we want to make is near https://github.com/mobify/progressive-web-scaffold/blob/develop/app/containers/app/actions.js#L35. That's where I imagine the decision to happen as to whether we load HTML, JSON, etc.

mikenikles:
Closing this in favour of what we're going to build in the current sprint.

PR #170, opened 2016-11-17
Feedback: Move PW Component Styles

jeffkamo:
Feedback we received during the Partner Training was that having both the local components and SDK component themes in the same location was confusing. It was suggested that we move the PW component styles to somewhere different to help clarify the difference.

## Changes
- Move all PW-SDK component styles to `app/styles/themes/pw` directory
- Group the node module base style and local theme style imports together in pairs for each component, instead of in two separate groups.

**Before**

```
/app
    /components
        /[local-component]
            _base.scss
        /[pw-component]
            _theme.scss
```

**After**

```
/app
    /components
        /[local-component]
            _base.scss
    /styles
        /themes
            _pw-components.scss
            /pw-components
                _[pw-component].scss
```

**Rationale*

My thought here is to not create a PW directory for just components at the `/app` or `app/components/` level, and leverage the  scaffold's theme folder structure instead to style the PW-SDK components. This folder structure is more or less identical of the project had some other form of theming.

## How to test-drive this PR
- run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Confirm that all the progressive web component styles work as intended still (nothing broke)
- Consider whether this folder structure is a good idea or not (speak up if not!)


marlowpayne:
Is "PW" an already-established namespace for SDK things? Seems a little unclear to me, since technically _everything_ in a project is Progressive Web / PWA related. If it's meant to indicate SDK stuff, maybe "SDK" / `sdk/` namespacing would be clearer?

jeffkamo:
@marlowpayne good question. I only chose `pw` because that's what we chose for the class namespace: `pw-`

jeffkamo:
Previously, we didn't go with "SDK" as that felt too generic and could mean any sdk. We didn't choose PWSDK because that just seemed too long and ugly.

fractaltheory:
Looks good, however I think I agree that `pw` doesn't really convey much information. What about something like `pw-components` ?

jeffkamo:
@marlowpayne @fractaltheory changed `pw` > `pw-components`

fractaltheory:
👍

PR #171, opened 2016-11-17
Clean up dispatch

DanielBrookRoberge:
This PR brings the various dispatch functions in the scaffold up to current best practice, namely:
 - simplifying `mapDispatchToProps` by using a plain object instead of a function
 - removing unneeded exports on `mapStateToProps` and `mapDispatchToProps`

 **JIRA**: N/A

## Changes
- Clean up containers

## How to test-drive this PR
- Preview
- See that things still work.


fractaltheory:
One negative about not using anonymous function wrappers in `mapDispatchToProps` is that callbacks like `onClick` pass the React synthetic `Event` argument directly to the action, which results in a bunch of access warnings:

screen shot 2016-11-17 at 2 38 12 pm

DanielBrookRoberge:
Yeah that's really annoying, actually. Probably this deserves some discussion.

mikenikles:
Do we know where exactly that warning gets thrown? Based on the documentation, we should be able to omit it by calling `event.persist()`.

> If you want to access the event properties in an asynchronous way, you should call event.persist() on the event, which will remove the synthetic event from the pool and allow references to the event to be retained by user code.

DanielBrookRoberge:
I'm thinking this is a situation where we have
```javascript
const mapDispatchToProps = {
    openPopup: containerActions.openPopup
}

const Component = ({openPopup}) => (
    <button onClick={openPopup}>Open Popup</button>
)
```
which doesn't really give us a situation to call `event.persist()`, especially if `openPopup` isn't expecting an event or always going to receive one. 

I suppose we could protect it from the event with
```javascript
const mapDispatchToProps = {
    openPopup: () => containerActions.openPopup()
}
```
but that seems awkward.

fractaltheory:
The thing is that we don't need to persist the event. It's the passing along of the entire Synthetic Event that's causing an issue.

DBR's second suggestion above is the workaround I found to work best.

DanielBrookRoberge:
Ok I've written a utility to do the workaround for that second suggestion. (so that it's better self-documenting) in https://github.com/mobify/progressive-web-scaffold/pull/171/commits/9ece71202694d4a81f529c8c0ff7606420ac194e

PR #172, opened 2016-11-18
Fix Flash of Unstyled Stuff

jeffkamo:
Before this change, there was a chance that the site's content would display before our CSS loaded. This meant a frame (or worse, many seconds) of massive black and white SVG images, unstyled text, etc. This PR addresses the issue by waiting for the CSS to finish loading before we hide the Preloader.

## Changes
- Hide the app via inline CSS by default
- The CSS unhides the app as soon as it's loaded
- Hide the Preloader only _after_ the CSS is loaded

## How to test-drive this PR
- run `npm run dev`
- In loader.js, replace `getAsset('main.css')` with `href: 'http://www.deelay.me/10000/https://dl.dropboxusercontent.com/u/4782540/test.css'` (delays the CSS for 10 seconds before actually loading)
- Preview Merlin's Potions
- Confirm that the Preloader stays in place for the entire time we wait for the CSS to load
- Confirm that there are no frames where the content appears unstyled


fractaltheory:
Nice fix! I could see us supplying the inline style via prop and controlling it through redux, but in my opinion that just adds needless complexity for the sake of using redux.

One suggestion is to namespace this to `window.Progressive` which we already set up in the loader, i.e. `window.Progressive.stylesheetLoaded`

kschmidtdev:
👍 works and all the changes seem reasonable - but would prefer someone else to also review and 👍 other than myself 😃

fractaltheory:
Also might want to leave a short comment in `loader.js` next to the `onload` callback, explaining why that exists.

DanielBrookRoberge:
I'm :+1: on this though I agree about MQ's comment on namespacing the var.

PR #174, opened 2016-11-22
"Cache" data in the redux store, keyed to their URLs

guisers:
A pattern where slices of state "cache" data keyed to URLs. The currently routed URL is determined by a `selector` value in that slice of state. Skeleton data is keyed to `placeholder` for use during transitions. Also allow PDP to listen to PLP page received to prepopulate some basic data.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-994

## Changes
- Extract some URL-keyed state caching behavior into utility functions
- Refactor PLP to use these utilities
- Implement it in PDP
- Improve PDP propType validation
- Prepopulate PDP with information parsed from PLPs ("shared state")

## How to test-drive this PR
- Visit various PLP and PDPs. 
- Go back to previously visited pages.
- Notice that they load almost instantaneously, due to their data already being in the cache!


guisers:
@fractaltheory Hey, this is my attempt at pulling out your PLP routing logic into something reusable. It's not 100% yet but wanted to get your feedback on the road I've taken before I keep going with it 👀

app/containers/pdp/container.js
```diff
@@ -68,14 +64,62 @@ class PDP extends React.Component {
...
+    quantityAdded: PropTypes.number
 
+}
 
-export const mapStateToProps = ({pdp}) => ({pdp})
+export const mapStateToProps = (state) => {
+    const routedState = getRoutedState(state.pdp)
+    return {
+        routedState,
+        ...routedState.toJS()
```
jeffkamo: I could be mistaken, but I could have sworn this was determined to be an anti-pattern? @DanielBrookRoberge or @fractaltheory do you recall if it was decided whether `toJS()` should be called here or in the elsewhere?

DanielBrookRoberge: Yeah, this is a fairly new development but the new recommendation is to do the `.toJS()` in the render method rather than `mapStateToProps` to avoid passing redundant data in props.

app/utils/router-utils.js
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Determine whether we should display content or transition into a placeholder state
+ */
+export const getNextSelector = (state, currentURL) => { return state.has(currentURL) ? currentURL : PLACEHOLDER }
+
+
+/**
+ * Takes a slice of state and returns the key of the current route
+ */
+export const getSelectorFromState = (state) => {
```
fractaltheory: Nitpick: could rewrite this shorter as `(state) => state.get(SELECTOR)`

app/utils/router-utils.js
```diff
@@ -0,0 +1,65 @@
...
+    return state.get(SELECTOR)
+}
+
+/**
+ * Takes a slice of state and returns the data mapped to the current route
+ *
+ * eg. plp > 'http://www.merlinspotions.com/potions.html' > {data}
+ */
+export const getRoutedState = (state) => {
```
fractaltheory: Same thing here, as above

app/containers/pdp/reducer.js
```diff
@@ -1,47 +1,82 @@
...
 
-const initialState = Immutable.fromJS({
-    contentsLoaded: false,
+import PLP from '../plp/container'
+import {basicPlpParser} from './parsers/basic-plp'
+
+import {onPageReceived, onRouteChanged} from '../app/actions'
+import {SELECTOR, PLACEHOLDER} from '../app/constants'
+
+export const initialState = Immutable.Map({
```
fractaltheory: I believe you could rewrite this as `Immutable.fromJS` and it'll convert deeply any structures to Immutable objects (i.e. instead of having to set `product: Immutable.Map`

fractaltheory:
I think I've had a good look through so far - resolve the comments above and I can take another look 👀

guisers:
@fractaltheory ok this is ready for review again :D

app/containers/pdp/reducer.test.js
```diff
@@ -1,12 +1,16 @@
...
 import * as pdpActions from './actions'
 
-jest.mock('./parsers/pdp')
-import pdpParser from './parsers/pdp'
+import {jquerifyHtmlFile} from 'progressive-web-sdk/dist/test-utils'
+import {SELECTOR, PLACEHOLDER} from '../app/constants'
+import {baseInitialState} from '../../utils/router-utils'
+
+const untouchedState = baseInitialState.set(PLACEHOLDER, initialState)
+const $content = jquerifyHtmlFile('app/containers/pdp/parsers/pdp-example.html')
```
DanielBrookRoberge: Ideally, we'd like to have the reducer tests not depend on the parsers at all; this will help when we introduce the middleware layer and the reducers need to be parser-independent. Can you think of a way to avoid this?

guisers: I don't think we are using the parser here; it's just an html file from the `/parsers` directory

app/containers/pdp/container.js
```diff
@@ -69,14 +67,28 @@ class PDP extends React.Component {
...
+     */
+    setQuantity: PropTypes.func.isRequired,
 }
 
-
-const mapStateToProps = ({pdp}) => ({pdp})
+export const mapStateToProps = (state) => {
+    const routedState = getRoutedState(state.pdp)
+    return {routedState}
+}
```
guisers: nice!! es6 is so beautiful 😍

app/containers/pdp/parsers/basic-plp.js
```diff
@@ -0,0 +1,28 @@
...
+                    {
+                        img: img.src,
+                        position: '0'
+                    }
+                ]
+            }),
+            contentsLoaded: false,
+            isPlaceholder: false
+        })
+    })
```
guisers: After playing around with it I decided to keep everything done here as JS, and I do a `fromJS()` conversion once with the entire result when this parser is actually used (in the reducer). Is that ok? 🤔

app/containers/pdp/reducer.test.js
DanielBrookRoberge: I meant more that the reducer is calling the parser in these tests, rather than having it mocked out, so they depend on the details of the parser and input rather than being independent. I'm not sure this is a priority to fix here, though.

guisers: Ohhhh gotcha. I think I'll have some spare time tomorrow so I'll see what can be done

app/containers/pdp/parsers/basic-plp.js
```diff
@@ -0,0 +1,29 @@
+import {parseTextLink, parseImage} from '../../../utils/parser-utils'
+import {initialState} from '../reducer'
+
+export const basicPlpParser = ($, $html) => {
+    const $products = $html.find('.item.product-item')
+    const productMap = {}
+    $products.each((_, product) => {
```
jansepar: Trying to understand where we get the array of products that we need as defined by here:

https://github.com/mobify/progressive-web-scaffold/pull/174/files#diff-e1cdd7f0a5d742baac35cac6c1188dd2R16

Could you help me trace this down?

jansepar: HOKAY, so @fractaltheory guided me through what is going on here. I didn't realize this was a parser in the `pdp`!

So it sounds like what we're doing is duplicating this data for storage inside of the PDP. This is different from how we had done it in Pure Formulas. In PF, we wanted to reduce as much duplication as possible by sharing state across the PLP and PDP.

The reason being is that duplicate data could potentially be confusing for someone who is looking to build a project that might have data in their PLP/PDP (take for example, a rating). They look at our build and think "oh wow, the PDP renders instantly, this means the data must be shared". Imagine a scenario where they tweak their PLP parser to grab the rating, and then go attempt to use that new `.rating` key in the PDP, and it fails and confusion ensues.

@fractaltheory has a different thought on the matter - from his perspective, it's easiest to just simply know that data for the PDP is within it's own container, and data for the PLP is in it's own container.

So this is something we should probably get some further feedback on :)

For some background, the PF code for this is [here](https://github.com/mobify/pure-formulas-mobile/blob/develop/react/containers/pdp/container.js#L9). Note that it's done in a O(n) way (could have sworn it was different...). Ideally if we're reducing data duplication and want fast lookups, we'd do something like this:

```
// plp page
plp =  { 'potions.html': [1,2,3,4], 'books': [5,6,7,8] }

pdp: {
    1: {title: 'Eye of Newt', ...},
    2: {title: 'Unicorn Blood', ...}
    etc...
}
```

jansepar:
I noticed that the product carousel flashes when the response comes in for the extra product info. Something we should fix.

app/containers/pdp/parsers/basic-plp.js
jeffkamo: Would it be viable to make a Product container? A container that is not a template but is a way of abstracting that shared part of the redux store without letting the different containers have true access to each other's stores?

```
// plp page
plp: {
    'potions': [1, 2, 3, 4],
    'books': [5, 6, 7, 8],
    ...
}

// pdp page
pdp: {
    product: 1,
    ...
}

// products container
products: {
    1: {title: 'Eye of Newt', ...},
    2: {title: 'Unicorn Blood', ...},
    ...
}
```

```jsx
// PDP example
<PDP>
    <Products get={plp.product}>
        {(product) => (
            <div>
                <h1>{product.title}</h1>
                ...
            </div>
        )}
    </Products>
</PDP>

// PLP example
<PLP>
    <Products get={plp.list}>
        {(products) => products.map((props) => (
            <ProductTile {...props} />
        ))}
    </Products>
</PLP>
```

Not sure if this is an unreasonable level of abstraction ¯\\\_(ツ)\_/¯

mikenikles: While the PF code has a single source of truth and stores products in a `plp` section of the Redux state, this to me is equally confusing since the PDP needs to access that `plp` section.

What I have in mind is more along the lines of Jeff's suggestion. We design the Redux state similar to this:

```json
{
  "catalog": {
    "products": []
  }
}
```

Regardless of where in the application we need access to products, they're accessed the same way.

guisers: Thanks @jansepar for reviewing! Sorry I didn't get a chance to look at this til this morning. Anyway, yeah as @fractaltheory said, we decided at the 30% feedback mark to let PLP and PDP mind their own business for compartmentalization's sake. I think the reason was that one thing that's different from the Merlin's implementation of this vs the PF implementation is that we actually save state for every visited PLP as well. Eg. if I've hit the Books page before, I get instant load when i go back. Whereas in PF, we store the `productList` and that's it.

That being said, I really like the catalog approach suggested by @jeffkamo and @mikenikles above. Also it doesn't seem important to store all the PLPs at all if we are able to store their individual products elsewhere... they are composed mostly of that product info anyway. 

If we can come to a consensus on the approach we want to take, I have time to work on it this week! 🌮

mikenikles: @guisers One of my tasks early this week is to create a first draft of that schema. I'll get the team's input once it's ready so that you can work off of it for the refactoring.

I'm ok if you merge this PR and do the refactoring in a new PR. Please address Shawn's comment first though, about the product carousel that flashes.

guisers: @mikenikles  sounds good

jansepar: This approach sounds good to me 👍 This change looks great, excited to see it live. Solutions Engineers will be very thankful :)

guisers:
@jansepar @jeffkamo 
The Carousel flash is solved at the component level in this SDK PR here: https://github.com/mobify/progressive-web-sdk/pull/413

app/containers/pdp/_pdp.scss
```diff
@@ -49,3 +49,7 @@
 
     fill: $neutral-40;
 }
+
+.t-pdp__carousel .c-image__transition {
```
jeffkamo: This customization of the image transition should probably be part of the image component's theme styles so it can be reused elsewhere.

I also have some comments about the classname `c-image__transition` itself in the SDK PR.

guisers: Cool! I fixed this

guisers:
The last blocker for this PR, https://github.com/mobify/progressive-web-sdk/pull/413, has been merged so merging and closing this one as per discussion above and starting a new PR.

PR #175, opened 2016-11-22
Service worker for Merlins - Phase 1

DanielBrookRoberge:
We need a service worker for Merlin's Potions to increase performance, improve our Lighthouse score, and enable offline mode and add to homescreen on Chrome. This PR will add one, along with the machinery needed to load it.

Functionality:
 - A 'bundle cache' for bundle contents, that prevents requests from going onto the network.
 - An image cache, for the 40 most recent images loaded. These images are served by the cache but updated from the network in the background.
 - A general cache for all other content, with an expiration of 24 hours for content.

The current PR only works for local preview, production support will be added in a follow-up.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-991
 **Linked PRs**: #193 

## Changes
- Add a service worker build configuration
- Add a service worker file
- Load the worker in the app initialization
- Enable caches

## How to test-drive this PR
- Use Charles to map the file `service-worker-loader.js` to `https://www.merlinspotions.com/service-worker-loader.js`.
- Load the site and see that a service worker is available in the Application tag of dev tools
- Reload the page with the Network tab open and see the lack of network requests for bundle resources
- Look in the Application tab at the Browser Cache, and see the multiple caches managed by the worker.

### DIFF
--- a/app/loader.js
+++ b/app/loader.js
@@ -13,10 +13,17 @@ const isReactRoute = () => {
 initCacheManifest(cacheHashManifest)
 
 const CAPTURING_CDN = '//cdn.mobify.com/capturejs/capture-latest.min.js'
+const SW_LOADER_PATH = `/service-worker-loader.js?preview=true&b=${cacheHashManifest.buildDate}`
 
 import preloadHTML from 'raw!./preloader/preload.html'
 import preloadCSS from 'css?minimize!./preloader/preload.css'
 import preloadJS from 'raw!./preloader/preload.js' // eslint-disable-line import/default
+
+const loadWorker = () => (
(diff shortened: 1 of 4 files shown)


webpack/dev.js
```diff
@@ -23,4 +24,10 @@ mainConfig.plugins = mainConfig.plugins.concat([
     new webpack.NoErrorsPlugin()
 ])
 
-module.exports = [mainConfig, loaderConfig]
+workerConfig.plugins = workerConfig.plugins.concat([
+    new webpack.DefinePlugin({
+        DEBUG: true
```
jansepar: Are we missing a `webpack/prod.js` in this PR?

PR #184, opened 2016-11-24
Refactor Home & Skeletons

jeffkamo:
Refactor the home page's skeletons (and a few other things) so the page loads more seamlessly.

## Changes
- Refactor home container and partials
- Clean up spacing and dimensions so when skeletons disappear, the page doesn't shift
- Clean up the home page skeletons to appear less busy (closer to the final content)
- Tweak the home page CSS and markup a bit
- Change the carousel theme selectors to use `pw-` namespace instead of `c-`

| Before | After |
| ------ | ------ |
| skeletongs-before | skeletongs-after |

> Notice how the content below the carousel shifts down after the content loads!

## How to test-drive this PR
- Run `npm run dev`
- [Preview Merlin's Potions](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- On the homepage, confirm that the skeletons to real-content transition is smooth


jansepar:
The animation looks a bit choppy but maybe that's just the gif? Could you show me this at your desk?

jeffkamo:
@jansepar Ya that's just the gif's low frame rate + compression. Let me know and I can show you it running when you have a minute.

kschmidtdev:
Chrome network throttling makes this pretty easy to test locally 😄

app/containers/home/partials/home-carousel.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+// The ratio of the banner image width:height is 1:.84. Since the banner will be
+// width=100%, we can use 84vw to predict the banner height.
+const imageHeight = '84vw'
+const placeholder = <SkeletonBlock height={imageHeight} />
+
+const HomeCarousel = ({banners}) => {
+    return (
+        <div className="t-home__carousel">
+            {banners ?
```
DanielBrookRoberge: Given that the propTypes require this variable, does it make sense to have this conditional here?

app/containers/home/partials/home-categories.jsx
```diff
@@ -0,0 +1,68 @@
...
+        <Image
+            src={getAssetUrl(`static/img/categories/${text.trim().toLowerCase()}@2x.png`)}
+            alt={text}
+            height="60px"
+            width="60px"
+        />
+    ) : (
+        <SkeletonBlock height="60px" width="60px" />
+    )
+}
```
DanielBrookRoberge: In the context of this function, is `text` still the best parameter name?

app/containers/home/reducer.js
```diff
@@ -5,7 +5,7 @@ import {onPageReceived} from '../app/actions'
 import homeParser from './parsers/home'
 
 const initialState = fromJS({
-    categories: ['', '', '', ''],
+    categories: ['', '', '', '', '', ''],
```
DanielBrookRoberge: This seems to be getting a little awkward. One replacement could be `new Array(CATEGORY_COUNT).fill('')`.

app/containers/home/partials/home-carousel.jsx
jeffkamo: it is needed for while we're waiting to receive the full array of images. I'll tweak this ternary operator to make more sense.

app/containers/home/partials/home-categories.jsx
jeffkamo: Changed to `alt`, since that's what it's being passed into.

PR #185, opened 2016-11-24
Improve preview SSL

DanielBrookRoberge:
The certificate shipped with version 1.14.1 of `webpack-dev-server` is expired, as of March this year. This PR updates to a newer version with a non-expired cert.

With this PR in place, the `node_modules/webpack-dev-server/ssl/server.crt` certificate can be added to the trust store of a Mac to avoid having to re-confirm SSL exceptions for any progressive build with this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,26 @@ To run tests in watch mode and only run test related to files you modify during 
 npm run test:watch
 ```
 
+## Prevent SSL Errors in Preview (on a Mac)
+
+The development server uses a self-signed SSL certificate which is
+valid, but treated as suspect by browsers. This means that we must
+create and reconfirm security exceptions for it, and avoid localhost
+for certain use cases (such as service workers).
+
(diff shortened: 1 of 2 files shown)


kschmidtdev:
This works! These are the steps I used to set up not being required to re-confirm SSL exceptions (maybe we want these added to the README?)

1. ```open node_modules/webpack-dev-server/ssl/server.crt```
2. Open `Keychain Access` -> go to `Certificates`, find `localhost`:
screen shot 2016-11-24 at 3 19 28 pm
4. Right-click on it and click `Get Info`
5. Expand the `Trust` section
6. Set `Secure Socket Layer (SSL) to `Always Trust`
7. Close the window (it will ask you to authenticate yourself)

👍

DanielBrookRoberge:
Yeah, adding that to the readme is a good idea. Thanks for putting that together!

PR #187, opened 2016-11-25
Login component (sign in + register actions)

sap1ens:
Refactoring Login page component to include registration form. This was tricky, because use-case is not typical: you have multiple "backends" for one container, so you need to support multiple parsers and forms, but also keep the state in one reducer.

That's why I had to implement some additional routing, as well as app provider changes. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-979

## Changes
- new parser, form and associated tests for registration page
- minor tweaks to app-level provider and action
- refactored login component to include new registration stuff

## How to test-drive this PR
- `npm run dev`
- open Preview 
- go to the Login page
- toggle between "Sign In" and "Register" sections
- 💰 profit


app/containers/login/actions.js
```diff
@@ -23,39 +23,116 @@ const validateForm = (formValues) => {
...
+            _error: 'Please fill in the form'
+        }
+    }
+
+    const {
+        firstname,
+        lastname,
+        email,
+        password,
+        password_confirmation: passwordConfirmation
```
kschmidtdev: TIL you can destructure to a different variable name 💡

app/containers/login/actions.js
```diff
@@ -13,7 +13,7 @@ const validateForm = (formValues) => {
     const email = formValues.login.username
     if (!email) {
         errors.login.username = 'Email address is required'
-    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(email)) {
+    } else if (!isEmail(email)) {
```
kschmidtdev: I believe we are already bringing in the npm package 'validator' which has an email validation function, maybe that would work for this?

`import {isEmail} from 'validator'`

app/containers/reducers.js
```diff
@@ -27,6 +27,7 @@ const rootReducer = combineReducers({
     pdp,
     plp,
     form: formReducer,
+
```
sap1ens: I was going to, but then I saw `// THIS IS A GENERATED FILE, DO NOT EDIT` message at the top of the file :) 

I'll remove it

kschmidtdev: Oh right...bad code generator!

sap1ens: Fixed

app/containers/login/actions.js
sap1ens: Fixed

app/containers/login/partials/register.js
```diff
@@ -0,0 +1,126 @@
...
+                                        name={name}
+                                        label={labelNode}
+                                        component={FieldComponent} // Progressive Web SDK Field Component
+                                    >
+                                        <input type={type} />
+                                    </Field>
+
+                                    {tooltip &&
+                                    (<div>
+                                        <a href="#remember-me" onClick={openModal}>{tooltip.title}</a>
```
sap1ens: Fixed in https://github.com/mobify/progressive-web-scaffold/pull/187/commits/0f4143f3b5cd1a1d3c875cdb231d62f801917afa 

Happy to see more elegant way ;-)

sap1ens:
From @fractaltheory (in Slack):

```
Michael Quan [3:26 PM]  
I find that if I attempt to create an account using mobifyqa@gmail.com / p4ssword
Michael Quan [3:26]  
I’m sent back to the registration page with no error messages
Michael Quan [3:26]  
but I do see a request made to captcha.html - but it’s not shown anywhere on desktop
Michael Quan [3:29]  
unsure if that’s anything to do with it though
Michael Quan [3:29]  
@sap1ens okay I think it has more to do with attempting to register with a previously registered account
```

So, I checked that. Form actually does have an error: 

screen shot 2016-11-27 at 7 31 02 pm

It's very generic, because Magento uses JavaScript-based validation. So we can only detect if redirect happened to the dashboard (success) or back to the form (failure).

sap1ens:
All issues are resolved, should be good to review again.

fractaltheory:
👍

sap1ens:
Thanks for great review 👍

PR #188, opened 2016-11-25
Testing a lighthouse integration...

guisers:
**JIRA:** https://mobify.atlassian.net/browse/WEB-996

Testing running CircleCI with Lighthouse on PR push. 

So far, we are only able to run Lighthouse on the production site (https://www.merlinspotions.com), and save the result as a CircleCI artifact. For more details, see jira ticket.

Note: `LIGHTHOUSE_CHROMIUM_PATH` env var must be set to `/usr/bin/google-chrome-stable` in Circle for this to work. At first I had it in my Lighthouse command but then decided it was better to keep it in Circle. Also we must use a newer version of Node (i picked 6.9.1)

### TODO:
- [x] Figure out how to run Lighthouse with Mobify Preview on a locally served bundle
- [x] Fail the build if Lighthouse score is lower than 85
- [ ] Can we make this run faster? Or selectively run it (eg. only on merge to `develop`)


jansepar:
Can we make this fail the build if the score is lower than 85?

guisers:
@jansepar Yes, i believe so... we're blocked on this one now though because we aren't able to run Lighthouse while previewing a bundle, so we can't run the audit on anything that isn't production. So putting it aside until we figure out what to do about that

jansepar:
We just gotta get this in:

https://github.com/mobify/progressive-web-scaffold/pull/190

:)

guisers:
The problem though is that there isn't a good way to run Lighthouse and Preview simultaneously... so even if we can push up bundles, we still can't get Lighthouse to audit anything except for the bundle live in production, which defeats the purpose of being able to know the Lighthouse score of a change before merging it.

mikenikles:
Let's explore a way to leverage the excellent work done in https://github.com/mobify/paulaschoice-progressive/blob/develop/bin/proxy.js.

The proxy will intercept network requests made by Lighthouse and inject the correct bundle

package.json
```diff
@@ -92,6 +93,7 @@
     "lint:fix": "npm run lint:js -- --fix",
     "lint:js": "eslint '**/*.{js,jsx}'",
     "lint:sass": "sass-lint -c .sass-lint.yml 'app/**/*.scss' -v",
+    "lighthouse": "lighthouse --output=html --output-path=./lighthouse/audit.html https://www.merlinspotions.com",
```
johnboxall: I'd love for this command to live under the "test" namespace in the set of commands we expose.

Its really just a specific kind of test (performance & functionality) – the fact that it uses Lighthouse under the covers is an implementation detail.

The same goes for our Nightwatch tests. They should be exposed via `npm run test:e2e`.

mikenikles:
We could deploy Merlin's Potions to a staging environment whenever we merge into `develop`. Once deployed, we run Lighthouse against that URL.

@sap1ens What would it take to set up a staging environment?

johnboxall:
Assuming that the challenge here has to do with loading a "preview" bundle, there are many bad ways to work around this:

0. [Preview Cookie Spoofing](https://gist.github.com/johnboxall/ec9a6c5a0909c02b35d9): If you have the ability to set cookies on the browser being used, you can use a script like this to grab a preview cookie. Then you can just pass the regular old preview params to the page to get it to load.
1. Modify Preview to always accept a special header: If we could modify Preview to always work in a special condition that can only be replicated in a test environment, we could get this to work.
2. HTTP Proxy: There are many different ways we could do this that range from tag rewriting as mentioned above to other crazy ideas like DNS remapping cdn.mobify.com to the local bundle we're testing.

sap1ens:
@mikenikles 

> What would it take to set up a staging environment?

it's planned for Q1, I don't think it'll happen earlier

sap1ens:
So, I don't think that staging environment is going to help us:
- You need to be able to run this check locally for your local files. This is a normal dev workflow, it shouldn't involve any 3rd party systems. 
- People will override changes in staging all the time, it's not reliable. It's just not possible to use it concurrently. 
- We **might** run a test against staging on CircleCI when we merge to `develop`, for example. But in this case we should deploy our changes **before** all tests are finished, to be able to see them. And what do we do if check fails? Roll it back? 

I think we should find a way to run this against local files somehow. I'll try to tackle it today/tomorrow (using John's advices).

sap1ens:
So, it's working! I still want to do some cleanup, but it's working! Check the changes.

I followed this advice from John 😉 :
> crazy ideas like DNS remapping cdn.mobify.com to the local bundle we're testing.

More comments coming tomorrow.

package.json
```diff
@@ -38,6 +38,8 @@
     "chalk": "1.1.3",
     "classnames": "2.2.5",
     "copy-webpack-plugin": "4.0.1",
+    "concurrently": "^3.1.0",
```
mikenikles: Do you mind fixing the version by removing the `^`?

package.json
```diff
@@ -55,7 +57,7 @@
     "jest-cli": "18.0.0",
     "jsdom": "9.9.1",
     "json-loader": "0.5.4",
-    "lighthouse": "^1.2.1",
+    "lighthouse": "^1.3.0",
```
mikenikles: Same as above with the version.

proxy/interceptor.py
```diff
@@ -0,0 +1,12 @@
+class Replacer:
```
mikenikles: Python...? 😮 Any chance we can convert that to a node script?

package.json
```diff
@@ -95,6 +97,7 @@
     "lint:fix": "npm run lint:js -- --fix",
     "lint:js": "eslint '**/*.{js,jsx}'",
     "lint:sass": "sass-lint -c .sass-lint.yml 'app/**/*.scss' -v",
+    "postinstall": "scripts/install.sh",
```
mikenikles: This fails for me. Logs attached.
[npm-debug.txt](https://github.com/mobify/progressive-web-scaffold/files/668308/npm-debug.txt)

sap1ens:
So, I fixed versions for concurrently and lighthouse.

proxy/interceptor.py
sap1ens: I wish! :-/ 

I spent the whole day yesterday trying node stuff - node-http-proxy, proxy-tamper and a few others. There are no working node proxies that support SSL and Gzip (which is very reasonable requirement) and allow content modification.

And mitmproxy just works! It is a big dependency though and I'm not sure if we want to install it by default. We can only install it in CircleCI, for example. And you'll need to install it manually locally if you want to run lighthouse tests against local files. WDYT?

package.json
sap1ens: I can't reproduce it, could you send me your `npm i` output? Something like `npm i > output.txt`.

sap1ens:
Additional TODO (for myself):
- [ ] Document a few gotchas with proxy (setup and Mac OS X firewall warning)
- [ ] Fail CircleCI build based on lighthouse score (use JSON output?)
- [ ] Optimize CircleCI build, right now it's too slow

proxy/interceptor.py
sap1ens: Actually, I can try https://github.com/mobify/paulaschoice-progressive/blob/develop/bin/proxy.js and see if it's going to work for Merlin's

sap1ens:
@mikenikles please review again, no Python anymore :)

sap1ens:
CircleCI failed, let me fix it first

sap1ens:
Also added score check. CircleCI fails if it's less than 40, for now. It's easy to change (just package.json).

sap1ens:
@mikenikles so, everything worked until I switched to CircleCI 2.0. Now obviously Docker and Google Chrome don't work well with each other by default, we need to either find a different base image (may be https://hub.docker.com/r/justinribeiro/lighthouse/) or prepare proper headless setup.

Do you think it's worth trying? Or should we just rollback to CircleCI 1.0?

proxy/index.js
```diff
@@ -0,0 +1,80 @@
...
+    app.use((req, res, next) => {
+        res.writeHead(301,
+            {Location: 'https://localhost' + req.url}
+        )
+        res.end()
+        next()
+    })
+    http
+        .createServer(app)
+        .listen(80, '127.0.0.1')
```
mikenikles: Listening on port 80 requires `sudo` privileges. While not ideal, we can't get around it. I added it to my documentation tasks for Q1.

sap1ens: The same for 443 btw. And yeah, it's hard to avoid. 

But sudo is not a problem anymore :)

mikenikles:
@sap1ens Ideally, I'd like to stay on CircleCI 2.0 since it lowers build times significantly. Could you please see if that Lighthouse docker image works?

mikenikles:
@sap1ens Let me change my mind on that. With Progressive Mobile moving into this same repo, now is not the right time to migrate to Circle 2.0.

sap1ens:
@mikenikles CircleCI is happy! Can I merge now?

mikenikles:
For sure! Thanks for getting that to work!!

PR #190, opened 2016-11-26
Get bundle live on site

kschmidtdev:
One option to get the scaffold bundle live on https://www.merlinspotions.com - just push to the existing Mobify Cloud project. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-992

## Changes
- Set the `package.json` `name` to `platform-scaffold`
- Add a `projectSlug` value to `package.json` - use that for Cloud project name (uploading bundles) and `a.js` loading parameter

## How to test-drive this PR
- Run Preview (see `npm run dev` output)
- Observe the Web Push ask still occurs (make sure to emulate a non-iOS browser, clear cookies/site data as required to get the ask to show)


package.json
```diff
@@ -1,5 +1,6 @@
 {
-  "name": "progressive-web-scaffold",
+  "name": "platform-scaffold",
```
jeffkamo: Hmm, is this something that needs to change? Does it not make sense for this to read the same as the repo's name?

mikenikles: I agree with Jeff. Let's keep it as `progressive-web-scaffold` until we move to the new `platform-scaffold` repo.

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "progressive-web-scaffold",
+  "projectSlug": "merlinspotions-2",
```
kschmidtdev: `merlinspotions-2` is the currently-live cloud project - so we could make a new one, but we would then have to update the cloud config/tag on merlins..

mikenikles: Fair point, what about the existing `progressive-web-scaffold` slug? We would have to update the tag on Merlin's but at least it's an existing project in Cloud, and more descriptive.

kschmidtdev: Yes - we can use that and update the tag.

kschmidtdev: We'll still have to technically use teh `merlinspotions-progressive` cloud 'project' because it's hooked up to web push and we don't want to mess that up right now.

But I can update its tag to point to `progressive-web-scaffold` when it gets a cloud bundle (and I will document this setup)

mikenikles:
👍  Keeping in mind that we have to update the Mobify tag on www.merlinspotions.com as a follow-up task in order to deploy progressive web.

PR #191, opened 2016-11-26
Refactor pdp

jeffkamo:
Tweak PDP to improve its placeholder skeletons.

| Before | After |
| --- | --- |
| skeletons-before | skeletons-after |

> Notice that in the **Before** animation that the product description is not represented in the skeleton, that the carousel jumps in size after it loads, and that the form fields are missing.

## Changes
- Update SDK to 0.10.6 and add Ratio component base styles
- Add `display` utility class for `display: block`
- Tweak `disabled` styles for buttons
- Replace remnant `c-` with `pw-` in carousel theme styles
- Tweak PDP JSX files to improve its skeleton placeholders

## How to test-drive this PR
- Run `npm run dev`
- [Preview Merlin's Potions](https://preview.mobify.com/?url=http%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Visit any PDP
- Confirm that the PDP skeleton placeholders look decent


app/containers/pdp/partials/pdp-carousel.jsx
```diff
@@ -2,28 +2,54 @@ import React, {PropTypes} from 'react'
...
+                        alt: '', // no alt text available :(
+                        src: item.img,
+                        hidePlaceholder: true,
+                        ratio: {aspect: '1:1'},
+                        loadingIndicator: <SkeletonBlock height="100%" />
+                    }
+
+                    return (
+                        <CarouselItem key={idx}>
+                            <Image {...imgProps} alt={imgProps.alt} />
```
DanielBrookRoberge: This line doesn't seem like it needs `alt={imgProps.alt}`.

app/containers/pdp/partials/pdp-carousel.jsx
```diff
@@ -2,28 +2,54 @@ import React, {PropTypes} from 'react'
...
+        nextIcon: 'chevron-right',
+        iconSize: 'medium',
+        className: 'pw--frame pw--side-controls t-pdp__carousel u-padding-md u-bg-color-neutral-20'
+    }
+
+    // So long as we have items, display the carousel as intended!
+    if (items.length) {
+        return (
+            <Carousel {...carouselProps}>
+                {items.map((item, idx) => {
```
DanielBrookRoberge: If this becomes `items.map(({img}, idx) =>` then we don't need `item.`

app/containers/pdp/partials/pdp-carousel.jsx
fractaltheory: I think we'd receive an accessibility ESlint warning for not including it.

jeffkamo: Yup, it's just a work around to avoid the lint warning.

app/containers/pdp/partials/pdp-carousel.jsx
jeffkamo: Fair point. Fixed!

app/containers/pdp/partials/pdp-carousel.jsx
DanielBrookRoberge: Might be good to have a comment then (or even a lint disable), since that's not clear from the code.

jeffkamo: Fixed!

PR #192, opened 2016-11-29
Refactor Sign-in & Register Skeletons

jeffkamo:
Refactor the Sign-in and Register Skeletons

## Changes
- Improve skeletons for Sign-In and Register pages
- Improved button background transitions (i.e. from disabled to enabled)

| Before | After |
| --- | --- |
|  signin-before | signin-after |

| Before | After |
| --- | --- |
| register-before | register-after |

## How to test-drive this PR
- Run `npm run dev`
- [Preview Merlins Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2Feye-of-newt.html&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Visit the [Sign-in page](https://www.merlinspotions.com/customer/account/login/)
- Visit the [Register page](https://www.merlinspotions.com/customer/account/create/)
- Observe the page skeletons as they load and ensure all of them appear consistent in proportion to their final content.


app/containers/login/login.scss
```diff
@@ -47,3 +50,12 @@
     }
 }
 
+
+// Registration
+// ---
+
+.t-login__register-fieldset {
+    & + & {
```
sap1ens: Can be a bit hard to read. I'd replace it with something like `nth-child(2)`

app/containers/login/reducer.js
```diff
@@ -9,6 +9,85 @@ import {onPageReceived} from '../app/actions'
 import signinParser from './parsers/signin'
 import registerParser from './parsers/register'
 
+const signinFields = [
```
sap1ens: This predefined set of fields improves general layout, but it can be too much... I'm not sure if it's a best practice, just want to comment on it ¯\\_(ツ)_/¯

jeffkamo: Ya, I'm not sure we have a defined best practice for this sort of thing. I'm open to taking a better approach if there is one.

From my standpoint, I want to prevent reflows as much as possible, and offer the ability for users to interact with the forms sooner (for perceived performance) rather than waiting for the content to appear. But I understand that could conflict with our desire to use the desktop site as a source for that content.

Thoughts?

jansepar: In my mind, we are the ones perscribing best practices here, we should be opinionated. If I was to write down what our line of thinking should be when implementing these forms, it would be:

1. Use our defaults

sap1ens: @jansepar I just wanted to mention that those fields are used only until we get a "real" set of fields from the page/parser, so they're just used as a placeholder.

PR #195, opened 2016-12-06
Production bundle fixes

kschmidtdev:
If we build for production, some of the optimization/obfuscation steps break how we do routing. 

 **Linked PRs**: #190 

## Changes
- Use the component type instead of component name for routing

## How to test-drive this PR
- Observe the site works as it did previously
- Push a bundle
- Preview the production bundle - observe the site works as it did previously

### DIFF
--- a/app/containers/home/parsers/home.test.js
+++ b/app/containers/home/parsers/home.test.js
@@ -9,27 +9,27 @@ describe('the home parser', () => {
         const expected = {
             categories: [
                 {
-                    href: 'http://www.merlinspotions.com/potions.html',
+                    href: 'https://www.merlinspotions.com/potions.html',
                     text: 'Potions'
                 },
                 {
-                    href: 'http://www.merlinspotions.com/books.html',
+                    href: 'https://www.merlinspotions.com/books.html',
                     text: 'Books'
                 },
                 {
-                    href: 'http://www.merlinspotions.com/ingredients.html',
+                    href: 'https://www.merlinspotions.com/ingredients.html',
                     text: 'Ingredients'
                 },
                 {
-                    href: 'http://www.merlinspotions.com/supplies.html',
+                    href: 'https://www.merlinspotions.com/supplies.html',
                     text: 'Supplies'
                 },
                 {
-                    href: 'http://www.merlinspotions.com/charms.html',
+                    href: 'https://www.merlinspotions.com/charms.html',
                     text: 'Charms'
                 },
                 {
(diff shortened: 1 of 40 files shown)


app/containers/home/parsers/home.test.js
```diff
@@ -9,40 +9,40 @@ describe('the home parser', () => {
         const expected = {
             categories: [
                 {
-                    href: 'http://www.merlinspotions.com/potions.html',
+                    href: 'https://www.merlinspotions.com/potions.html',
```
mikenikles: HTTPS everywhere - thanks!

mikenikles:
👍  from me, but I'd like @DanielBrookRoberge to add his feedback too.

app/containers/pdp/reducer.js
```diff
@@ -29,9 +29,9 @@ export const initialState = Immutable.fromJS({
 
 const reducer = createReducer({
     [onPageReceived]: (state, action) => {
-        const {$, $response, pageType, url, currentURL} = action
+        const {$, $response, pageComponent, url, currentURL} = action
 
-        if (RouterUtils.isPageType(pageType, PDP)) {
+        if (RouterUtils.isPageType(pageComponent, PDP)) {
```
DanielBrookRoberge: If I've read the definition of `isPageType` correctly, this will work if `PDP` is the wrapped component from the default export of `pdp/container.js`. Do you know if that works?

app/containers/home/reducer.js
```diff
@@ -13,9 +14,9 @@ const initialState = fromJS({
 
 export default createReducer({
     [onPageReceived]: (state, action) => {
-        const {$, $response, pageType} = action
+        const {$, $response, pageComponent} = action
 
-        if (pageType === 'Home') {
+        if (pageComponent === Home) {
```
DanielBrookRoberge: Could this be updated to use the `RouterUtils.isPageType` function?

app/containers/pdp/reducer.js
kschmidtdev: I'm not sure I fully understand - I had to change the import of the `PDP` type from the default export to destructuring (and the container exporting the type explicitly)

I thought the wrapped check was for when the router returns a wrapped container.

kschmidtdev: It works! I've updated the PR and now the explicit exports aren't required, all the */container.js files are now unchanged 🎉

PR #197, opened 2016-12-08
Add a "catalog" key to redux store and put product data in there

guisers:
Currently, our Redux store is organized by page component. Eg. the PLP and PDP both have a top-level slice of state. To "share information", the PDP reducer listens for `onPageReceived` of PLP pages and parses out the information it needs for its own state. This results in doubled information in the Redux store, which isn't ideal.

In Pure Formulas, the pattern is slightly different. PLP and PDP are still top-level slices of state, but the PDP container maps the parts of the PLP's state that it needs to its own props to use them. This eliminates doubled information but we lose some separation of concerns.

This PR aims to refactor the way our Redux data store is organized. At the top level, we want to have a new `Catalog` slice of state that holds shared data. Product data will go into here, and the PLP and PDP will both look to the Catalog to find the information they need, while the Catalog reducer handles parsing this data as it arrives. PLP and PDP reducers are left to only figure out UI-related information.

**JIRA**: https://mobify.atlassian.net/browse/WEB-1018

## Changes
- Add `Catalog > Products` state
- Refactor reducers, containers, parsers to reflect this

## How to test-drive this PR
- Visit a PLP (https://www.merlinspotions.com/potions.html)
- Click on one of the products (https://www.merlinspotions.com/eye-of-newt.html)
- Observe how the title, price, and image load right away because it comes from PLP


guisers:
Hey all, I've just tagged whoever was involved/interested in the [previous iteration of this](https://github.com/mobify/progressive-web-scaffold/pull/174)!

fractaltheory:
Resurrecting this one; it looks good to me 👍 . Much of the code is the same (just moved from separate plp/pdp parsers/reducers to the catalog). You could argue about the state key names of `catalog.products` (i.e. what's wrong with just a `products` or `catalog` top-level state, or putting it in `app` instead) but that doesn't seem like a blocker.

DanielBrookRoberge:
This works for me and looks good as a starting point for the refactors we'll need for the middleware. Merging!

PR #198, opened 2016-12-09
Support for using a custom a.js slug

kschmidtdev:
Since we need web push to still work for demos on Merlins Potions, and we are using a different cloud project than what is live (and is set up for web push) - we need to load the a.js file from a different slug. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-992
 **Linked PRs**: #190, #195

## Changes
- Support for separate a.js slug in `package.json` and used in `main.jsx`

## How to test-drive this PR
- Preview
- Observe a.js loads from the right place (check the Network Debug Panel)

### DIFF
--- a/app/main.jsx
+++ b/app/main.jsx
@@ -20,7 +20,7 @@ import {initMobifyAnalytics} from 'progressive-web-sdk/dist/analytics'
 
 polyfill()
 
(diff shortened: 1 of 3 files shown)


sap1ens:
LGTM

PR #199, opened 2016-12-09
Remove the `build` directory on `prod:build`

kschmidtdev:
The `build` directly doesn't get cleared out when you do a production build - this has been a sneaky problem for me in the past. 

## Changes
- When running `prod:build`, `build/` is removed

## How to test-drive this PR
- Without this PR, run `prod:build`, put some random file(s) in `build/`, run `prod:build` again, observe those files are still around
- With this PR, perform the above again, observe the random files are gone in the end


package.json
```diff
@@ -87,7 +87,7 @@
     "add:page": "sdk-generate page",
     "build-sprites": "svg-sprite --symbol --symbol-dest='.' --symbol-sprite='sprite.svg' --shape-id-generator='shoppicon-%s' --dest=./app/static/sprite ./app/static/svg/*.svg",
     "dev": "sdk-get-routes && sdk-create-hash-manifest && node ./dev-server/index.js",
-    "prod:build": "sdk-get-routes && sdk-create-hash-manifest && webpack --config webpack/production.js -p --display-error-details --optimize-dedupe",
+    "prod:build": "rm -rf build/ && sdk-get-routes && sdk-create-hash-manifest && webpack --config webpack/production.js -p --display-error-details --optimize-dedupe",
```
mikenikles: A more generic approach that also works on Windows is to use https://www.npmjs.com/package/rimraf. It's a dev dependency we could add to make sure our customers on Windows won't run into any trouble.

mikenikles:
Thanks for updating that. 👍

PR #200, opened 2016-12-09
Fix the 'New Arrivals' asset when previewing pushed bundle

kschmidtdev:
If you preview a bundle from cloud, the 'new arrivals' asset fails to load. This is because of a space in the 'section title' that we end up using as a filename. We’ve been using the image name with a %20 in it, which works with web pack - however once deployed to a CDN, the image cannot be accessed that way (the %20 is url decoded into a space). So let’s just convert spaces to dashes and rename the asset. 

## Changes
- Rename assets with spaces to have dashes instead
- Convert untrimmed whitespace to dashes during asset url transformation

## How to test-drive this PR
- Preview this PR locally
- Observe the 'new arrivals' asset loads correctly
- Preview https://go.mobify.com/progressive-web-scaffold/b/32 (a bundle created prior to this PR)
- Observe the 'new arrivals' asset does not load
- Preview https://go.mobify.com/progressive-web-scaffold/b/33 (a bundle created from this PR)
- Observe the 'new arrivals' asset loads correctly


mikenikles:
👍

PR #205, opened 2016-12-13
Custom header/footer support (checkout feature)

kschmidtdev:
Most of the checkout flow has a different header and footer (according to the mocks) The scaffold currently does not support overriding/customizing this, so this PR addresses this. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1015

## Changes
- Allow routes to optionally specify a header and/or footer to use
- Implement a basic header/footer for the checkout

## How to test-drive this PR
- Preview and observe the header/footer throughout the site is unchanged
- Go to https://www.merlinspotions.com/checkout/shipping/ and observe the header and footer look different


app/containers/checkout-footer/container.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import logo from '../../static/svg/logo.svg'
+import DangerousHTML from 'progressive-web-sdk/dist/components/dangerous-html'
+
+const CheckoutFooter = function(props) {
```
jeffkamo: It probably doesn't _have_ to be driven by props, but perhaps the option might be nice. In our case for Merlin's, it's probably not necessary.

app/containers/checkout-header/_checkout-header.scss
```diff
@@ -0,0 +1,38 @@
...
+}
+
+
+// Title
+// ---
+
+.t-checkout-header__title {
+    padding-top: 11px;
+    padding-bottom: 13px;
+    padding-left: 12px;
```
jeffkamo: For padding, as much as possible these should be based on the `$unit` value. Magic numbers should be avoided as much as possible. In this case the padding doesn't need to be exact, and you can rely on flexbox `alignment-items` to get the logo text centred instead.

jeffkamo: Actually this extra padding shouldn't be necessary: I assume you added it to help get the header container to be the right height, but the remaining height comes from the lock icon and signin button that is currently missing.

screen shot 2016-12-14 at 11 32 00 am

You can see in the normal header bar heights (i.e. in the main nav and mini-cart) are dictated by their close buttons, as per the following:

screen shot 2016-12-14 at 11 33 34 am

app/containers/checkout-footer/container.js
```diff
@@ -0,0 +1,24 @@
...
+        <footer className="t-checkout-footer">
+            <div className="t-checkout-footer__inner u-padding-lg u-text-align-center">
+                <div className="u-flex">
+                    <div className="t-checkout-footer__logo-wrapper">
+                        <DangerousHTML html={logo}>
+                            {(htmlObj) => <div className="t-checkout-footer__logo" dangerouslySetInnerHTML={htmlObj} />}
+                        </DangerousHTML>
+                    </div>
+                </div>
+                <div className="t-checkout-footer__copyright u-padding-top u-padding-bottom">
```
jeffkamo: Since you already have padding on all sides of the footer inner wrapper, this extra bottom padding here is probably not necessary.

app/containers/checkout-footer/_checkout-footer.scss
```diff
@@ -0,0 +1,44 @@
...
+    opacity: 0.5;
+    color: #fff;
+    font-size: 12px;
+    line-height: $font-size;
+}
+
+.t-checkout-footer__logo {
+    width: 67px;
+    height: 28px;
+    opacity: 0.5;
```
jeffkamo: Minor: You could reduce the two opacity declarations into one having only one on the `__inner` container, and move the `background-color` to the root `t-checkout-footer` container.

app/containers/checkout-footer/_checkout-footer.scss
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Copyright
+// ---
+
+.t-checkout-footer__copyright {
+    margin: 0;
+
+    opacity: 0.5;
+    color: #fff;
```
jeffkamo: Minor: this could use the `$neutral-10` variable.

app/containers/checkout-header/_checkout-header.scss
```diff
@@ -0,0 +1,38 @@
...
+.t-checkout-header__bar {
+    position: fixed;
+    top: 0;
+    left: 0;
+    z-index: $z2-depth; // 2
+
+    overflow: hidden;
+    width: 100vw;
+
+    padding: 0;
```
jeffkamo: I feel like this padding reset should be avoided if possible. See the padding comments for `__title`.

app/containers/checkout-header/_checkout-header.scss
```diff
@@ -0,0 +1,44 @@
...
+    align-items: center;
+    font-size: $unit*3;
+    line-height: $unit*3.5;
+}
+
+
+// Lock icon
+// ---
+
+.t-checkout-header__icon-lock {
```
jeffkamo: Ya that's fine by me.

jeffkamo: actually the way you have it now is fine! 👍

jeffkamo:
👍

PR #209, opened 2016-12-16
Notifications

kschmidtdev:
Port of the notification system from [Crabtree](https://github.com/mobify/crabtree-evelyn-progressive) ([original PR here](https://github.com/mobify/crabtree-evelyn-progressive/pull/52))

Renamed to Notifications. Required for Progressive Checkout. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-808

## Changes
- Add Notification system

## How to test-drive this PR
- In `app/containers/pdp/actions.js` add the following import:
`import {addNotification} from '../app/actions'`
then within `submitCartForm` add the following (line 23):
`        dispatch(addNotification({
            content: 'Hello message',
            id: 'charitableSubmissionMessage',
            showRemoveButton: true
        }))
`
- Preview and add an item on a PDP
- Observe the notification pops into view
- Observe that tapping the 'x' button in the notification dismisses it.


jeffkamo:
Aaah ya, this is so awesome! Looking good, and works as advertised.

The only thing that feels uncomfortable to me is having to pull in another container's actions. Is a "correct" way of doing things? Should actions only be "visible" to their parent container?

app/containers/header/container.js
```diff
@@ -66,63 +70,79 @@ class Header extends React.Component {
...
+
+                        <HeaderBarActions>
+                            <Button className="u-position-relative" innerClassName={innerButtonClassName} onClick={onMiniCartClick}>
+                                <IconLabel label="Cart" iconName="cart" iconSize="medium" />
+                                {cartCounterBadge}
+                            </Button>
+                        </HeaderBarActions>
+                    </HeaderBar>
+                    {notifications &&
+                        <NotificationManager notifications={notifications} actions={notificationActions} />
```
jeffkamo: Since the actions belong to App, would it make more sense to add these components to the App container instead of the site Header? That way this wouldn't need to be redefined for the checkout header.

app/containers/header/partials/notification-manager.jsx
```diff
@@ -0,0 +1,22 @@
...
+        </div>
+    )
+}
+
+NotificationManager.propTypes = {
+    actions: PropTypes.object,
+    notifications: PropTypes.array
+}
+
+export default NotificationManager
```
jeffkamo: Could this be a top level component instead of a template partial? Maybe the `Notification` component should be a sub-component of a `NotificationManager` component?

app/components/notification/_base.scss
```diff
@@ -0,0 +1,86 @@
...
+// Notification
+// ===
+
+// Notification root
+// ---
+//
+// 1. Slight increase of height transition duration to give the staggered collapse animation
+
+.c-notification {
+    display: flex;
```
jeffkamo: This could use the `u-flexbox` utility class instead of adding it via CSS

kschmidtdev:
@jeffkamo Can you point out where you are seeing this? To post a notification you have to import the 'app actions', something we do in almost every reducer, so I think it's common?

app/components/notification/_base.scss
```diff
@@ -0,0 +1,86 @@
...
+    overflow: hidden;
+    margin-bottom: $unit/2;
```
jeffkamo: There is also the `u-margin-bottom-sm` utility class

app/components/notification/_base.scss
```diff
@@ -0,0 +1,86 @@
...
+
+    background: rgba($brand-color, 0.85);
+    color: $neutral-10;
```
jeffkamo: And a `u-color-neutral-10`

app/components/notification/_base.scss
```diff
@@ -0,0 +1,86 @@
...
+
+// State: Dismissing
+// ---
+//
+// 1. Override inline styles to collapse notification
+// 2. Prevent contents from leaking out
+
+.c-notification.c--dismissing {
+    overflow: hidden; // 2
+    height: 0 !important; // 1
```
jeffkamo: could this be overridden in the JSX while the dismissing state is being applied? Hopefully to avoid the `!important`

jeffkamo:
@kschmidtdev re app actions: oh, I see. There's already what looks like an unavoidable precedent for importing app actions elsewhere... well shucks

app/containers/header/container.js
kschmidtdev: That's a good idea - I'll look into that!

app/containers/header/partials/notification-manager.jsx
kschmidtdev: I almost wonder if they should be two separate components 🤔  ?

app/components/notification/_base.scss
kschmidtdev: I will look into this!

app/styles/animations/_visibility.scss
```diff
@@ -0,0 +1,43 @@
+// Visibility Animations
```
jeffkamo: @kschmidtdev I went ahead and moved the animation keyframes you created into their own file. Sorry for butting in 🙇

app/components/notification/_base.scss
kschmidtdev: 🤔  not sure exactly how to do this but I'll grab you if I get stuck.

app/styles/animations/_visibility.scss
kschmidtdev: I simply ported (mostly copy&paste) from @cdegit 's fine work - and no worries!

app/containers/header/_header.scss
```diff
@@ -112,3 +111,24 @@
...
+.t-header__notification-manager {
+    position: absolute; // 1
+    top: 100%;
+    right: 0;
+    left: 0;
+
+    pointer-events: none; // 2
+
+    .c-notification {
+        pointer-events: auto; // 3
```
jeffkamo: This could probably be moved to the notification component's styles

app/components/notification-manager/README.md
```diff
@@ -0,0 +1,7 @@
+## Notification Usage
+
+    <Notification showRemoveButton content="This is the name that displays when you submit a product review" />
```
jeffkamo: Can this also demonstrate how to use the actual `NotificationManager` component as well?

For example...

```jsx
<NotificationManager notifications={notifications} actions={notificationActions} />
```

Or perhaps demonstrate when and where to use one over the other? Can you even use `<Notification />` on it's own?

app/components/notification-manager/_base.scss
```diff
@@ -47,3 +47,24 @@ $notification__easing: cubic-bezier(0.45, 0.1, 0.5, 1);
...
     animation: visibility-hide $notification__duration 1 forwards;
 }
+
+// Notification Manager
+// ---
+//
+// 1. Position relative to `.t-header__bar`
+// 2. Prevent absolute positioned `<div>` from blocking clicks on elements
+//    below it
+// 3. Allow interactions with the notifications inside notification manager container
```
kschmidtdev: So that's what the numbering was for...

app/components/notification/_base.scss
```diff
@@ -0,0 +1,47 @@
...
+
+
+// State: Dismissing
+// ---
+//
+// 1. Prevent contents from leaking out
+
+.c-notification.c--dismissing {
+    overflow: hidden; // 1
+    height: 0 !important;
```
jeffkamo: Since this is so important after all, can we add a line comment to document the reason so we don't forget? This commit message would probably do the trick:

> required for `onTransitionEnd` to actually fire for that component

PR #210, opened 2016-12-18
Enable code splitting

mikenikles:
Split the `react`, `react-dom` and `react-router` libraries into a separate `vendor` bundle. Browsers can cache this independently from our application code, which changes more frequently.

Bundle sizes in development are:
- `main.js` => 1.5 MB
- `vendor.js` => 702 KB
- `loader.js` => 14.4 KB
- `worker.js` => 46.3 KB

Bundle sizes in production are:
- `main.js` => 549 KB
- `vendor.js` => 144 KB
- `loader.js` => 7.88 KB
- `worker.js` => 15.1 KB

 **Linked PRs**: #211 

## How to test-drive this PR
- `npm run dev`
  - Make sure www.merlinspotions.com previews as expected
- `npm run prod:build`
  - Validate that the build process completes without errors
  - Deploy the bundle to production and make sure www.merlinspotions.com loads as expected.

PR #211, opened 2016-12-18
Use preact-compat at runtime to reduce the bundle size

mikenikles:
As an alternative to splitting React into a `vendor` bundle as seen in #210, we could instead use `preact` at build time.

Bundle sizes in development are:
- `main.js` => 1.45 MB
- `loader.js` => 14.1 KB
- `worker.js` => 46.3 KB

 **Linked PRs**: #210 

## Changes
- Alias `react` and `react-dom` to `preact-compat`

## How to test-drive this PR
- `npm run dev`
  - Check that webpack builds properly


mikenikles:
After discussions with the team, we're closing this in favour of #210.

PR #216, opened 2016-12-21
QA Test Parallelism Setup

jasecode:
QA Test Parallelism Setup

## Changes
- added test_workers in nightwatch.json , whenever you run locally or on circle with only a single container, nightwatch will start chiild processes and distribute the tests evenly among them to decrease test run time
- added to smoke-test.sh parallelism logic: if on circle and there are multiple containers, it will find all test files under /workflows to distribute evenly among the circle machines for parallelism to decrease test run time. 

- [x] QA Engineer +1


package.json
```diff
@@ -102,6 +102,7 @@
     "smoke-test": "./tests/system/smoke-test.sh",
     "test": "cross-env NODE_ENV=test jest",
     "test:all": "npm run lint && npm test",
+    "test:e2e": "node ./node_modules/nightwatch/bin/runner.js -c ./tests/system/nightwatch-config.js --suiteRetries 1",
```
ellenmobify: This is the same as the `nightwatch` task above. This is open to further discussion but perhaps we can remove the `nightwatch` task in favour of `test:e2e` since the latter hides the implementation detail (ie. we're running *end-to-end* tests, not just Nightwatch tests)

ellenmobify: Scaffold changes won't affect projects that have already been generated. We can remove the `nightwatch` task here; other projects can update their README if/when they switch to the new name.

ellenmobify:
👍

package.json
johnboxall: @jasecode @ellenmobify we discussed this briefly in person, but I'd love to see us collapse the number of test commands if at all possible here. We should have commands to ...

* Lint
* Run Unit Tests
* Run E2E Tests
* Run Performance Tests (PWA Tests)
* Run All! the Tests! (potentially on CI)

All test commands should be namespaced under the same prefix. It should be clear from the name of the test command what they test. We should optimize should give the most commonly used test command the shortest alias (`npm test`)...

PR #220, opened 2016-12-28
Revert back to CircleCI Classic.

mikenikles:
Two main reasons for this change:
- Progressive Mobile is becoming part of this repo.
- We run into some issues with the new Lighthouse tests on CircleCI 2.0

Based on the above, now is not a good time to migrate to a new CI platform that hasn't been publicly released.

## How to test-drive this PR
- Make sure CircleCI tests pass.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
 test:
     override:
         - npm run lint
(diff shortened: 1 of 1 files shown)


mikenikles:
Yes 😏

PR #225, opened 2017-01-04
Astro: Set up header

jeremywiebe:
Sets up the native header bar in Merlins Astro. 

 **JIRA**: [**HYB-918**](https://mobify.atlassian.net/browse/HYB-918)
 **Linked PRs**: https://github.com/mobify/astro/pull/744

## TODO
- [ ] ~~Title in navigation bar~~  👈  will do in future PR
- [ ] ~~Test/ensure working on Android~~  👈  will do in future PR

## Changes
- Adds native header bar
- Suppresses web header and footer when running in app
- Adds tests for header and footer suppression
- Converts `onChange` in app-provider.jsx to be asynchronous. 
   **Note**: *The logic in this method has gotten significantly more complicated and it would be great for someone else to review/validate.*

## How to test-drive this PR
- Run the app (note that no web header or footer is displayed on any page)
- Preview the web app and note that the header and footer are rendered


native/app/config/baseConfig.js
```diff
@@ -7,11 +7,19 @@ const localPreviewUrl = Astro.isRunningInIOSApp()
     ? '//localhost:8443/loader.js'
     : '<local_ip>:8443/loader.js'
 
+const colors = {
+    primaryColor: '#4E439B',
+    secondaryColor: '#007ba7',
```
kerrmarin: Random thought, but could these colours come from the `web` folder somehow? That way if the colours change in the web style we wouldn't need to update this

kerrmarin: Might not be worth the effort

web/app/app-provider.jsx
```diff
@@ -51,22 +61,39 @@ const AppProvider = ({store}) => {
         }
     }
 
-    const onChange = (prevState, nextState) => {
+    const onChange = (prevState, nextState, replace, callback) => {
         const prevURL = getURL(prevState)
         const nextURL = getURL(nextState)
+        const coordinateWithNativeApp = Astro.isRunningInApp() &&
```
kerrmarin: (Trivial) Should this be `shouldCoordinateWithNativeApp` or `isCoordinatingWithNativeApp`?

web/app/app-provider.jsx
```diff
@@ -51,22 +61,39 @@ const AppProvider = ({store}) => {
         }
     }
 
-    const onChange = (prevState, nextState) => {
+    const onChange = (prevState, nextState, replace, callback) => {
```
kerrmarin: `replace` doesn't seem to be used -- is this coming from React?

web/app/containers/app/container.js
```diff
@@ -10,7 +10,8 @@ import Navigation from '../../containers/navigation/container'
 import * as navActions from '../../containers/navigation/actions'
 import * as miniCartActions from '../../containers/mini-cart/actions'
 import sprite from '../../static/svg/sprite-dist/sprite.svg'
-import * as appActions from '../app/actions'
+import * as appActions from './actions'
```
kerrmarin: hah

web/app/containers/app/container.js
```diff
@@ -57,7 +67,7 @@ class App extends React.Component {
 
                 <div id="app-wrap" className={currentTemplate}>
                     <div id="app-header" role="banner">
-                        <CurrentHeader onMenuClick={openNavigation} onMiniCartClick={requestOpenMiniCart} />
+                        <CurrentHeader onMenuClick={openNavigation} onMiniCartClick={requestOpenMiniCart} isRunningInAstro={isRunningInAstro} />
```
kerrmarin: 🤔  should we move away from using "astro" in our codebase and use "app"? Is this ever going to be accessed by a partner-developer? `app` might make more sense to them than `astro`? Probably not as important if this is not going to be used outside of Mobify

web/app/containers/footer/container.js
```diff
@@ -32,11 +32,14 @@ class Footer extends React.Component {
     }
 
     render() {
+        if (this.props.isRunningInAstro) {
+            return false
```
kerrmarin: Is `false` the return value expected here? I was under the impression it was `null` if we didn't want anything rendered? https://facebook.github.io/react/blog/2014/07/17/react-v0.11.html#rendering-to-null

web/app/containers/footer/container.js
```diff
@@ -53,9 +56,13 @@ Footer.propTypes = {
...
     footer: PropTypes.object,
     /**
+     * Defines whether we're being hosted in an Astro app
+     */
+    isRunningInAstro: PropTypes.bool,
+    /**
      * Submit the newsletter subscription form to the backend
      */
-    submitNewsletter: PropTypes.func
+    submitNewsletter: PropTypes.func,
```
kerrmarin: Extra comma?

web/app/containers/header/container.js
```diff
@@ -52,6 +52,10 @@ class Header extends React.Component {
     }
 
     render() {
+        if (this.props.isRunningInAstro) {
+            return false
```
kerrmarin: Same as the `footer`

web/app/containers/header/container.js
```diff
@@ -112,22 +116,28 @@ Header.propTypes = {
...
 
     onMenuClick: PropTypes.func,
     onMiniCartClick: PropTypes.func,
 }
 
-const mapStateToProps = (state, props) => {
+const mapStateToProps = (state) => {
     return {
-        header: state.header
+        header: state.header,
```
kerrmarin: Are the extra commas at the end of an object a react thing? Reminds me of python 🐍

web/app/app-provider.jsx
jeremywiebe: Unfortunately, I know of no way to remove `replace` as the `callback` parameter has to be in the 4th place for it to work.

web/app/containers/app/container.js
jeremywiebe: I'm not sure about this. We've talked about naming before and I'm not sure **app** is the right choice given that we have "web apps" and "native apps".  I think we're ok still using the 'astro' codename in code.

web/app/containers/footer/container.js
jeremywiebe: I think the rest of the code uses trailing commas on hashes like this (guessing so you can add/remove any item without worrying to add/remove a trailing comma).

web/app/containers/header/container.js
jeremywiebe: See previous comment to trailing commas

web/app/app-provider.jsx
kschmidtdev: Unless the linter is complaining about unused `replace` I think it's ok to leave, or to rename to `_` which has been done elsewhere in the codebase.

jeremywiebe: Linter isn't complaining.  I think I'd prefer to leave `replace` as is so you don't have to head back to the docs to understand what that parameter does. 
[Docs](https://github.com/ReactTraining/react-router/blob/master/docs/API.md#onchangeprevstate-nextstate-replace-callback)
> If callback is listed as a 4th argument, this hook will run asynchronously, and the transition will block until callback is called.

native/app/config/baseConfig.js
jeremywiebe: I've been thinking about this more. I think it's ok to duplicate this. I'm not certain that the colours used on the web would always be the same as in app.  ?  Also, it doesn't look like a straightforward task to share those SASS variables with .js

kerrmarin: Sounds good :)

web/app/app-provider.jsx
kerrmarin: Seems fine to me -- re that quoted block, I think we need to make sure that we are calling `callback()` under every possible combination of input. At the moment, we are not calling `callback()` when you are coordinating with the app but the URL is the same as it was before.

jeremywiebe: @kerrmarin Nice catch!  Fixed that up.

native/app/controllers/tabController.js
```diff
@@ -30,10 +32,20 @@ TabController.init = async function(tabItem) {
     await layout.addTopView(headerBar)
     await layout.setContentView(navigationView)
     await navigationView.setHeaderBar(headerBar)
+
+    await headerBar.setCenterIcon(baseConfig.logoUrl, 'logo')
+    await headerBar.setTextColor(baseConfig.colors.whiteColor)
+    await headerBar.setBackgroundColor(baseConfig.colors.primaryColor)
+    await headerBar.setOpaque()
```
DanielBrookRoberge: Do these calls need to be serial? This could be parallelized using `await Promise.all([call1(), call2(),...])`

web/app/app-provider.jsx
```diff
@@ -51,22 +61,44 @@ const AppProvider = ({store}) => {
...
+                // We need to call this to allow react to continue
+                callback()
+            }
+        } else {
+            callback()
+
+            if (urlHasChanged) {
+                triggerChange()
+            }
+        }
```
DanielBrookRoberge: The logic here is simpler if the `isCoordinatingWithNativeApp` if and the `urlHasChanged` if blocks are interchanged, since if `!urlHasChanged` the necessary code is always just `callback()`

web/app/containers/app/reducer.js
```diff
@@ -2,11 +2,14 @@ import {createReducer} from 'redux-act'
...
 import * as appActions from './actions'
-import {FETCH_IN_PROGRESS, CURRENT_URL} from './constants'
+import {FETCH_IN_PROGRESS, CURRENT_URL, IS_IN_ASTRO_APP} from './constants'
+
+import Astro from '../../vendor/astro-client'
 
 const initialState = Map({
     [FETCH_IN_PROGRESS]: false,
     [CURRENT_URL]: false,
+    [IS_IN_ASTRO_APP]: Astro.isRunningInApp(),
```
DanielBrookRoberge: I know you've done a fair amount of work putting this in the Redux store, but looking at how it's used it almost doesn't make sense after all. It might be worth writing an `astro-integration.js` or something that just exports this value (and any other 'do x if in astro else do y' stuff) and then is imported where necessary.

jeremywiebe: Fixing in https://github.com/mobify/progressive-web-scaffold/pull/238

PR #226, opened 2017-01-06
Add to Homescreen Support

DanielBrookRoberge:
This PR adds add to homescreen support for Merlins on Chrome for Android. It also themes the address bar with the Merlin's brand colour.

Circle Lighthouse integration gives this branch 77/100, even though it doesn't seem to load the service worker correctly.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-935

## Changes
- Add a manifest file for Merlins
- Modify the loader to add the manifest to the page
- Remove manifest serving from the service worker (unnecessary since it can be hosted in the bundle)
- Add theme colour tag on load

## How to test-drive this PR
- Preview Merlins with this branch.
- In the Application tab of the dev tools, look for the Manifest panel
- Press the "add to homescreen" link in that panel
- See that the manifest information is populated. Without modifying the manifest to point to the local icon, it will fail until the icon is shipped in a production bundle


app/static/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "Merlin's Potions",
+  "short_name": "Merlin's",
+  "start_url": "https://www.merlinspotions.com/",
+  "background_color": "#fff",
+  "theme_color":" #4e439b",
```
fractaltheory: If we are already adding this as a `meta` tag on all pages via `loader.js` [above](https://github.com/mobify/progressive-web-scaffold/pull/226/files#diff-85044fad63ed16d573b29b4da020c90bR47), is this line necessary? From what I'm reading about it, the meta tag takes precedence over the web application manifest.

DanielBrookRoberge: Lighthouse looks for it, and I think it might be used in the from-homescreen startup process.

fractaltheory:
I was worried about hosting the manifest on the CDN (since we ran into this problem way back on PureFormulas), but it looks like cross-origin manifests are valid now: https://github.com/w3c/manifest/pull/377

fractaltheory:
LGTM 👍

app/static/manifest.json
mikenikles: https://developer.mozilla.org/en-US/docs/Web/Manifest#theme_color (and the rest of the website) has proven useful a few times when I worked on PWAs.

PR #230, opened 2017-01-10
Astro: Merlins onboarding

kerrmarin:
Initial onboarding flow

 **JIRA**: https://mobify.atlassian.net/browse/HYB-979

## Changes
- Add react app that renders the onboarding

## How to test-drive this PR
- Run the app and use the onboarding


jeremywiebe:
The button sizes being different doesn't look good, IMHO. Maybe check with @andrewr0se ?

</img>

native/onboarding/config/webpack.config.app.js
```diff
@@ -0,0 +1,234 @@
+var autoprefixer = require('autoprefixer');
```
jeremywiebe: I had to manually run `npm run build` before I could test this onboarding. Do you think we could hook that into `build-js.sh` so that it's all automatic when you hit `⌘R` in Xcode?

native/app/config/onboardingConfig.js
```diff
@@ -0,0 +1,6 @@
+
+const onboardingConfig = {
+    url: 'file:///build/index.html'
```
jeremywiebe: We should probably name the base folder for onboarding `onboarding` in the app bundle. If we add other small embedded sites/www content, it'll get confusing what's what in the bundle.

native/app/controllers/onboarding/onboardingController.js
```diff
@@ -0,0 +1,48 @@
+
```
jeremywiebe: Really like having this in it's own folder.  One pattern I've seen that I really like is not organizing by object type (`controllers`, `views`, etc) but rather by feature.  What do you think about having all onboarding stuff in `native/app/onboarding`?

native/app/controllers/onboarding/onboardingController.js
```diff
@@ -0,0 +1,48 @@
...
+    this.navigationView = navigationView
+}
+
+OnboardingController.init = async function() {
+    const [
+        webView,
+        layout
+    ] = await Promise.all([
+        WebViewPlugin.init(),
+        AnchoredLayoutPlugin.init()
```
jeremywiebe: What's the reasoning for using an AnchoredLayout here?  It looks like we don't ever use the `addTopView()` or `addBottomView()` anywhere....

native/app/controllers/onboarding/onboardingModalController.js
```diff
@@ -0,0 +1,86 @@
...
+    })
+
+    return onboardingModalController
+}
+
+OnboardingModalController.prototype.show = async function(params) {
+    params = Astro.Utils.extend({forced: false}, params)
+    const shouldShowOnboardingModalKey = 'shouldShowOnboardingModal'
+
+    const isFirstRun = await SettingsStore.get(shouldShowOnboardingModalKey) === null
```
jeremywiebe: Will this not work if somewhere we set `shouldShowOnboardingModal` to `true`?

native/app/controllers/onboarding/onboardingModalController.js
```diff
@@ -0,0 +1,86 @@
...
+        this.isShowing = true
+        this.modalView.show({animated: true})
+
+        // Promise will be resolved when onboarding modal is dismissed
+        AppEvents.on(OnboardingModalEvents.onboardingHidden, () => {
+            SettingsStore.set(shouldShowOnboardingModalKey, 'false')
+        })
+        AppEvents.trigger(OnboardingModalEvents.onboardingShown)
+    } else {
+        throw new Error()
```
jeremywiebe: Why do we throw an error here?  Can we add a message to the error if it's needed, or remove the `throw` if not?

native/app/controllers/onboarding/onboardingModalController.js
```diff
@@ -0,0 +1,86 @@
...
+    const shouldShowOnboardingModalKey = 'shouldShowOnboardingModal'
+
+    const isFirstRun = await SettingsStore.get(shouldShowOnboardingModalKey) === null
+
+    // Onboarding modal should be triggered when the app first runs
+    if (isFirstRun || params.forced) {
+        this.isShowing = true
+        this.modalView.show({animated: true})
+
+        // Promise will be resolved when onboarding modal is dismissed
```
jeremywiebe: This comment doesn't make sense to me. I don't see a promise here anywhere.

native/app/global/app-event-names.js
```diff
@@ -0,0 +1,7 @@
+const RpcEventNames = {
+    onboardingShow: 'onboardingShow',
+    onboardingHide: 'onboardingHide',
+    onboardingHasHeader: 'onboardingHasHeader'
```
jeremywiebe: What's the intent for this event? I don't see it used anywhere.

native/app/global/app-event-names.js
```diff
@@ -0,0 +1,7 @@
+const RpcEventNames = {
```
jeremywiebe: I think it'd be good to have a bit cleaner separation between "events" and "rpc methods".  Do you think we could separate into `global/app-event-names` and `app-rpc-method-names`?  I find it a bit misleading to have the structure called `RpcEventNames` but them not be events.

native/onboarding/.gitignore
```diff
@@ -0,0 +1,15 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
+
+# dependencies
+node_modules
```
jeremywiebe: I think most of these already exist in the root `.gitignore` file.  I'd say we could remove them if they do, and properly work in subfolders.

native/onboarding/src/components/Onboarding.scss
```diff
@@ -0,0 +1,47 @@
+
+.c-carousel, .carousel-item {
+  height: 100%;
+}
+
+.item-title {
+  margin: 10px 46px 0px 46px;
+  font-size: 20px;
+  line-height: 24px;
+  color: #333;
```
jeremywiebe: Is it possible/worthwhile to pull in the colour definitions from the web side .scss file(s)?

native/onboarding/config/webpack.config.app.js
kerrmarin: Since it's a 1-off process, how about instead of that I create a `build-dependencies.sh` file and in `npm run build` for the main `native` project I change it from just calling webpack to calling this new shell file? And in the shell file I run both webpack and the `npm run build` inside the onboarding?

native/app/config/onboardingConfig.js
kerrmarin: Good idea 👍

native/app/controllers/onboarding/onboardingModalController.js
kerrmarin: This is badly named -- If you want to show the onboarding after the first run then you should call this method with the `{forced: true}` param. I've changed the name.

native/onboarding/src/components/Onboarding.scss
kerrmarin: I've removed this one and added the class it represents, but I couldn't find a clean way to include the rest of the colours in this file

native/app/controllers/onboarding/onboardingController.js
kerrmarin: I don't have a strong preference either way, I've changed it 👍

native/app/controllers/onboarding/onboardingController.js
kerrmarin: I think this was a copy-paste error. I've removed it now.

native/onboarding/config/webpack.config.app.js
jeremywiebe: @kerrmarin I think that'd work!

kerrmarin: Done :)

jeremywiebe:
I think we should make the status bar dark at least for onboarding. We can swap it back to light text when onboarding dismisses.

http://astro.mobify.com/latest/api-reference/application/#set-status-bar-dark-text

</img>

native/onboarding/package.json
```diff
@@ -0,0 +1,78 @@
...
+    "webpack-manifest-plugin": "1.1.0",
+    "whatwg-fetch": "1.0.0",
+    "spline-scss": "2.2.0",
+    "susy": "2.2.12"
+  },
+  "dependencies": {
+    "react": "^15.4.1",
+    "react-dom": "^15.4.1",
+    "progressive-web-sdk": "0.10.10",
+    "mobify-progressive-app-sdk": "mobify/astro#pwa-support",
```
jeremywiebe: Since we reference the `mobify-progressive-app-sdk` in the parent `node_modules` dir, we don't need this anymore, do we?

native/app/onboarding/onboardingModalController.js
```diff
@@ -80,4 +83,4 @@ OnboardingModalController.prototype.canGoBack = function() {
     return this.onboardingController.canGoBack()
 }
 
-export default OnboardingModalController
+export {OnboardingModalController, OnboardingModalEvents}
```
jeremywiebe: In my head it makes sense that the `OnboardingModalController` be exported as the default and anything else as a named export.  We can do that like this:
```
export {OnboardingModalEvents}
export default OnboardingModalController
```

jeremywiebe:
Circle's failing because of the port 9222 issue.  :(  Merging!

👍  @kerrmarin

PR #232, opened 2017-01-11
Port the actions and reducers to `redux-actions`

DanielBrookRoberge:
The `redux-act` library is very helpful but is not very widely used. A similar library made by people close to the core Redux team is `redux-actions`.  This PR ports the scaffold from `redux-act` to `redux-actions`. 

A benefit to `redux-actions` is that its actions and reducers support a separate path for errors in each action, triggered by passing an `Error` object as the action parameter, which will cut down on the number of distinct actions in our application. It is also 13% smaller minified and gzipped. 

The main user-visible changes are in the reducer. The `handleActions` function takes the place of the `createReducer` function, and the individual action handler functions are now passed the full action object rather than just the payload. The utility function `createAction` has been updated compatibly to use `redux-action` rather than `redux-act`.

NOTE: This breaks compatibility with the container generator, prior to a follow-up SDK PR.

 **JIRA**: N/A (preparatory for middleware work)

## Changes
- Replace `redux-actions` with `redux-act`

## How to test-drive this PR
- Run the tests
- Preview and see that the site still functions as before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "react-redux": "4.4.5",
     "react-router": "2.7.0",
     "redux": "3.6.0",
-    "redux-act": "1.1.0",
     "redux-actions": "1.2.0",
     "redux-thunk": "2.1.0",
     "sw-toolbox": "3.4.0",
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "react-router": "2.7.0",
     "redux": "3.6.0",
     "redux-act": "1.1.0",
+    "redux-actions": "1.2.0",
     "redux-thunk": "2.1.0",
     "sw-toolbox": "3.4.0",
     "whatwg-fetch": "1.0.0"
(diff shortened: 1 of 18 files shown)


DanielBrookRoberge:
I'm going to merge this into the data architecture staging branch to keep my branching strategy from getting too complex. I'll follow up on any review comments later.

package.json
```diff
@@ -14,12 +14,13 @@
...
     "es6-promise": "4.0.5",
     "immutable": "3.8.1",
+    "lodash.frompairs": "4.0.1",
     "react": "15.4.1",
     "react-dom": "15.4.1",
     "react-redux": "4.4.5",
     "react-router": "2.7.0",
     "redux": "3.6.0",
-    "redux-act": "1.1.0",
+    "redux-actions": "1.2.0",
```
mikenikles: Once we update the generator to use `redux-actions`, we have to make sure to mark the next SDK release with a breaking change and add a link to this PR right here so that anyone who wants to upgrade can see what changes they need to make to their project.

PR #214, opened 2016-12-20
🚀  Astro integration

jeremywiebe:
This PR lands basic support for Astro integration. It forms the basis for Astro becoming a part of the Mobify Platform Scaffold. Future PRs will build on this one so that we can build/bring in the new Progressive App scaffold one piece at a time.

This PR is primarily a folder re-organization and changes to make CircleCI work. It is the base branch for https://github.com/mobify/progressive-web-scaffold/pull/236 which is the PR that forms our main "`develop`" branch until we get everything merged and working on the real `develop` branch. Once we land this branch we will re-target https://github.com/mobify/progressive-web-scaffold/pull/236 to use `develop` as the base branch.

 **JIRA**: https://mobify.atlassian.net/browse/HYB-962
 **Linked PRs**: https://github.com/mobify/progressive-web-scaffold/pull/236, https://github.com/mobify/astro/pull/744, https://github.com/mobify/astro-scaffold/pull/155

## Outstanding questions
There are several questions that we need to address while we integrate the app and web platforms into a single scaffold (ie. in this PR):
- [x] How will we configure CircleCI?
       _[@jeremywiebe] There will be very little configuration needed for progressive apps as we primarily use BuddyBuild for CI. Some of this might be dependent on our decision of how to manage branching (only do certain build actions for web/app depending on the branch)._
- [x] What will the branching strategy be like?
       _[@jeremywiebe] All branches will be taken from `develop`. PRs get labelled with `web` or `app` or `platform` depending on who needs to review the code._
- [x] Will we release everything at once or still be able to release app vs web features individually. How?
       _[@jeremywiebe] At this point the scaffold is a "continuous release" type of project (we are always generating based on whatever is on `master`). I think in the near future we should move to deprecate the `develop` branch and use release branches to host development of different teams (as we've discussed already). This will keep us out of each other's hair and prevent unreleasable code from being merged to `master`._
- [x] How will we mark web / app PRs to make sure the right people review the right PRs?
       Maybe use labels? `web` and `astro`?
       _[@jeremywiebe] Yep, labels!_

## Changes
- Trigger Astro event in app-provider's `onChange` if running in an Astro app.

## How to test-drive this PR
- Run the tests: `cd web && npm test`
- Ensure CircleCI passes
- Run the web scaffold and make sure it works

PR #236, opened 2017-01-12
Astro: Base for Merlins implementation

jeremywiebe:
Initially, it lands basic support for Astro integration. It forms the basis for Astro becoming a part of the Mobify Platform Scaffold. Future PRs will build on this one so that we can build/bring in the new Progressive App scaffold one piece at a time.

JIRA: https://mobify.atlassian.net/browse/HYB-962
Linked PRs: https://github.com/mobify/progressive-web-scaffold/pull/214, mobify/astro#744, mobify/astro-scaffold#155

# Outstanding questions

There are several questions that we need to address while we integrate the app and web platforms into a single scaffold (ie. in this PR):

- [x] How will we configure CircleCI?
       [@jeremywiebe] There will be very little configuration needed for progressive apps as we primarily use BuddyBuild for CI. Some of this might be dependent on our decision of how to manage branching (only do certain build actions for web/app depending on the branch).
- [x] What will the branching strategy be like?
  - Docs at https://github.com/mobify/branching-strategy/blob/master/continuous-deployment.md#platform-repo.
- [ ] Will we release everything at once or still be able to release app vs web features individually. How?
- [x] How will we mark web / app PRs to make sure the right people review the right PRs?
  - `web`, `app` and `platform` labels are available in this repo now.

# Changes

Trigger Astro event in app-provider's onChange if running in an Astro app.


web/app/containers/onboarding/container.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+
+import CarouselItem from 'progressive-web-sdk/dist/components/carousel/carousel-item'
+import Carousel from 'progressive-web-sdk/dist/components/carousel'
+import Button from 'progressive-web-sdk/dist/components/button'
+
+const Onboarding = ({carouselData}) => {
+    const carouselItems = Object.keys(carouselData).map((key, index) => {
+        return <OnboardingScreen data={carouselData[key]} key={index} index={index} />
```
jnurse: Do all these props need to be nested under data? or could they be passed directly to the component like this?
`<OnboardingScreen {...carouselData[key]} key={index} index={index} />`

web/app/containers/onboarding/container.js
```diff
@@ -0,0 +1,55 @@
...
+    })
+    return (
+        <Carousel>
+            {carouselItems}
+        </Carousel>
+    )
+}
+
+const OnboardingScreen = (props) => {
+    const data = props.data
```
jnurse: Typically we destructure props in the function definition like this:
`const OnboardingScreen = ({data, key}) => {`

web/app/containers/onboarding/container.js
jnurse: Also I don't think we need to pass both a `key` and an `index` into this component. Just a key should be good enough, then you can use the `key` prop bellow instead of `index`.

web/app/containers/pdp/actions.js
```diff
@@ -22,6 +25,15 @@ export const process = ({payload}) => {
     return receiveData({[urlToPathKey(url)]: parsed})
 }
 
+export const goToCheckout = () => (dispatch) => {
+    dispatch(closeModal(PDP_ITEM_ADDED_MODAL))
+    if (isRunningInAstro) {
+        Astro.trigger('open:cart-modal')
+    } else {
+        // open web checkout
```
jnurse: The go to checkout button was fixed in `develop` to correctly navigate to the checkout, we should add something here to make sure it still works for web.

web/app/containers/checkout-header/container.js
```diff
@@ -7,33 +7,38 @@ import {getIsLoggedIn} from '../app/selectors'
...
-                            innerClassName="u-color-neutral-10"
-                        >
-                            <Icon name="user" className="u-margin-end-sm" />
-                            <span>Sign in</span>
-                        </Button>
-                    </div>
-                }
-            </HeaderBar>
-        </header>
+        <div>
```
jnurse: Is there a reason why we're wrapping this in a div now?

web/app/store/cart/actions.js
```diff
@@ -20,6 +22,22 @@ const baseHeaders = {
 
 export const receiveCartContents = utils.createAction('Received Cart Contents')
 
+export const openSignIn = () => {
+    if (isRunningInAstro) {
+        Astro.trigger('sign-in:clicked')
+    } else {
+        // web open signIn
```
jnurse: We should update these actions so they navigate to the right page on web.

web/app/containers/onboarding/container.js
kerrmarin: Didn't know about `...var` -- I've changed this to use the spread shorthand

web/app/containers/onboarding/container.js
kerrmarin: 👍  done

web/app/containers/checkout-header/container.js
kerrmarin: No idea why this was added, so I've removed it.

web/app/app-provider.jsx
```diff
@@ -5,27 +5,48 @@ import {Provider} from 'react-redux'
...
-import {Cart, CheckoutConfirmation, CheckoutPayment, CheckoutShipping, Home, Login, PLP, PDP} from './containers/templates'
+import {Cart, CheckoutConfirmation, CheckoutPayment, CheckoutShipping, Home, Login, ProductList, ProductDetails} from './containers/templates'
 import CheckoutHeader from './containers/checkout-header/container'
 import CheckoutFooter from './containers/checkout-footer/container'
 
+import {getURL} from './utils/utils'
+import {isRunningInAstro, pwaNavigate} from './utils/astro-integration'
+
+// We define an initial OnChange as a no-op for non-Astro use
+let OnChange = () => {}
```
jnurse: We've got a noop function in the SDK you could use here instead (you can find it under progressive-web-sdk/dist/utils/utils)

web/app/utils/astro-integration.js
```diff
@@ -0,0 +1,19 @@
+import Astro from '../vendor/astro-client'
```
DanielBrookRoberge: It looks like you've removed this file, is that true?

web/app/app-provider.jsx
```diff
@@ -9,10 +9,30 @@ import {Cart, CheckoutConfirmation, CheckoutPayment, CheckoutShipping, Home, Log
...
+        if (nextState.location.action === 'POP') {
+            callback()
+            return
+        }
+
+        pwaNavigate({url: getURL(nextState)}).then(() => {
+            callback()
+        })
+    }
+}
```
jeremywiebe: @DanielBrookRoberge Yes, we actually still need it here. The `Template` doesn't provide an asynchronous notification system that blocks the transition until `callback()` is called. Astro needs that in order to have the web side wait until the app side finished snapshotting the component that is unmounting before it transitions.

DanielBrookRoberge: Does this mean that `Template` shouldn't be calling `Astro.trigger('pwa-navigate', url)`?

DanielBrookRoberge: Also, probably there should be a comment about why this is done here (since you want to do this before the old component is unmounted, while everything that touches the Redux store needs to be done afterwards)

web/app/containers/cart/container.js
```diff
@@ -51,10 +54,11 @@ const EmptyCartContents = ({hide}) => {
 }
 
 EmptyCartContents.propTypes = {
-    hide: PropTypes.bool
+    hide: PropTypes.bool,
 }
 
 const Cart = ({contentsLoaded, hasItems}) => {
+    Astro.trigger('checkout:disable-alert')
```
kerrmarin: I think this should be fine just firing once when the component is mounted, I'll update.

web/app/containers/onboarding/container.js
```diff
@@ -0,0 +1,54 @@
...
+                    }
+                    {actionButton &&
+                        <Button className="c--secondary u-flex" text={actionButton.title} onClick={actionButton.action} />
+                    }
+                </div>
+            </div>
+        </CarouselItem>
+    )
+    return item
+}
```
kerrmarin: When I moved it up the lint error appeared, so I fixed it 👍

web/app/utils/astro-integration.js
kerrmarin: Mhm, no that file is still there? Why do you think it's been removed?

kerrmarin: Ah, I get it. It used to be the non-minified version that was included in this, now it's the minified version. Not sure why this has changed - do we have a preference for `vendor` scripts?

web/app/containers/onboarding/container.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+
+import CarouselItem from 'progressive-web-sdk/dist/components/carousel/carousel-item'
+import Carousel from 'progressive-web-sdk/dist/components/carousel'
+import Button from 'progressive-web-sdk/dist/components/button'
+
+const Onboarding = ({carouselData}) => {
+    const carouselItems = Object.keys(carouselData).map((key, index) => {
+        return <OnboardingScreen {...carouselData[key]} key={index} />
+    })
```
kerrmarin: What do you mean by `I think it's probably clearer to include this map inside the JSX.` ?

DanielBrookRoberge: To do, instead, 
```javascript
const Onboarding = ({carouselData}) => {
    return (
        <Carousel>
            {Object.keys(carouselData).map((key) => <OnboardingScreen {...carouselData[key]} />)}
        </Carousel>
    )
}
```

kerrmarin: Ah, that makes sense, yes I've updated this now 👍

DanielBrookRoberge:
Thanks for the changes, I'm now :+1:

PR #238, opened 2017-01-13
Astro: Astro integration PR review changes

jeremywiebe:
A few small changes addressing PR review comments from @DanielBrookRoberge (see #236 )

1) Originally I implemented this by pushing the flag into the Redux store, but that isn't optimal.  Reworked to just use a module that exports the flag and pull it in where needed.  The Header and Footer components still have a property that can be passed in instead of looking at the module's flag directly.

2) Moved the `pwaNavigate` into the astro-integration module (see `app-provider.jsx`)

 **JIRA**: N/A
 **Linked PRs**: https://github.com/mobify/progressive-web-scaffold/pull/236

## Changes
- Removed `isRunningInAstro` from the store
- Implemented `astro-integration` module for the flag (and future Astro behaviours)
- Updated tests

## How to test-drive this PR
- Run `npm run test`
- Run the app
   - Verify that the header and footer are suppressed in app
   - Verify that the PLP page doesn't include the page title


kerrmarin:
Works as expected 👍

PR #244, opened 2017-01-16
Refactor header

DanielBrookRoberge:
In preparation for the data architecture changes, I discovered some awkwardness in the Header container. This PR refactors the container to be simpler and easier to convert to the new architecture.

 **JIRA**: N/A

## Changes
- Split the Header component into partials
- Fix bug in registering and unregistering the scroll listener
- Change the header state to only include the relevant part of the cart data.

## How to test-drive this PR
- Preview Merlins
- See that the header works correctly (changing when scrolling, displaying a badge if and only if there are cart items)


app/containers/header/container.js
```diff
@@ -4,103 +4,58 @@ import throttle from 'lodash.throttle'
...
-    }
-}
+import {HeaderBar} from 'progressive-web-sdk/dist/components/header-bar'
+
+import NavigationAction from './partials/navigation-action'
+import HeaderTitle from './partials/header-title'
+import StoresAction from './partials/stores-action'
+import CartAction from './partials/cart-action'
+
+const HEADER_HEIGHT = 52
```
mikenikles: @jeffkamo Can you think of a way to dynamically calculate this? I worry we forget to update that constant when we refactor the header in the future and the height changes.

jeffkamo: It would probably fine to directly reference the height of the `.t-header` element.

mikenikles: Good call. We add the `handleScroll` function as an event handler in the `componentDidMount` function. At that time, we have access to the height of the `t-header` element, could set a class variable (is this too ugly?) and use that in the `handleScroll` function.

mikenikles:
👍

PR #245, opened 2017-01-16
Remove Immutable from React components

DanielBrookRoberge:
This PR adds the `selectorToJS` utility function which will handle the Immutable -> JS conversion in a way that preserves the JS object. This, combined with writing the `mapStateToProps` function as a Reselect selector, allows `react-redux` to determine on its own whether a connected component needs to be updated.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1032

## Changes
- Add utility functions for handling Immutable using Reselect
- Convert all the containers to use them
- Convert a number of containers to stateless components.

## How to test-drive this PR
- Run the tests
- Preview Merlins and see that everything works.

### DIFF
--- a/app/containers/plp/container.js
+++ b/app/containers/plp/container.js
@@ -36,10 +36,6 @@ const renderNoResults = (bodyText) => {
 }
 
 class PLP extends React.Component {
-    shouldComponentUpdate(nextProps) {
-        return this.props.plp !== nextProps.plp || !Immutable.is(nextProps.products, this.props.products)
-    }
-
     render() {
         const {
             hasProducts,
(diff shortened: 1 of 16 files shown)


app/containers/plp/container.js
```diff
@@ -108,23 +100,10 @@ PLP.propTypes = {
...
-        })
-        return {
-            products,
-            plp: routedPlp
-        }
-    }
-)
+const mapStateToProps = createStructuredSelector({
+    products: selectorToJS(selectors.getPlpProducts),
+    plp: selectorToJS(selectors.getSelectedPlp)
```
mikenikles: No change required for this PR, but as we continue with the refactoring of the state, I'd like to discuss how we can cleanly separate `ui` vs `data` selectors. Maybe it's as simple as `getSelectedPlpUi`. There should always be maximum one `...Ui` selector for a given component, I think. However, a component could select multiple properties from the `data` section of the state.

DanielBrookRoberge: We'll see how this ends up looking after the next two/three steps; I can see partials subscribing to part of the `ui` state; it continues to make sense to me to define that on a template-by-template basis.

PR #248, opened 2017-01-17
WEB-1071: When the mini-cart is empty, the button should say ‘Continu…

twangtina:
When the mini-cart is empty, the button should say ‘Continue Shopping’

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1071

## Changes
- When the mini-cart is empty, the button should say ‘Continue Shopping’
- When there's items in the mini-cart, the button should say ‘Go To Checkout’

## How to test-drive this PR
- Follow the instructions in the ticket: https://mobify.atlassian.net/browse/WEB-1071

### DIFF
--- a/app/containers/mini-cart/container.js
+++ b/app/containers/mini-cart/container.js
@@ -93,6 +93,26 @@ class MiniCart extends React.Component {
         )
     }
 
+    renderButton(hasItems) {
+        if (hasItems) {
+            return (
+                <div className="u-padding-top-lg u-flex-none">
+                    <Button href="#" className="c--primary u-width-full u-text-uppercase">
+                        Go To Checkout
+                    </Button>
+                </div>
+            )
+        } else {
+            return (
+                <div className="u-padding-top-lg u-flex-none">
+                    <Button href="#" className="c--primary u-width-full u-text-uppercase">
+                        Continue Shopping
+                    </Button>
+                </div>
+            )
+        }
+    }
+
     render() {
         const {miniCart, closeMiniCart} = this.props
         const {cart, contentsLoaded, isOpen} = miniCart.toJS()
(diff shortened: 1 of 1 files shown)


jeffkamo:
@twangtina quick heads up! When creating PRs for the Product Team projects, be sure to stick to the PR templates instead of the older CST style templates. Those look something [like this](https://github.com/mobify/progressive-web-scaffold/blob/develop/.github/PULL_REQUEST_TEMPLATE.md) (and they pre-populate the PR when being created)

The main reason is that the PR status is designated using the PR's labels, and the reviewers are also designated using Github's "Reviewers" functionality! Like so:

screen shot 2017-01-17 at 2 21 27 pm

Lastly, the "+1" as a todo is no longer necessary, as that is handled when the reviewers perform a [Github PR review](https://help.github.com/articles/about-pull-request-reviews/).

app/containers/mini-cart/container.js
```diff
@@ -93,6 +93,26 @@ class MiniCart extends React.Component {
...
+        } else {
+            return (
+                <div className="u-padding-top-lg u-flex-none">
+                    <Button href="#" className="c--primary u-width-full u-text-uppercase">
+                        Continue Shopping
+                    </Button>
+                </div>
+            )
+        }
+    }
```
jeffkamo: Nice, good logic! One thing I'll suggest is that there is a way to make this smaller and more concise using pretty much the same logic.

Consider this: the markup in both results are identical. What's different is just the children of the Button (the part that reads either "Go To Checkout" or "Continue Shopping". With that in mind, it's easier to do this...

```jsx
renderButton() {
    return (
        <div className="u-padding-top-lg u-flex-none">
            <Button href="#" className="c--primary u-width-full u-text-uppercase">
                {hasItems ? 'Go To Checkout' : 'Continue Shopping'}
            </Button>
        </div>
    )
}
```

We're able to leverage ternary operators (the `condition ? truthyResult : falseyResult` part) to make selecting the text that much easier without writing tons and tons of HTML.

jeffkamo: I'm not sure how familiar you are with ternary operators, but if you aren't I would highly recommend [brushing up on the concept](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)! We use ternary operators (and other similar conditional operators) very regularly [in react](https://facebook.github.io/react/docs/conditional-rendering.html).

twangtina: Makes sense. Thanks for the feedback!

twangtina: I've made the change. How does it look now @jeffkamo?

jeffkamo:
@twangtina 👍

PR #249, opened 2017-01-17
Connect further down in Header, Footer, Home, and PDP

DanielBrookRoberge:
Splitting this stage of the refactor into smaller pieces because it's a big one. This PR re-architects the structure of the Header, Footer, Home, and PDP containers so that the Redux store is accessed closer to where the resulting data is used. Footer, Home, and PDP are now functional components with no props, making the top-level code very clean.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1032

## Changes
- Connect the Redux store lower down the component tree

### DIFF
--- a/app/containers/footer/container.js
+++ b/app/containers/footer/container.js
@@ -17,30 +17,16 @@ const social = [
     ['http://www.youtube.com/#TODO', 'static/svg/youtube.svg', 'Youtube'],
 ]
 
-class Footer extends React.Component {
-    constructor(props) {
-        super(props)
+const Footer = ({footer, submitNewsletter}) => {
+    const {navigation, newsletter} = footer
 
-        this.onSubmitNewsletter = this.onSubmitNewsletter.bind(this)
-    }
-
-    onSubmitNewsletter(data) {
-        const {method, action} = this.props.footer.newsletter
-        this.props.submitNewsletter(method, action, data)
-    }
-
-    render() {
-        const {footer} = this.props
-        const {navigation, newsletter} = footer
-
-        return (
-            <footer className="t-footer">
-                <FooterNewsletterSubscription newsletter={newsletter} onSubmit={this.onSubmitNewsletter} />
-                <FooterSocialIcons social={social} />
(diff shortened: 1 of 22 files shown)


app/containers/footer/container.js
```diff
@@ -17,53 +11,14 @@ const social = [
...
-    footer: selectorToJS(selectors.getFooter)
-})
-
-const mapDispatchToProps = {
-    submitNewsletter: actions.signUpToNewsletter
+const Footer = () => {
+    return (
+        <footer className="t-footer">
+            <FooterNewsletterSubscription />
+            <FooterSocialIcons social={social} />
```
jeffkamo: The socials array could probably be moved to the FooterSocialIcons partial. Doesn't look like it's used anywhere else anyway.

app/utils/selector-utils.js
```diff
@@ -8,7 +8,7 @@ export const createImmutableComparingSelector = createSelectorCreator(
 
 export const selectorToJS = (selector) => createImmutableComparingSelector(
     selector,
-    (raw) => raw.toJS()
+    (raw) => { return raw ? raw.toJS() : null }
```
mikenikles: Out of curiosity, what prompted this change? In other words, when would `raw` be `undefined`?

app/containers/footer/container.js
DanielBrookRoberge: Fixed.

app/utils/selector-utils.js
DanielBrookRoberge: There are some places where an array or object is nullable with different semantics than simply having an empty object/array. So I'm not expecting `raw` to be `undefined`, but I am expecting it to occasionally be `null`. (In addition, I tripped over this several times before just going in and changing the utility, and found it difficult to debug, with an unhelpful message and stack trace)

mikenikles: Thanks, that makes sense.

mikenikles:
Let's get this merged.

PR #254, opened 2017-01-20
WEB-1072: Add Breadcrumbs Component to PDP

twangtina:
Add Breadcrumbs Component to PDP

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1072

## Changes
- Add Breadcrumbs component to PDP
- And styled it

## How to test-drive this PR
- Please follow the instructions in the ticket: https://mobify.atlassian.net/browse/WEB-1072


app/containers/catalog/products/reducer.js
```diff
@@ -13,6 +13,7 @@ import {PLACEHOLDER} from '../../app/constants'
 
 export const initialState = Immutable.fromJS({
     [PLACEHOLDER]: {
+        breadcrumbs: [],
```
jeffkamo: I'm not sure this is the best place for the breadcrumb state. The breadcrumbs have less to do with the product itself, and more to do with the current state of the app (i.e. what page you're currently on).

Put another way, what if one of these products is being viewed somewhere else, such as the PLP, or in the Cart, etc. Does that product's internal breadcrumb value still make sense?

app/containers/pdp/_pdp.scss
```diff
@@ -6,6 +6,38 @@
     // wrapper element.
 }
 
+// Breadcrumbs
+// ---
+//
+// 1. Visually hide the label
+
+.t-pdp__breadcrumbs {
```
jeffkamo: We already know that Breadcrumbs are a global component, but you are styling them as if they only exist on the PDP page.

Since it's coming from the SDK, we know that it already [has some base styles](http://docs.mobify.com/progressive-web/0.0.1/components/#!/Breadcrumbs) that we can import into our styles.

Once those base styles are in, you can probably move what's left of this code to the appropriate location for SDK theme styles. Let me know if you're not sure where those are put.

app/containers/pdp/partials/pdp-heading.jsx
```diff
@@ -1,8 +1,15 @@
...
 import SkeletonBlock from 'progressive-web-sdk/dist/components/skeleton-block'
+import Breadcrumbs from 'progressive-web-sdk/dist/components/breadcrumbs'
 
-const PDPHeading = ({title, price}) => (
+const PDPHeading = ({breadcrumbs, title, price}) => (
     <div className="t-pdp-heading u-padding-md u-box-shadow u-position-relative u-z-index-1">
+        {breadcrumbs ?
+            <Breadcrumbs className="t-pdp__breadcrumbs u-margin-bottom-md" items={breadcrumbs} />
+        :
+            <SkeletonBlock width="50%" height="12px" className="u-margin-bottom-md" />
```
jeffkamo: I think we might be able to get away with not having a skeleton by using the breadcrumb's initialState.

For example, it's reasonable to assume that breadcrumbs will always, always start with `Home >`. Then, once the state receives content from desktop, the rest of the content can be filled in automatically.

Thoughts?

jeffkamo: Mitch recommends `Home > ...`

twangtina:
@jeffkamo I've addressed your above feedback. How does it look now?

app/containers/pdp/partials/pdp-heading.jsx
```diff
@@ -1,8 +1,18 @@
...
+import Link from 'progressive-web-sdk/dist/components/link'
 
-const PDPHeading = ({title, price}) => (
+const PDPHeading = ({breadcrumbs, title, price}) => (
     <div className="t-pdp-heading u-padding-md u-box-shadow u-position-relative u-z-index-1">
+        <div className="t-pdp__breadcrumbs u-margin-bottom-md">
+            {breadcrumbs ?
+                <Breadcrumbs items={breadcrumbs} />
+            :
+                <Link href="/">Home &gt; ...</Link>
```
jeffkamo: Right now there is a spacing inconsistency around the `>` character when you compare the "loading" and "loaded" versions of the breadcrumb.

As you can see in this chunk of code that the "loading" and "loaded" versions use different markup. One uses `Breadcrumbs`, the other uses a `Link`.

Can you find a way to make this work using just `Breadcrumbs`? Hint: you  shouldn't need the ternary statement :)

jeffkamo: Here's what I mean btw:

screen shot 2017-01-25 at 5 31 08 pm

mlworks: @twangtina  Let's try and reuse the component as much as possible that way you don't have to write any new custom markup / styles.

For example: `<Breadcrumbs items={[{href: '/', text: 'Home'}, {text: '...'}]} />`

PR #256, opened 2017-01-23
Finish connecting the store further down

DanielBrookRoberge:
This PR connects the store deeper on the remaining components, while refactoring them to reflect our best practices. This has led to a lot of modularization, plus deduplication of code in the Login container.

After this PR, no further changes to the React components are expected, outside perhaps their arrangement in the filesystem.

## Changes
- Split up so many components

## How to test-drive this PR
- Run the tests
- Preview and see that nothing has changed.


app/containers/cart/container.js
```diff
@@ -10,162 +9,78 @@ import {Grid, GridSpan} from '../../components/grid'
...
-        })
-        return (
-            <div>
-                <GridSpan tablet={{span: 6, pre: 1, post: 1}} desktop={{span: 7}}>
-                    <CartProductList
-                        cart={cart}
-                        onSaveLater={this.openWishlistModal}
+import CartItems from './partials/cart-items'
+
+const EmptyCartContents = ({hide}) => {
```
jvoll: likely newb question... I don't see a `render` function which I suspect is because these are functional components? Though, if I'm right, would it make sense to put the normal cart and the empty cart into partials?

app/containers/mini-cart/partials/mini-cart-header.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+MiniCartHeader.propTypes = {
+    closeMiniCart: PropTypes.func
+}
+
+const mapDispatchToProps = {
+    closeMiniCart: stripEvent(actions.closeMiniCart)
+}
+
+export default connect(() => ({}), mapDispatchToProps)(MiniCartHeader)
```
DanielBrookRoberge: Yeah, that does seem a little redundant on reflection, since the `MiniCart` component needs to provide the function to the `Sheet` anyways. I'll change it back.

app/containers/cart/container.js
fractaltheory: @jvoll that's right - the return statement of a functional (stateless) component is effectively the `render` method.

app/containers/mini-cart/selectors.js
```diff
@@ -10,3 +10,12 @@ export const getMiniCart = createSelector(
...
 export const getMiniCartIsOpen = createGetSelector(getMiniCart, 'isOpen')
+export const getMiniCartItems = createGetSelector(getCartObject, 'items')
+export const getMiniCartSubtotal = createGetSelector(getCartObject, 'subtotal')
+export const getMiniCartHasItems = createSelector(
+    getMiniCartItems,
+    (items) => items.size > 0
+)
+export const getMiniCartSummaryCount = createGetSelector(getCartObject, 'summary_count')
+export const getSubtotalExcludingTax = createGetSelector(getCartObject, 'subtotao_excl_tax')
+export const getSubtotalIncludingTax = createGetSelector(getCartObject, 'subtotao_incl_tax')
```
DanielBrookRoberge: I'm not sure how to test this, but it's definitely something to keep in mind.

PR #257, opened 2017-01-24
Add QA selector to remove item button in cart

ellenmobify:
Adds a class to the Remove button in the cart for end-to-end testing purposes.

## Changes
- Adds `qa-cart-remove-item` class to the Remove button in the cart

## How to test-drive this PR
- Checkout this branch
- `npm install`
- `npm run dev`
- Add an item to cart
- Go to https://www.merlinspotions.com/checkout/cart/ and inspect the 'Remove' button. Verify that it has class `qa-cart-remove-item`

### DIFF
--- a/app/containers/cart/partials/cart-product-list.jsx
+++ b/app/containers/cart/partials/cart-product-list.jsx
@@ -96,7 +96,7 @@ const CartProductList = ({cart, onSaveLater}) => {
                             </Button>
 
                             <Button
-                                className="u-text-small u-color-brand"
+                                className="u-text-small u-color-brand qa-cart-remove-item"
                                 innerClassName="u-padding-end-0 u-padding-bottom-0"
                             >
                                 Remove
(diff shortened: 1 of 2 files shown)


app/containers/cart/partials/cart-product-list.jsx
```diff
@@ -96,7 +96,7 @@ const CartProductList = ({cart, onSaveLater}) => {
                             </Button>
 
                             <Button
-                                className="u-text-small u-color-brand"
+                                className="u-text-small u-color-brand qa-cart-remove-item"
```
jeffkamo: I would suggest this className be `qa-cart__remove-item` to better show the relationshop between the parent template and the subject element at hand.

jeffkamo: (this also aligns a bit better with our CSM classnaming convention)

jeffkamo:
@jasecode maybe not that necessary. Developers probably shouldn't be changing classnames for any unpredictable reason. And you don't want to accidentally match similarly named classes, like `qa-cart__remove-options` or something.

PR #262, opened 2017-01-25
iPhone 5 pdp coverage fix

sshin-mobify:
iPhone 5 pdp coverage fix
- when using small devices such as iPhone 5 (width <= 320) t-plp__item-added-modal coverage is set to 60% in order to show all of its contents fully

 **JIRA**: https://mobify.atlassian.net/browse/HYB-983

## Changes
- on the container implemented a logic that checks for window size and sets the coverage to 40% if the device width is less than 320
- ideally use breakpoint to set the coverage % 

## Todo
- [x] review with someone who is familiar with scss 

## How to test-drive this PR
- run the simulator on iPhone 5
- select a product to buy and add it to cart
- on the slide up modal all of the contents should be delayed


web/app/containers/pdp/container.js
```diff
@@ -40,6 +40,8 @@ class PDP extends React.Component {
             closeItemAddedModal
         } = this.props
 
+        const coverage = window.innerWidth > 320 ? "50%" : "60%"
```
jeremywiebe: Two thoughts:
- This feels very iOS-specific. `320` is an iOS width, how will this work out on an Android device?
- I think we should focus on the height, not the width. Related to my previous point, what if we have an Android device that is a bit wider than `320` but short and we have the same problem we're trying to fix?

jeremywiebe:
@sshin-mobify Sorry, I tried on an iPhone 4s and I think we need to adjust the height a bit more. 

</img>

jeremywiebe:
@sshin-mobify Some linting errors in the CircleCI build... can you fix those up, please?

jeremywiebe:
👍

PR #239, opened 2017-01-13
Data Architecture 2.0

DanielBrookRoberge:
This is the meta branch for the preliminary work for the middleware bridge, as described in https://docs.google.com/a/mobify.me/document/d/1yHzxVOtsnA-GVp2llcIT6HyRd5pHJw_nS0xG0SC0ORU/edit?usp=sharing

- [x] `redux-act` => `redux-actions` (#232)
- [x] Selectorize the access to the store (#241)
- [x] Remove Immutable from the components (#245)
- [x] Break up the component access to the store (#246)
- [x] Move the connection down the tree as appropriate (#249 and #256) 
- [x] Decouple reducers from parsers (#264)
- [x] Move page change logic into a Template higher-order component (#276)
- [x] Rearrange the store to match the shape of the data better (#261, #266, #277, #283)
- [x] Add better utilities for creating selectors. (#284)
- [x] Run a full regression test by pushing and testing a production bundle
- [ ] Merge this PR (i.e. #239)
- [ ] Rearrange the file structure breaking apart the `containers` dir
- [ ] Organize/simplify the selectors to match the new store and components.
- [ ] Schematize the store (a little in #271)
- [ ] Write tests as needed
- (not in scope here) Move code to the SDK

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1032
 **Linked PRs**: (links to corresponding PRs, optional)

## Changes
- (change1)

## How to test-drive this PR
- (necessary config changes)
- (necessary corresponding PRs)
- (how to access the new / changed functionality -- fixtures, URLs)

PR #264, opened 2017-01-25
Decouple parsers from reducers. 

DanielBrookRoberge:
This PR moves all parsing calls from reducers to actions. For most pages, this is done through a wrapper function called `process` in each action file, which takes the `onPageReceived` action as input and returns a new page-specific action with the parsed data. 

Ultimately, this is a step towards two things:
 - Building an API barrier between the reducers and actions
 - Reducing the processing in reducers as much as possible

To the second end, most of the new actions include a payload that can be directly merged with the state. This PR contains utilities to make this common case simpler.

 **Linked PRs**: #239 

## How to test-drive this PR
- Run the tests
- See that all pages load content as before

### DIFF
--- a/app/utils/reducer-utils.js
+++ b/app/utils/reducer-utils.js
@@ -1,1 +1,7 @@
 export const mergePayload = (state, {payload}) => state.mergeDeep(payload)
+
+export const mergeHandlersFor = (...actions) => {
+    const handlers = {}
+    actions.forEach((action) => { handlers[action] = mergePayload })
+    return handlers
+}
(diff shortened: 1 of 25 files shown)


app/utils/reducer-utils.js
```diff
@@ -0,0 +1,7 @@
+export const mergePayload = (state, {payload}) => state.mergeDeep(payload)
+
+export const mergeHandlersFor = (...actions) => {
```
mikenikles: Let's use `mergePayloadForActions` as discussed.

PR #265, opened 2017-01-26
WEB-1029: Offline mode functionality

fractaltheory:
Adds a couple of components and some logic to allow for 'offline mode functionality' if a network request for a page fails (for whatever reason).

 **JIRA**: [WEB-1029](https://mobify.atlassian.net/browse/WEB-1029)

## Changes
- Add `Offline` and `OfflineBanner` stub components based on design
- `onPageReceived` reducer adds fetched URL to a new `state.app`-level keys `fetchedUrls` and `fetchError`
- `App` container uses `fetchedUrls` and `fetchError` keys to determine whether to display offline components
- Worker uses custom fetch handler for modifying response body of `//online.mobify.net/online.json` to `{offline:true}` when request fails

## Todo
- [x] Refactor after Data Architecture 2.0 is merged
- [x] Fix problem of Offline component not dispatching router/fetch actions properly
- [x] Styling of offline components
- ~Custom fetch handler cache expiry~
- Implement whitelist / blacklist ?

## How to test-drive this PR
- Preview the app
- Emulate offline (DevTools > Network > Check 'Offline' checkbox)
- Navigate back to home, see that offline banner displays
- Navigate to any other page, see that offline splash displays


web/app/containers/app/reducer.js
```diff
@@ -1,18 +1,31 @@
...
+// we handle more types of errors, but will do for now
+export const isOffline = (state) => !!selectors.getFetchError()
+
 export default handleActions({
     [appActions.onRouteChanged]: (state, {payload: {currentURL}}) => {
         return state.set(CURRENT_URL, currentURL)
     },
+    [appActions.onPageReceived]: (state, {payload: {url}}) => {
+        const path = urlToPathKey(url)
+        return state.set(FETCHED_PATHS, state.get(FETCHED_PATHS).set(path, true))
```
DanielBrookRoberge: This could be `state.setIn([FETCHED_PATHS, path], true)`

web/app/containers/app/container.js
```diff
@@ -58,15 +66,18 @@ class App extends React.Component {
...
             >
                 <IconSprite sprite={sprite} />
                 <SkipLinks items={skipLinksItems} />
 
                 <div id="app-wrap" className="t-app__wrapper u-flexbox u-direction-column">
                     <div id="app-header" className="u-flex-none" role="banner">
-                        <CurrentHeader />
+                        <CurrentHeader>
+                            <OfflineBanner />
+                        </CurrentHeader>
```
jeffkamo: I'm not sure this format is very intuitive. What exactly does it mean if you pass children to `CurrentHeader`? Should it be passed in as a named prop to help clarify?

jeffkamo: I could see it being reasonable to give the header a prop for displaying notifications of some the [Feedback](http://docs.mobify.com/progressive-web/latest/components/#!/Feedback) or an [Alerts](http://docs.mobify.com/progressive-web/latest/components/#!/Alerts) component.

fractaltheory: Yeah I'm not convinced this is the way to do it either - the banner needs to be positioned in a particular way given it's fixed to the top. Maybe each header is aware of whether we're offline and includes the offline banner itself (rather than passed as a child, or as a prop).

web/worker/main.js
```diff
@@ -70,7 +73,113 @@ toolbox.router.get(/(?:cdn\.mobify\.com|localhost:8443)\/.*loader\.js$/, toolbox
...
+                    reject(['Bad response', request, response])
+                }
+            })
+            // This should happen if there's a network failure
+            .catch((error) => {
+                toolbox.options.debug && console.error(error)
+                reject([error, request])
+            })
+    })
+}
```
DanielBrookRoberge: This should be doable without the promise constructor; if you move the `.catch` handler above the `.then` handler then you can just return for resolve and throw for reject and return the result from `fetch().catch().then()` from the function.

web/worker/main.js
```diff
@@ -70,7 +73,113 @@ toolbox.router.get(/(?:cdn\.mobify\.com|localhost:8443)\/.*loader\.js$/, toolbox
...
-toolbox.router.get('/', toolbox.networkFirst, {cache: bundleCache})
+toolbox.router.get('/', customNetworkFirst, {
+    cache: bundleCache
+})
 
-toolbox.router.default = toolbox.networkFirst
+// Cache all other requests that aren't already matched by previous handlers
+toolbox.router.get(/.*/, customNetworkFirst, {
+    cache: toolbox.options.cache
+})
```
DanielBrookRoberge: Do we ever expire things from the cache doing it this way? If we just let it grow then eventually we just get our whole cache thrown out by the browser.

web/worker/main.js
```diff
@@ -70,7 +73,113 @@ toolbox.router.get(/(?:cdn\.mobify\.com|localhost:8443)\/.*loader\.js$/, toolbox
...
+ * @param {Response} response - the Response for the given Request
+ * @param {string} cacheName - the name of the cache to look for the Request in
+ * @param {Promise}
+ */
+const cacheFallback = (error, request, response, cacheName) => {
+    toolbox.options.debug && console.error(`Network or response error, fallback to cache [${request.url}]`)
+
+    /* eslint-disable max-nested-callbacks */
+    return caches.open(cacheName).then((cache) => {
+        return cache.match(request).then((cachedResponse) => {
```
DanielBrookRoberge: We can flatten this to 
```
return caches.open(cacheName)
    .then((cache) => cache.match(request))
    .then((cachedResponse) => { //stuff
```
Might help you not need the eslint exception.

web/app/containers/app/actions.test.js
```diff
@@ -14,37 +17,70 @@ afterAll(() => {
...
-//                 .toEqual(onPageReceived('$', '$response', pageType, url, 'test', '/'))
-//         })
-// })
+    return thunk(fakeDispatch, getState)
+        .then(() => {
+            expect(fakeGet).toBeCalledWith('x-mobify-progressive')
+            expect(fakeDispatch.mock.calls[0][0])
+            .toEqual(setPageFetchError('Failed to fetch, cached response provided'))
+        })
+})
```
fractaltheory: Agreed: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/717971f931a262897c67c81408a11109198e5621

web/app/containers/app/container.js
```diff
@@ -94,20 +116,26 @@ class App extends React.Component {
...
-    notifications: selectorToJS(selectors.getNotifications)
+    notifications: selectorToJS(selectors.getNotifications),
+    fetchError: selectors.getFetchError,
+    hasFetchedCurrentPath: selectors.hasFetchedCurrentPath
 })
 
 const mapDispatchToProps = {
-    removeNotification: appActions.removeNotification
+    removeNotification: appActions.removeNotification,
+    fetchPage: (url, pageComponent, routeName) => appActions.fetchPage(url, pageComponent, routeName)
```
fractaltheory: It sure does ... not sure what I was trying to do there: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/9b86a121048dcedd7949a90affe0ae14ec8ca285

web/app/containers/app/reducer.js
```diff
@@ -1,18 +1,31 @@
...
-    notifications: []
+    [CURRENT_URL]: window.location.href,
+    notifications: [],
+    fetchError: null,
+    [FETCHED_PATHS]: {}
 })
 
+// This will need to become more complicated when
+// we handle more types of errors, but will do for now
+export const isOffline = (state) => !!selectors.getFetchError()
```
fractaltheory: We don't actually use this anywhere anymore ... https://github.com/mobify/progressive-web-scaffold/pull/265/commits/a5feacf9e971feea363d38d489d51f438b8d9a79

web/app/containers/offline/offline.test.js
```diff
@@ -0,0 +1,34 @@
...
+import OfflineModal from './partials/offline-modal'
+
+// Avoid errors with accessing state that does not exist (within `OfflineModal`)
+jest.mock('reselect')
+
+const store = {
+    subscribe: () => {},
+    dispatch: () => {},
+    getState: () => ({})
+}
```
fractaltheory: Nice suggestion: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/77d8da20ff979b0c80051681093add05d4d4c238

web/worker/main.js
fractaltheory: I like it, great suggestion: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/366d61b74bc229711fda21f9d91a4d721f651f18

web/worker/main.js
fractaltheory: Yep, that worked, thanks: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/366d61b74bc229711fda21f9d91a4d721f651f18#diff-d1e334c00f26a87194902937d1b06b27R121

web/worker/main.js
fractaltheory: I definitely did not include any type of expiry here and that's a problem. Sucks to include even more custom code here.

fractaltheory:
Great feedback @DanielBrookRoberge - thank you.

web/app/containers/app/actions.test.js
DanielBrookRoberge: The test for the page fetch error action should be in a `then` handler, not a `catch` handler. These assertions aren't even being run.

fractaltheory: Oh boy.. oops: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/0ae6c1935e916938caa0ce35320f3754df14a206

web/app/containers/app/actions.test.js
```diff
@@ -17,35 +19,101 @@ afterAll(() => {
 // jest.mock('progressive-web-sdk/dist/jquery-response')
 // import {jqueryResponse} from 'progressive-web-sdk/dist/jquery-response'
 
+// const getState = () => ({ui: {app: Immutable.fromJS({[CURRENT_URL]: '/'})}})
```
jeremywiebe: Lots of commented out code here. Is it needed?

web/app/containers/app/actions.test.js
```diff
@@ -17,35 +19,101 @@ afterAll(() => {
...
+
+            expect(fakeDispatch).toHaveBeenCalledTimes(1)
+            expect(fakeDispatch.mock.calls[0][0]).toEqual(setPageFetchError('Network failure, using worker cache'))
+        })
+})
+
+test('checkIfOffline clears offline modal and page fetch errors when it receives untouched JSON from network', () => {
+    const mockResponse = {
+        obj: {},
+        json() {
```
jeremywiebe: This is inconsistent with line `70` above. Wouldn't this function share the same problem though?

https://github.com/mobify/progressive-web-scaffold/pull/265/files#diff-e3020d85657851283feb88c16b5d0e27R70

web/app/containers/app/constants.js
```diff
@@ -3,3 +3,7 @@
  */
 export const CURRENT_URL = 'currentURL'
 export const PLACEHOLDER = 'placeholder'
+export const FETCHED_PATHS = 'fetchedPaths'
+
+// This is the asset url we use to check if we're offline when making requests
+export const OFFLINE_ASSET_URL = '//online.mobify.net/online.json'
```
jeremywiebe: Nitpick, but do we expect this to ever be fetched over `http`?

web/app/containers/app/reducer.js
```diff
@@ -1,20 +1,27 @@
...
 import * as appActions from './actions'
-import {CURRENT_URL} from './constants'
+import {CURRENT_URL, FETCHED_PATHS} from './constants'
 
 const initialState = fromJS({
-    [CURRENT_URL]: false,
-    notifications: []
+    [CURRENT_URL]: window.location.href,
+    notifications: [],
+    fetchError: null,
```
jeremywiebe: Asking because I just don't know... do we need to put keys into `initialState` if they're `null` or `undefined`?

web/app/containers/offline/container.js
```diff
@@ -0,0 +1,36 @@
+import React, {PropTypes} from 'react'
```
jeremywiebe: We agreed in our Mobile Dev Sync that files containing React components should have a `.jsx` extension. Could you change this file to be `container.jsx`?

web/app/containers/app/actions.test.js
fractaltheory: This is actually just me reverting this test back to the way it is before I messed with it. I could either revert it completely (I had made changes to it, and then commented it out in this branch) or I could just remove the test entirely.

web/app/containers/offline/container.js
fractaltheory: Gotcha: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/7ff3625fae8f4481c1b9be1affaea1cdc74e6f11

web/app/containers/app/actions.test.js
fractaltheory: Good catch! I missed changing this based on DBR's feedback, woops: https://github.com/mobify/progressive-web-scaffold/pull/265/commits/c7b468a76361152d142a074105f0be758010b6c4

web/app/containers/app/actions.test.js
jeremywiebe: I think I vote for delete commented code. :)  But I'm cool if you just revert it too.  Changes to commented out code in a PR always look a bit strange to me.

web/app/containers/app/constants.js
fractaltheory: No, we don't - at least not in a PWA with service worker context. I thought that omitting the protocol where possible was a good practice, but I might be wrong?

web/app/containers/app/reducer.js
fractaltheory: I can either put it here, or in the selector as a `defaultValue`. Now that you mention it, I think we're likely trending toward the latter as a practice across the app, so maybe that's the way to go.

PR #266, opened 2017-01-26
Convert PLP data to Categories

DanielBrookRoberge:
This PR primarily moves the PLP data from the PLP container into its own area in the store, called `categories`. The new area contains more normalized data, with the original data model being recreated through selectors.

## Changes
- Move PLP data to Categories
- Remove unneeded fetch in progress flag
- Clean up notification reducer
- Remove duplicated item count in header

## How to test-drive this PR
- (necessary config changes)
- (necessary corresponding PRs)


app/containers/app/reducer.js
```diff
@@ -1,40 +1,27 @@
...
-import {FETCH_IN_PROGRESS, CURRENT_URL} from './constants'
+import {CURRENT_URL} from './constants'
 
-const initialState = Map({
-    [FETCH_IN_PROGRESS]: false,
+const initialState = fromJS({
     [CURRENT_URL]: false,
-        // If we use a regular array, React doesn't seem to catch all the updates
-    notifications: List()
+    notifications: []
```
fractaltheory: So the bug this comment referenced is no longer an issue?

app/store/categories/actions.js
```diff
@@ -0,0 +1,12 @@
+import {createAction} from '../../utils/utils'
+import plpParser from './parsers/plp'
+
+export const receiveCategory = createAction('Receive Category Data')
+
+export const process = ({payload}) => {
+    const {$, $response, url} = payload
+
+    return receiveCategory({
+        [new URL(url).pathname]: plpParser($, $response)
```
fractaltheory: I'm wondering about cases where we have different markup based on the same `pathname` but with different query parameters - these would all overwrite the same key in the store. Do you think that could pose problems?

I can think of a case where you'd visit `/foo.html?filter=bar` and see results of whatever was stored first in `/foo.html` until the network request completes and we update the `/foo.html` key with the new parsed content from `/foo.html?filter=bar`.

app/containers/app/reducer.js
```diff
@@ -16,20 +16,12 @@ export default handleActions({
...
-            const existingNotification = notifications.find((notification) => {
-                return notification.id === payload.id
-            })
-
-            if (existingNotification) {
-                return notifications
-            } else {
-                return notifications.push(payload)
-            }
+            return notifications.every(({id}) => id !== payload.id) ? notifications.push(payload) : notifications
```
fractaltheory: Very nice.

app/containers/app/reducer.js
```diff
@@ -16,20 +16,12 @@ export default handleActions({
...
         })
     },
     [appActions.removeNotification]: (state, {payload}) => {
         return state.update('notifications', (notifications) => {
-            return notifications.filterNot((notification) => notification.id === payload)
+            return notifications.filterNot(({id}) => id === payload)
```
fractaltheory: I think this is no longer a valid function, since `notifications` is back to a regular Array, right?

app/containers/app/reducer.js
DanielBrookRoberge: It is, I've just converted it to use `Immutable.fromJS` so it is automatically converted (less repetitious and more consistent)

app/store/categories/actions.js
DanielBrookRoberge: Yeah, this should really be the full URL with the protocol and domain stripped off. Thanks for the catch.

app/containers/app/reducer.js
fractaltheory: Ah, never mind - per your other comment about `fromJS`, `notifications` is converted to a `List` so we have access to this method 👍

fractaltheory:
I'm finding that PLP is not pulling previously stored data from the store when I re-visit one. I think it likely has to do with the storage keys different between full URLs and pathnames.

fractaltheory:
I think we just need to modify some filename paths inside `app/containers/catalog/products/parser.test.js` and `app/store/categories/parsers/plp.test.js`, then it's a 👍

PR #271, opened 2017-01-30
Schematize links

DanielBrookRoberge:
This PR introduces an Immutable Record type for a text link, as parsed by the `parseTextLink` function. The Record definition is temporarily placed in parser-utils.js in the absence of a general schema.

## Changes
- Introduce the Record definition
- Use it where necessary

## How to test-drive this PR
- Run the tests
- See that the footer and PLP continue to work correctly.

### DIFF
--- a/web/app/utils/parser-utils.js
+++ b/web/app/utils/parser-utils.js
@@ -1,9 +1,17 @@
+import Immutable from 'immutable'
+
+export const TextLink = Immutable.Record({
+    href: '',
+    text: '',
+    title: ''
+})
+
 export const parseTextLink = ($link) => {
-    return {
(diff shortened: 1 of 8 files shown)


web/app/utils/parser-utils.js
```diff
@@ -1,9 +1,17 @@
+import Immutable from 'immutable'
+
+export const TextLink = Immutable.Record({
```
mikenikles: I'm ok with that name for now. Once we have a central place for all `Record`s, it'll be more meaningful based on where we import `TextLink` from. Right now when I see code that just references `TextLink()` I wonder what it means. No action required, just a note.

PR #274, opened 2017-01-31
Astro: remove title and tabs in user's account tab

sshin-mobify:
Removes tabs and title in the user's registration/sign-in page.  This is a supplement PR to #278 

 **JIRA**: https://mobify.atlassian.net/browse/HYB-982
 **Linked PRs**: https://github.com/mobify/progressive-web-scaffold/pull/278

## Changes
- on login container, extracted login section and register section so that it can be retrieved separately depending on running in astro and route name.

## How to test-drive this PR
- on `tarBarConfig` set the account `rootUrl` to https://www.merlinspotions.com/customer/account/create/ to test register page
- on `tarBarConfig` set the account `rootUrl` to https://www.merlinspotions.com/customer/account/login/ to test sign-in page

PR #275, opened 2017-01-31
Astro: merlins add segmented view (tabs) into the account tab

sshin-mobify:
Creates native tab on account tab that switches between registration and login page.
This is supplement to #274

 **JIRA**: https://mobify.atlassian.net/browse/HYB-982
 **Linked PRs**: https://github.com/mobify/progressive-web-scaffold/pull/274, #278 

## Changes
- on account tab instead of normal navigationView, it leads to a new view with tabs and webview. 

## How to test-drive this PR
- start the app, go on the account tab. Switch between the tabs on the top. It should load the correct page


native/app/config/accountConfig.js
```diff
@@ -0,0 +1,13 @@
+const register = {
+    key: 'register',
+    text: 'Register',
+    url: 'https://www.merlinspotions.com/customer/account/create/'
```
jeremywiebe: Let's make these just relative URLs (I need to go back and fix the tabConfig.js so it uses relative URLs also).  I'd like to be able to hook in Preview so that we get the `Site URL` that was previewed and use that to drive all URLs that we load.  

So something like this:
```javascript
    url: '/customer/account/create/'
```

native/app/controllers/accountSegmentationController.js
```diff
@@ -0,0 +1,50 @@
+import WebViewPlugin from 'progressive-app-sdk/plugins/WebViewPlugin'
+import AnchoredLayoutPlugin from 'progressive-app-sdk/plugins/anchoredLayoutPlugin'
+import SegmentedPlugin from 'progressive-app-sdk/plugins/segmentedPlugin'
+
+import accountConfig from '../config/accountConfig'
+
+const AccountSegmentationController = function(layout, webView, segmentedView) {
+    this.layout = layout
```
jeremywiebe: One of these should be a `viewPlugin` to conform to the convention: controllers that export a view that can be placed into a layout export a single, named property called `viewPlugin` :) 

Eg. https://github.com/mobify/progressive-web-scaffold/blob/astro-merlins-base/native/app/controllers/tabController.js#L16

native/app/controllers/tabController.js
```diff
@@ -93,6 +99,11 @@ TabController.prototype.reload = async function() {
 }
 
 TabController.prototype.activate = function() {
+    // Don't know how to handle this yet so on account, don't do anything for now....
```
jeremywiebe: Can you ask Rose what the app should do on the Account tab if you tap it when it's already the selected tab and then fill me in?  I think a reasonable action is to pop to the root of the currently selected segment.

native/app/config/accountConfig.js
```diff
@@ -0,0 +1,13 @@
+const register = {
```
jeremywiebe: What do you think about the `tabBarConfig.js` having the concept of a tab containing multiple segments, each with their own relative URL?

Then the `tabController.js` could just check if it's tab config has a `segments` entry... if so, it creates a segmented plugin (like your accountSegmentationController, but just generic).  Then our scaffold could support any tab having a segmented control at the top and we wouldn't have to have `if (tab == 'Account') ...` anywhere in the tab controller.

native/app/controllers/tabController.js
```diff
@@ -6,9 +6,11 @@ import NavigationPlugin from 'progressive-app-sdk/plugins/navigationPlugin'
 import CounterBadgeController from 'progressive-app-sdk/controllers/counterBadgeController'
 
 import CartModalController from './cartModalController'
+import AccountSegmentationController from './accountSegmentationController'
 
 import baseConfig from '../config/baseConfig'
 import cartConfig from '../config/cartConfig'
+// import tabBarConfig from '../config/tabBarConfig'
```
jeremywiebe: Let's just delete this instead of commenting it out. :)

web/app/containers/login/container.js
```diff
@@ -10,13 +10,19 @@ import SkeletonText from 'progressive-web-sdk/dist/components/skeleton-text'
...
+import {isRunningInAstro} from '../../utils/astro-integration'
+
 import * as actions from './actions'
 
 class Login extends React.Component {
 
     // a few constants to make refactoring easier in future
-    static get SIGN_IN_SECTION() { return 'signin' }
-    static get REGISTER_SECTION() { return 'register' }
+    static get SIGN_IN_SECTION() {
```
jeremywiebe: Were these changes made because of the linter?  Typically we shouldn't re-format code unless: 
a) the linter calls it out
b) it doesn't conform to the [Mobify coding style](https://github.com/mobify/mobify-code-style)

web/app/containers/login/container.js
```diff
@@ -10,13 +10,19 @@ import SkeletonText from 'progressive-web-sdk/dist/components/skeleton-text'
 import {Tabs, TabsPanel} from 'progressive-web-sdk/dist/components/tabs'
 import {isRunningInAstro} from '../../utils/astro-integration'
 
+import {isRunningInAstro} from '../../utils/astro-integration'
```
jeremywiebe: This is imported on line 11 already.

native/app/controllers/tabController.js
```diff
@@ -39,8 +40,13 @@ TabController.init = async function(tabItem) {
     const counterBadgePlugin = await counterBadgeController.generatePlugin()
 
     await layout.addTopView(headerBar)
-    await layout.setContentView(navigationView)
-    await navigationView.setHeaderBar(headerBar)
+    if (tabItem.id === 'account') {                                                 // The account page does not require navigationView, instead using webView
```
jeremywiebe: We should try to avoid having knowledge about specific tabs bleeding into the `TabController`.  That becomes a slippery slope.  :)   Let's discuss this design when you come back to work on these comments.

native/app/controllers/TabHeaderController.js
```diff
@@ -0,0 +1,50 @@
...
+
+    headerBar.on(`click:${cartConfig.cartIcon.id}`, async () => {
+        const cartModalController = await CartModalController.init()
+        cartModalController.show()
+    })
+
+    return tabHeaderController
+}
+
+TabHeaderController.prototype.back = function() {
```
jeremywiebe: Do you think this will need to be called from outside this class?  If not, I think we could just trigger the 'back' event from the header bar `click:back` handler.

native/app/controllers/accountSegmentationController.js
```diff
@@ -0,0 +1,84 @@
+import WebViewPlugin from 'progressive-app-sdk/plugins/WebViewPlugin'
```
jeremywiebe: I know I'm being picky... but could we name this the `accountTabController.js`?  I think it makes it clearer that it's a special tab controller especially since it has to "conform" to the informal interface that a tab's controller has with the tabBarController.

native/app/controllers/tabBarController.js
```diff
@@ -24,9 +25,15 @@ TabBarController.init = async function() {
     const tabControllers = {}
 
     const tabControllerPromises = tabBarConfig.items.map((item) => {
-        return TabController.init(item).then((controller) => {
-            tabControllers[item.id] = controller
-        })
+        if (item.type === 'custom') {
+            return AccountSegmentationController.init().then((controller) => {
```
jeremywiebe: Thinking into the future... I think we should maybe also check the `item.id === 'account'` just so we have a pattern for other tabs in the future that may be `custom`.
```javascript
if (item.type === 'custom') {
    switch (item.id) {
        case 'account':
            return AccountTabController.init()....
            break;
        case ....... :
    }
}
```

native/app/controllers/tabController.js
```diff
@@ -1,71 +1,49 @@
...
+        headerController,
     ] = await Promise.all([
         AnchoredLayoutPlugin.init(),
-        HeaderBarPlugin.init(),
         NavigationPlugin.init(),
-        CounterBadgeController.init(cartConfig.cartIcon.imageUrl, 'headerId', {})
+        TabHeaderController.init()
     ])
 
-    const counterBadgePlugin = await counterBadgeController.generatePlugin()
```
jeremywiebe: Nice cleanup here!  :)

web/app/containers/login/container.js
jeremywiebe: @sshin-mobify Did you see this question? ☝️

sshin-mobify: My editor made lint changes by accident I thought I reversed most changes. I'll fix this right away

jeremywiebe:
👍

PR #276, opened 2017-01-31
Higher order template component

DanielBrookRoberge:
Add a Template higher-order component and use it to replace the logic in app-provider.jsx. This allows the Redux store update to occur after the outgoing template has unmounted, preventing render functions from seeing the store in an inconsistent state. 

## Changes
- Move logic to the higher-order component.
- Ensure that the page view analytics call is made on every page view.

## How to test-drive this PR
- Run the tests
- See that all the routes still work.

### DIFF
--- a/web/app/app-provider.jsx
+++ b/web/app/app-provider.jsx
@@ -4,10 +4,7 @@ import {Provider} from 'react-redux'
 
 // Containers
 import App from './containers/app/container'
-import {Cart, Home, Login, PLP, PDP} from './containers/templates'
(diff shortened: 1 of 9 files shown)


kschmidtdev:
I would suggest pushing a bundle and previewing it just to make sure all the routing works in production (there were production-only issues due to uglification in the past with this area of the codebase)

mikenikles:
Hey @kschmidtdev, thanks for the heads up. I added a note to #239 to make sure we run regression tests on a production bundle before we merge the new architecture.

PR #278, opened 2017-02-01
Astro: merlins add sign-in and registration

sshin-mobify:
Adds sign-in and registration component to merlins potions demo app. This branch is a combination of 3 different PRs: #274, #275 and #279.

 **JIRA**: https://mobify.atlassian.net/browse/HYB-982
 **Linked PRs**: #274, #275, #279

 Once all 3 PRs are approved, this is the branch that should be merged into the base

## Changes
- removes web title and web tabs in the account #274 
- adds native tabs for registration and sign-in #275 
- from onboarding clicking registration or sign-in leads the user to the account tab with the correct section #279

## How to test-drive this PR
- navigate to `native/onboarding` and run `npm run build` to load the new onboarding page
- launch the app with onboarding on
- a. click register : leads the user to the account tab with registration page loaded
- b. click sign in : leads the user to the account tab with sign in page loaded
- switch between the register and sign in tab, the app should load the correct page

PR #279, opened 2017-02-01
Astro: merlins sign in registration onboarding

sshin-mobify:
From onboarding pressing, sign in or registration would lead the user to the account tab, with the correct page loaded. This is a supplement to #275

 **JIRA**: https://mobify.atlassian.net/browse/HYB-982
 **Linked PRs**: #275 #278 

## Changes
- on onboarding.js created 2 methods to handle on register click and on signin click
- created RPC method to handle those clicks
- created App event that triggers the segment tabs to load the correct page

## How to test-drive this PR
- navigate to `native/onboarding` and run `npm run build` to load the new onboarding page
- launch the app with onboarding on
- a. click register : leads the user to the account tab with registration page loaded
- b. click sign in : leads the user to the account tab with sign in page loaded


native/app/app.js
```diff
@@ -33,6 +36,16 @@ window.run = async function() {
             Application.setStatusBarLightText()
         })
 
+        Astro.registerRpcMethod(rpcMethodNames.registerShow, [], () => {
```
jeremywiebe: What do you think about moving these into the `TabBarController` itself?  It's the one that is doing the work of the RPC method call and I don't see anything needed in app.js for this. 

If you do that, then the TabBarController could actually tell the Account tabController to show the SignIn or Register tab directly instead of using AppEvents which I think is a bit cleaner as you can easily tell exactly what the RPC method all does.

native/app/app.js
```diff
@@ -33,6 +36,16 @@ window.run = async function() {
...
+            tabBarController.selectTab('account')
+            AppEvents.trigger(AccountSegEvents.registerSelected)
```
jeremywiebe: This should be triggered from within the accountSegmentationController itself.  Objects shouldn't trigger events they don't own (and I'd think app.js doesn't own the `AccountSegEvents`).  This is to make it more obvious where to look for when something gets triggered (I wouldn't expect this event to be raised from somewhere else).

native/app/global/app-rpc-method-names.js
```diff
@@ -1,7 +1,9 @@
 
 const rpcMethodNames = {
     onboardingShow: 'onboardingShow',
-    onboardingHide: 'onboardingHide'
+    onboardingHide: 'onboardingHide',
+    signInShow: 'signInShow',
```
jeremywiebe: Let's change the value of these method names to be "[kebab cased](http://wiki.c2.com/?KebabCase)"   So
```javascript
signInShow: 'sign-in:show'
```

native/onboarding/src/components/Onboarding.js
```diff
@@ -23,37 +23,47 @@ class Onboarding extends Component {
...
-                <Button className="c--secondary u-flex" text={data.actionButtonTitle} />
+                <Button className="c--secondary u-flex" text={data.actionButtonTitle} onClick={onSignInTapped} />
             </div>
         </div>
     </CarouselItem>
     return item
 }
 
+function onRegisterTapped(event) {
+    Astro.jsRpcMethod(rpcMethodNames.registerShow, [])();
```
jeremywiebe: What do you think about having `registerShow` automatically hide the onboarding modal?  I think it is good if we make the web as dumb about the native side as possible. Having the web side know that it needs to dismiss the onboarding couples them more tightly.

sshin-mobify: This means that `onboardingModalController` is handling `Astro.registerRpcMethod` correct? 
So it goes from` onboardingModalController` -> `tabBarController` -> `accountSegmentationController` ?

sshin-mobify: It's not ton of change, I think that was my initial implementation lol. Doing this makes the code a lot simpler and less hopping from one controller to another, I like it a lot more.

PR #281, opened 2017-02-02
Replace @extend with @include

jeffkamo:
This is so we can add !important to the heading utility classes, which
wasn’t possible with extend.

## Changes
- Replace the heading utilities' `@extend` declarations with mixin `@include`s instead.
- Heading utilities now have `!important`, as all utilities are intended to have.

## How to test-drive this PR
- Run `npm run dev`
- Visit any page with an html heading and confirm it's base styles have NO `!important`
- Find a head utility class and confirm it has `!important`
- Confirm that all heading levels (1-6) work as intended.

### DIFF
--- a/web/app/styles/utilities/_heading.scss
+++ b/web/app/styles/utilities/_heading.scss
@@ -6,25 +6,25 @@
 }
 
 .u-h1 {
-    @extend %h1; // 28px/32px
+    @include heading-level-1(true); // 28px/32px
 }
 
 .u-h2 {
-    @extend %h2; // 20px/24px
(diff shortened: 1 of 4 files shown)


web/app/styles/utilities/_heading.scss
```diff
@@ -6,25 +6,25 @@
 }
```
mlworks: While we are at it, let's make `u-heading-family` to use `!important` as well

jeffkamo: Good point! Done

PR #284, opened 2017-02-02
Improve the clarity of the selectors with better utilities

DanielBrookRoberge:
This PR adds some more utilities to `selector-utils.js`, and improves the functionality of some existing utilities. It also adds tests and documentation to those utilities, and refactors all of the selectors to use the new utilities.

This PR also contains an exploration of how selectors might be specified more cleanly in the login selector file (i.e. `web/app/containers/login/selectors.js`). Feedback is welcome, and I am willing to move it to a later time for further development/discussion.

 **Linked PRs**: #239 

## How to test-drive this PR
- run the tests
- Run `npm run dev`
- check that things still work.

### DIFF
--- a/web/app/utils/selector-utils.js
+++ b/web/app/utils/selector-utils.js
@@ -6,6 +6,21 @@ export const createImmutableComparingSelector = createSelectorCreator(
     Immutable.is
 )
 
+/**
+ * Creates a selector that handles the conversion from Immutable
+ * objects to plain Javascript objects.
+ *
+ * If the result of the input selector is identical to its previous
+ * result result using Immutable.is, this will return the same JS
+ * object as the previous call.
+ *
+ * This simplifies downstream update checks substantially, since if
+ * the relevant store contents are unchanged the new object will be
+ * === to the old object. (i.e. this is where the magic happens)
+ *
+ * @param {function} selector - A selector returning an Immutable object
+ * @returns {function}
+ */
 export const selectorToJS = (selector) => createImmutableComparingSelector(
     selector,
     (raw) => { return raw ? raw.toJS() : null }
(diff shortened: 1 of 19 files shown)


web/app/utils/selector-utils.js
```diff
@@ -6,19 +6,80 @@ export const createImmutableComparingSelector = createSelectorCreator(
     Immutable.is
 )
 
+/**
+ * Creates a selector that handles the conversion from Immutable
+ * objects to plain Javascript objects.
+ *
+ * If the result of the input selector is identical to its previous
+ * result result using Immutable.is, this will return the same JS
```
mikenikles: `result result`. Let's 💣  one `result` :)

web/app/utils/selector-utils.js
```diff
@@ -6,19 +6,80 @@ export const createImmutableComparingSelector = createSelectorCreator(
...
+ * object as the previous call.
+ *
+ * This simplifies downstream update checks substantially, since if
```
mikenikles: By `downstream update checks` you refer to the `mapStateToProps` optimization we get from `react-redux`, right? Do you think it's worth pointing that out explicitly here?

web/app/utils/selector-utils.js
```diff
@@ -6,19 +6,80 @@ export const createImmutableComparingSelector = createSelectorCreator(
...
+/**
+ * Creates a selector that gets a value from a selected Immutable object.
+ *
+ * @param {function} selector - A selector returning an Immutable object
+ * @param {string|number|function} key - The key to be looked up on
+ *   the Immutable object. If a function is passed it is treated as a
+ *   selector returning the desired key.
+ * @param {*} [defaultValue] - An optional value to be returned if the
+ *   key does not exist in the Immutable object.
+  * @returns {function}
```
mikenikles: There seems to be an extra space here.

web/app/utils/selector-utils.js
```diff
@@ -6,19 +6,80 @@ export const createImmutableComparingSelector = createSelectorCreator(
...
+ */
+export const createGetSelector = (selector, key, defaultValue) => {
+    if (typeof key === 'function') {
+        return createSelector(
+            selector,
+            key,
+            (obj, keyValue) => obj.get(keyValue, defaultValue)
+        )
+    } else {
```
mikenikles: This `else` statement doesn't seem necessary. Since the `if` statement returns from the function, we could remove the `else` block and simply return the value currently wrapped in the `else` block.

The same applies to the `createHasSelector` function.

jnurse: [Non-blocking] Since your if statement is returning, this second case doesn't need to be wrapped in an else.

web/app/utils/selector-utils.js
```diff
@@ -6,19 +6,80 @@ export const createImmutableComparingSelector = createSelectorCreator(
...
+ * @returns {function}
+ */
+export const createHasSelector = (selector, key) => {
+    if (typeof key === 'function') {
+        return createSelector(
+            selector,
+            key,
+            (obj, keyValue) => obj.has(keyValue)
+        )
+    } else {
```
jnurse: [Non-blocking] This section doesn't need to be wrapped in an else.

web/app/utils/selector-utils.js
DanielBrookRoberge: In our case we use a `createStructuredSelector` to make `mapStateToProps`, triggering that optimization. But this is also applicable for something like `shallowCompare` or `React.PureComponent`, since it makes shallow comparisons express deep equality.

I'll rephrase this, is I guess what I'm getting at.

mikenikles:
👍

PR #286, opened 2017-02-03
Parser refactor and bugfixes

DanielBrookRoberge:
This PR refactors the Merlin's parsers and fixes some minor issues.

 **JIRA**: N/A

## Changes
- Add a utility for getting the text from a selected element.
- Fix all parsers that use the `$` object from jQuery to receive it as a parameter (the global is not guaranteed to either exist or be the right version)
- Refactor the code style to be more consistent.

## How to test-drive this PR
- (necessary config changes)
- Run `npm run dev`


web/app/containers/home/parsers/home.js
```diff
@@ -1,21 +1,24 @@
 import {parseImage} from '../../../utils/parser-utils'
 
+const BANNER_SELECTOR = 'strong.logo, .home-t-shirts, .home-erin'
```
mikenikles: I like that pattern. When selectors aren't self-explanatory anymore, creating a `const` with a meaningful name makes it easy to read and maintain.

web/app/containers/login/parsers/common.js
```diff
@@ -0,0 +1,30 @@
...
+import {getTextFrom} from '../../../utils/parser-utils'
+
+export const parseFields = ($, $fields) => {
+    return $fields.map((_, field) => {
+        const $field = $(field)
+        const $tooltip = $field.find('.tooltip.wrapper')
+        return {
+            label: getTextFrom($field, 'label'),
+            name: $field.find('input').attr('name'),
+            type: $field.find('input').attr('type'),
```
jnurse: You could store $field.find('input') in a variable instead of calling find for it twice.

jnurse: These functions also might make sense in the parser-utils file (or some other common parser file). I imagine these will be useful in the checkout as well.

DanielBrookRoberge: I think I'd like to hold off moving these to a more common utility file until I know more about how valid the selectors are for checkout. But it's definitely something to keep in mind.

PR #288, opened 2017-02-04
Disallow global jquery and jest

mikenikles:
Removes the ESLint `jquery` and `jest` environments globally and instead relies on explicit configuration in files where global variables of these environments are necessary.

The main benefit of this PR is to enforce the use of our internal `$` variable returned by the SDK's `jqueryResponse` function, rather than the global `$` bound to `window` in browsers.

## Changes
- Removed the ESLint env settings.
- Updated all test files accordingly.

## How to test-drive this PR
- Run `npm run test:js`.
  - Verify that it succeeds.
- Open any file and try to use `$`.
  - Verify that ESLint reports an error.
- Run `npm run dev`


web/app/containers/login/parsers/common.js
```diff
@@ -29,3 +29,5 @@ export const parseHiddenInputs = ($, $inputs) => {
         }
     })
 }
+
+export const isFormResponseInvalid = ($, $response, formSelector) => $response.find(formSelector).length
```
jnurse: It looks like `$` isn't needed here.

web/app/containers/login/actions.js
```diff
@@ -94,12 +96,10 @@ const validateRegisterForm = (formValues) => {
...
-        .then((response) => {
-            return response.text()
-        })
-        .then((responseText) => {
-            const $html = $(responseText)
-            if ($html.find(formSelector).length) {
+        .then(jqueryResponse)
+        .then((res) => {
+            const [$, $response] = res // eslint-disable-line no-unused-vars
+            if (isFormResponseInvalid($response)) {
```
jnurse: Shouldn't the formSelector be passed to this function too?

mikenikles:
@jnurse Let's never talk about the code I originally pushed... :). I addressed your comments in 37bbc9b

jnurse:
Looks good! 👍

PR #289, opened 2017-02-06
WEB-1146: Fixes Stepper on PDP on iOS 9.3

twangtina:
Fixes Stepper on PDP on iOS 9.3

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1146

## Changes
- Fixes the issue where the stepper component is squished on iOS 9.3

## How to test-drive this PR
- use iOS 9.3
- Run `npm run dev`
- Preview to: https://www.merlinspotions.com/
- Navigate to a PDP
- Look at the Quantity stepper
- The minus and plus icons should not be squished

### DIFF
--- a/web/app/containers/pdp/_pdp.scss
+++ b/web/app/containers/pdp/_pdp.scss
@@ -16,6 +16,14 @@
 }
 
 
+// In Stock Indicator
+// ---
+
+.t-pdp__indicator {
+    flex: 1 1 auto;
+}
+
+
 // Add to Cart Button
 // ---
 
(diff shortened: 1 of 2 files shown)


web/app/containers/pdp/_pdp.scss
```diff
@@ -16,6 +16,14 @@
 }
 
 
+// In Stock Indicator
+// ---
+
+.t-pdp__indicator {
+    flex: 1 1 auto;
```
jeffkamo: I would recommend instead of creating more CSS, there is an existing utility class that does the same thing: [`u-flex`](https://github.com/mobify/progressive-web-scaffold/blob/develop/web/app/styles/utilities/_flexbox.scss#L17) (which is based on [this Spline mixin](https://github.com/mobify/spline/blob/develop/dist/utilities/flexbox/_flexbox.scss#L137))

PR #295, opened 2017-02-07
WEB-1112: Checkout payment page refactor

mlworks:
**JIRA**: https://mobify.atlassian.net/browse/WEB-1112

## Changes
- Refactor payment sections into their own partials
- Wire up selectors to pull data from cart
- Add toggle functionality:
  - show/hide add new card form
  - show/hide new address form
  - show add company / apt field
  - show/hide fixed position place order container

## How to test-drive this PR
- Run `npm run dev`
- Add an item to your cart
- Navigate to: [Payment step](https://www.merlinspotions.com/checkout/payment/)


web/app/containers/checkout-payment/partials/product-list.jsx
```diff
@@ -0,0 +1,266 @@
...
+        this.handleScroll = throttle(this.handleScroll.bind(this), 200)
+    }
+
+    componentDidMount() {
+        window.addEventListener('scroll', this.handleScroll.bind(this))
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener('scroll', this.handleScroll.bind(this))
+    }
```
DanielBrookRoberge: If you bind the method here it won't actually remove it correctly, since it'll actually be a different function value than in `componentDidMount`. Since we already bind the function up in the controller we shouldn't need to do it again.

web/app/containers/checkout-payment/partials/product-list.jsx
```diff
@@ -0,0 +1,266 @@
...
+
+        return (
+            <div className="t-checkout-payment__product-list">
+                <div className="t-checkout-payment__title u-padding-top-lg u-padding-bottom-lg">
+                    <h2 className="u-h4">Order Summary</h2>
+                </div>
+
+                <List className="u-bg-color-neutral-00 u-border-light-top u-border-light-bottom">
+                    {cartItems.map((item, idx) => {
+                        const productImage = (
```
DanielBrookRoberge: It might be worth splitting out this function body into something like a PaymentCartItem component; it would help us notice whether we're using similar code in different templates.

web/app/containers/checkout-payment/reducer.js
```diff
@@ -1,25 +1,19 @@
...
-        return state.merge({isCompanyOrAptShown: true})
-    },
-    [checkoutPaymentActions.toggleFixedPlaceOrder]: (state, payload) => {
-        return state.merge({isFixedPlaceOrderShown: payload})
-    },
+const checkoutPayment = handleActions({
+    [checkoutPaymentActions.toggleFixedPlaceOrder]: mergePayload,
+    [checkoutPaymentActions.toggleCardInputRadio]: mergePayload,
+    [checkoutPaymentActions.toggleCompanyAptField]: mergePayload,
+    [checkoutPaymentActions.toggleNewAddressFields]: mergePayload
```
DanielBrookRoberge: There's another utility in `reducer-utils` called `mergePayloadForActions` that would make this even simpler; these four lines could become
```
...mergePayloadForActions(toggleFixedPlaceOrder, toggleCardInputRadio, etc.)
```

web/app/containers/checkout-payment/partials/product-list.jsx
mlworks: Ah right! bad copy pasta job

web/app/containers/checkout-payment/partials/product-list.jsx
```diff
@@ -0,0 +1,215 @@
...
+
+        return (
+            <div className="t-checkout-payment__product-list">
+                <div className="t-checkout-payment__title u-padding-top-lg u-padding-bottom-lg">
+                    <h2 className="u-h4">Order Summary</h2>
+                </div>
+
+                <List className="u-bg-color-neutral-00 u-border-light-top u-border-light-bottom">
+                    {cartItems.map((item, idx) =>
+                        <PaymentProductItem item={item} key={idx} />
```
DanielBrookRoberge: We could pass in the item details as separate props with `{...item}`; I'd recommend it so that the PaymentProductItem can be more explicit about what it expects.

PR #292, opened 2017-02-07
Update checkout shipping for Architecture 2.0 

jnurse:
This PR updates the checkout shipping template to the new architecture, and adds missing functionality to the page

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1015
 **Linked PRs**: n/a

## To Do
- [x] Refactor Checkout shipping sections into partials (#291)
- [x] Connect checkout shipping partials to the redux store (#293)
- [x] Parse content from the desktop site for the UI (#298)
- [x] Address design feedback (https://mobify.atlassian.net/browse/WEB-1099) (#308)
- [x] Add missing functionality to checkout shipping page (ex. recognize known email and show sign in form) (#319)
- [x] Add selectors to access the store for all applicable checkout shipping props (#298)
- [x] Persist shipping form data through the API (#305)
- [x] Refetch shipping methods on county or region change (#305)
- [x] Add initial values for shipping form (#314)

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- add an item to the cart
- navigate to: https://www.merlinspotions.com/checkout/
- the page should work as expected and defined in the invision mocks (see: https://mobify.atlassian.net/browse/WEB-1015

PR #298, opened 2017-02-08
Selectorize checkout shipping

jnurse:
This PR converts access to the store for existing props to selectors, and parses more data from the desktop site for the checkout shipping page. 

 **Linked PRs**: #292 

## To Do
- [x] Selectorize access to the store for existing checkout shipping props
- [x] Get shipping method data for the page from magento's API
- [x] Removed some hardcoded content from the shipping address form
- [x] Parse available countries and regions from the desktop page 
- [ ] ~~Get initial values for form if user is logged in (no saved address)?~~

## How to test-drive this PR
- Run `npm run dev`
- Add an item to the cart
- confirm the page works the same as it did before except now countries and regions are populated correctly


web/app/containers/checkout-shipping/parsers/checkout-shipping.js
```diff
@@ -0,0 +1,26 @@
+import {extractMegentoShippingStepData} from '../../../utils/magento-utils'
```
DanielBrookRoberge: This looks like it's misspelled `Magento`

DanielBrookRoberge:
I would recommend for the test failure you'd been having that you unwrap the Redux wrapped component as described at https://github.com/mobify/progressive-web-sdk/pull/486/files#diff-dad0396b9f0ee57f735dd31f06e660b9R1086

PR #302, opened 2017-02-08
Use relative URLs for cart calls

sap1ens:
We can't really use absolute URLs if we want to support multiple Merlin's environments (staging, demo, etc.).

It seems like we don't need to change anything else, it just works.

JIRA: https://mobify.atlassian.net/browse/WEB-1155

## How to test-drive this PR
- Run `npm run dev`
- Look at AJAX calls for cart endpoints


fractaltheory:
The remove and update requests aren't hooked up to anything, but the `getCart` fetch works. Shouldn't be any issue removing the host per the documentation for `whatwg-fetch` 👍

sap1ens:
Thanks 👍

PR #303, opened 2017-02-08
WEB-1112: Design feedback fixes

mlworks:
Address WEB-1099 design feedback for Payment Step (WEB-1112)

 **JIRA**:
https://mobify.atlassian.net/browse/WEB-1112
https://mobify.atlassian.net/browse/WEB-1099

## Changes
- Adjust various spacing as per design feedback
  - Tweaked Accordion's padding and units to match latest checkout mocks (44px height instead of 48px)
  - Tweaked Ledger's padding
- Use `CardInput` component from the SDK to get auto number formatting
- Removed `u-bg-color-neutral-10` from checkout payment container into `app.scss` modifier styles
   - Previously, the background color wasn't stretching to cover the available space leaving undesired whitespace between the main content and the footer. (Can easily reproduce in tablet landscape view with only 1 item in cart)

## How to test-drive this PR
- Run `npm run dev`
- Add an item to your cart
- Navigate to: [Payment step](https://www.merlinspotions.com/checkout/payment/)
- confirm that the items checked off in the WEB-1099 JIRA ticket have been fixed correctly


web/app/styles/themes/pw-components/_accordion.scss
```diff
@@ -13,23 +13,30 @@
     }
 }
 
-.pw-accordion__header {
-    padding: $unit*1.5;
+.pw-accordion__icon {
+    min-width: initial;
```
jeffkamo: Interesting, is this fixing a bug or some unexpected behaviour?

web/app/styles/themes/pw-components/_card-input.scss
```diff
@@ -0,0 +1,25 @@
...
+// ===
+
+// Root
+// ---
+//
+// 1. `pw-card-input` is used inside `c-field__input` which has `display: flex`
+//    using `flex: 1 1 auto` so card input will stretch and fill the available space
+
+.pw-card-input {
+    flex: 1 1 auto; // 1
```
jeffkamo: Is this something we should fix in the SDK base styles later?

web/app/styles/themes/pw-components/_ledger.scss
```diff
@@ -8,19 +8,29 @@
...
+
+.pw-ledger__item {
+    padding: 0; // 1
 
-.pw-ledger__item .pw-ledger__action {
-    color: $neutral-50;
-    font-size: 12px;
+    .pw-ledger__action {
+        color: $neutral-50;
+        font-size: 12px;
```
jeffkamo: Could we switch to `$small-font-size`?

web/app/containers/app/_app.scss
```diff
@@ -20,3 +20,14 @@
 .t-app__wrapper {
     min-height: 100vh; // 1
 }
+
+
+
+// Modifiers
+// ---
+
+.t-app.t-app--checkout-payment {
```
jeffkamo: If this is the way we go, you can probably just use `t-app--checkout-payment`

web/app/containers/app/_app.scss
```diff
@@ -20,3 +20,14 @@
...
+    [id="app-main"] {
```
mlworks: Yea I'm not so sure about that, I think it's gonna look a bit crazy with all the class names.  I took a look at the styles again, turns out I don't need to scope the styles to `[id="app-main"]` I'm able to style the background color in `t-app--checkout-payment` to achieve the same effect!

web/app/containers/app/_app.scss
mlworks: Would that break our CSS code style with modifier classes?  I guess technically this modifier class has `t-app` so it's still self-documenting.

web/app/containers/checkout-payment/_checkout-payment.scss
```diff
@@ -83,9 +83,20 @@
 
 // Desktop Styles
 // ---
+//
+// 1. Add some white space between the product list and footer
 
 @include susy-breakpoint($desktop-breakpoint, $desktop-layout) {
+    .t-checkout-payment__card {
+        border-right: 1px solid $neutral-15;
+        border-left: 1px solid $neutral-15;
```
mlworks: Ah my bad, I was looking at tablet landscape which happened to hit desktop breakpoint. Will make that fix.

web/app/styles/themes/pw-components/_card-input.scss
mlworks: Yea I think we should bring this into the base styles since the `CardInput` component is most likely used inside the `Field` component

web/app/containers/app/_app.scss
jeffkamo: I think self-documentation wise it's fine.

That said, the reason I went with `t-app--checkout-payment` was because `t--checkout-payment` looked too similar to `t-checkout-payment`. Maybe that's not a good enough reason to break convention, and maybe we shouldn't use that classname at all. Open to discussing an alternative solution

web/app/styles/themes/pw-components/_accordion.scss
mlworks: Wow so after taking a closer look. I've missed an update when I was tweaking the SDK accordion base styles. The `pw-accordion__icon` should be 40px, not 44px.  Maybe for now I'll update this style to `min-width: $unit*5` so its more descriptive

PR #305, opened 2017-02-09
Submit checkout shipping

jnurse:
This PR adds form submission to the checkout shipping page, and refetches shipping methods when the user has entered their country, region and postal code. 

 **Linked PRs**: #292 

## Changes
- Updated checkout shipping input names to match desktop
- Refetch shipping methods on postal code blur 
- Submit shipping form data to the back end (parse data into correct format and make API call)
- Switch to payment page on successful response

## How to test-drive this PR
- Run `npm run dev`
- add an item to your cart
- confirm the page works as expected


web/app/containers/checkout-shipping/actions.js
```diff
@@ -21,20 +25,78 @@ export const onShippingEmailRecognized = () => {
...
+        // Default values to use if none have been selected
+        const addressData = {country_id: 'US', region_id: '0', postcode: null}
+        if (formValues) {
+            addressData.country_id = formValues.country_id
+            addressData.region_id = formValues.region_id
+            addressData.postcode = formValues.postcode
+        }
         const getEstimateURL = `https://www.merlinspotions.com/rest/default/V1/${isLoggedIn ? 'carts/mine' : `guest-carts/${entityID}`}/estimate-shipping-methods`
-        makeJsonEncodedRequest(getEstimateURL, addressData, {method: 'POST'})
+        makeJsonEncodedRequest(getEstimateURL, {address: addressData}, {method: 'POST'})
```
DanielBrookRoberge: If you call the variable `address` this could just be `{address}`. Not 100% sure that's worth it.

web/app/containers/checkout-shipping/partials/checkout-shipping-form.jsx
```diff
@@ -53,10 +61,21 @@ const validate = (values) => {
     return errors
 }
 
+const mapStateToProps = () => {
+    return {}
+}
```
DanielBrookRoberge: This could become `const mapStateToProps = () => ({})`.

web/app/app-provider.jsx
```diff
@@ -10,7 +11,7 @@ import CheckoutFooter from './containers/checkout-footer/container'
 
 const AppProvider = ({store}) => (
     <Provider store={store}>
-        <Router>
+        <Router history={browserHistory}>
```
DanielBrookRoberge: This should already be done for us in the SDK. Does it not work if this isn't here?

web/app/containers/checkout-payment/container.jsx
```diff
@@ -82,18 +82,18 @@ class CheckoutPayment extends React.Component {
...
     showCompanyAndApt: PropTypes.func,
     toggleFixedPlaceOrder: PropTypes.func,
 }
 
 const mapStateToProps = (state) => {
     return {
-        checkoutPayment: state.checkoutPayment,
-        miniCart: state.miniCart
+        checkoutPayment: state.ui.checkoutPayment,
+        cart: state.cart.toJS()
```
jnurse: I wasn't planning on making any changes to the checkout payment page, I think Mike L is handling that, I just needed to resolve the JS errors so I could get the switch to the next page working.

web/app/app-provider.jsx
jnurse: It didn't work for me without this here. I remember we had to do this in the Lancome build because otherwise when you include browserHistory in your action it's pointing to a different browserHistory than the one being used by the router.

web/app/containers/checkout-payment/container.jsx
DanielBrookRoberge: :+1:

PR #310, opened 2017-02-10
Remove Unnecessary Styleguidist Stylesheet

jeffkamo:
We pull in the PWA's stylesheet into Styleguidist, but the way we were doing it before was kind of unnecessary, and had the possibility of being itself out of sync with the SDK component's base styles. Since the SDK component styles are pulled into the App styles anyway, this stylesheet is simply not necessary.

## Changes
- Delete the styleguide.scss from the styleguide directory
- Set `Renderer.jsx` to import the app's stylesheet directly

## How to test-drive this PR
- Run `npm run styleguide`
- Preview [the local style guide](http://localhost:4000/)
- Verify that the style guide more or less works, and pulls in the App stylesheet as expected.


mlworks:
Talked to Jeff, ideally in a base scaffold we'll have all options available and the developer will remove imports to the components they don't need. In the case of Merlin's we are going for optimization so I'm fine with this change.

PR #311, opened 2017-02-10
Astro merlins android integration

kerrmarin:
Add android implementation of MP

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1003

## Changes
- Add android project

## How to test-drive this PR
- Run `npm run dev`
- Run android app
- click about


web/service-worker-loader.js
```diff
@@ -1,8 +1,18 @@
-// THIS IS A STRING VALUE NOT A BOOLEAN, BE CAREFUL
-const isPreview = /preview=(true|false)/.exec(self.location.search)[1]
+import {getBuildOrigin} from 'progressive-web-sdk/dist/asset-utils'
```
jeremywiebe: We should check with @DanielBrookRoberge to make sure we can do `import`s here. This file is deployed on the client's domain so just want to make sure this is legal code here.

web/app/app-provider.jsx
```diff
@@ -72,6 +72,7 @@ const AppProvider = ({store}) => {
         if (urlHasChanged) {
             if (isCoordinatingWithNativeApp) {
                 pwaNavigate(nextURL).then(() => {
+                    console.log('Finishing up here')
```
jeremywiebe: 🔫

native/onboarding/onboarding-build/static/css/main.bcff1842.css
```diff
@@ -0,0 +1,2 @@
...gn:center;-webkit-align-items:center;-ms-flex-align:center;align-items:center}.t-checkout-shipping{min-height:100vh;-webkit-transition:min-height 2s ease-out;transition:min-height 2s ease-out}.t-checkout-shipping.t--loaded{min-height:0}.t-checkout-shipping__progress{max-width:540px;margin-right:auto;margin-left:auto}.t-checkout-shipping__form .c-field__label,.t-checkout-shipping__form .pw-field__label{font-size:15px}@media (min-width:0) and (max-width:419px){.t-checkout-shipping__email-title{padding-top:16px}.t-checkout-shipping__title{padding-left:12px}}@media (min-width:420px) and (max-width:1023px){.t-checkout-shipping__email-title{padding-top:16px}}@media (min-width:1024px){.t-checkout-shipping__email-title{padding-top:56px}}.t-checkout-payment__progress{max-width:540px;margin-right:auto;margin-left:auto}.t-checkout-payment__form .c-field__label,.t-checkout-payment__form .pw-field__label{font-size:15px}@media (min-width:0) and (max-width:419px){.t-checkout-payment__title{padding-left:12px}}#root,body,html{height:100%}*{-webkit-touch-callout:none;-webkit-user-select:none;-moz-user-select:none;-ms-user-select:none;user-select:none}body{background-color:#f7f7f7;margin:0;padding:0}
```
jeremywiebe: Is this a generated file?  If so, should it be in git?

native/app/controllers/tabBarController.js
```diff
@@ -85,4 +86,16 @@ TabBarController.prototype.showSignIn = function() {
...
+TabBarController.prototype.backActiveItem = function() {
+    if (this.canGoBack()) {
+        const activeTab = this.getActiveController()
+        activeTab.back()
+    }
+}
+
+TabBarController.prototype.canGoBack = function() {
+    const activeTab = this.getActiveController()
+    return activeTab.canGoBack()
```
jeremywiebe: I'm not sure, but should this be an `async` function as we're calling an `async` function in `activeTab.canGoBack()` and returning the value it returns (which I assume is a Promise).

native/app/controllers/tabBarController.js
```diff
@@ -21,6 +21,7 @@ const TabBarController = function(tabBar, layout, tabControllers) {
 
 TabBarController.init = async function() {
     const tabBar = await TabBarPlugin.init()
+    await tabBar.setColor('#4E439B')
```
jeremywiebe: Can we source this colour value from a config file instead of hard-coding it here?  I think this is the same as `baseConfig.primaryColor`.

native/app/app.js
```diff
@@ -57,12 +66,12 @@ window.run = async function() {
             previewController.presentPreviewAlert()
         })
 
-        const previewEnabled = await previewController.isPreviewEnabled()
-        if (previewEnabled) {
-            runAppPreview()
-        } else {
-            runApp()
-        }
+        // const previewEnabled = await previewController.isPreviewEnabled()
```
jeremywiebe: Does this need cleanup yet?  If not, can we delete this commented-out code?

native/app/app.js
```diff
@@ -40,6 +40,15 @@ window.run = async function() {
             tabBarController.showSignIn()
         })
 
+        // Android hardware back
+        const setupHardwareBackButton = async (alternativeBackFunction) => {
```
jeremywiebe: Why does this need to be `async`?

native/android/settings.gradle
```diff
@@ -0,0 +1,3 @@
+include ':scaffold', ':astro', ':astro:pushclient'
+project(':astro').projectDir = new File(settingsDir, '../node_modules/mobify-progressive-app-sdk/android')
+project(':astro:pushclient').projectDir = new File(settingsDir, '../node_modules/mobify-progressive-app-sdk/node_modules/mobify-push-android-client/pushclient')
```
jeremywiebe: I don't think this path is right if we're fully on Node 6+.  `npm v3+` no longer nests `node_modules` folders.

native/android/scaffold/src/main/res/drawable-xxhdpi/launch_screen.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
```
jeremywiebe: This doesn't look like it matches the Merlin's designs. 

</img>

native/onboarding/onboarding-build/asset-manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "main.css": "static/css/main.bcff1842.css",
```
jeremywiebe: Are all of these `bcff1842...` files generated?

native/onboarding/onboarding-build/index.html
```diff
@@ -0,0 +1 @@
+<!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><meta name="viewport" content="width=device-width,initial-scale=1,user-scalable=no,minimum-scale=1,maximum-scale=1"><title>Onboarding</title><link href="static/css/main.bcff1842.css" rel="stylesheet"></head><body><div id="root"></div><script type="text/javascript" src="static/js/main.877ec958.js"></script></body></html>
```
jeremywiebe: I ran the app on Android and I couldn't swipe through the onboarding carousel. Did it work for you?

web/service-worker-loader.js
DanielBrookRoberge: Yeah, we 100% can't import here. Also, if I remember correctly, `getBuildOrigin` uses the DOM to get information, which is not available in the worker context.

DanielBrookRoberge: Yes, it looks for the script tag in the DOM so it can't work. Any information you want to be available in the worker needs to be passed in the worker URL or in IndexedDB.

web/service-worker-loader.js
```diff
@@ -1,8 +1,18 @@
...
+import {isRunningInAstro} from './app/utils/astro-integration'
 
-if (isPreview === 'true') {
-    self.importScripts('https://localhost:8443/worker.js')
-} else {
-    self.importScripts('https://cdn.mobify.com/sites/progressive-web-scaffold/production/worker.js')
+try {
+    if (isRunningInAstro) {
+        throw 'Service worker not supported in Astro'
+    }
```
DanielBrookRoberge: With the changes in the loader above we won't need this.

native/android/scaffold/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+android {
+    compileSdkVersion 23
+    buildToolsVersion "23.0.2"
+
+    defaultConfig {
+        applicationId "com.mobify.astro.scaffold"
+        minSdkVersion 19
+        targetSdkVersion 19
+        versionCode 1
+        versionName "1.0"
```
kerrmarin: Can do -- but doesn't this only work because in BT we are not using `app.js` to set the IP address? I think this makes sense to add to the Astro SDK and make it available to the previewPlugin/Application, not to each individual project?

jeremywiebe: Hmm, you might be right. Wonder if this could be pushed into the SDK somehow. I guess right now if we always rely on Chrome port forwarding we _can_ just use `localhost`.  Let's not do this for now and see how things go.

native/onboarding/onboarding-build/static/css/main.bcff1842.css
kerrmarin: No it shouldn't -- I've removed it. Good catch

native/app/controllers/tabBarController.js
kerrmarin: Done

native/app/app.js
kerrmarin: It doesn't -- I've removed it

native/app/app.js
kerrmarin: Ah, so we can't use the isPreviewEnabled flag on Android because of the caching by the service worker. So for now I've placed a config value in `baseConfig` to trigger on/off the preview.

native/onboarding/onboarding-build/asset-manifest.json
kerrmarin: Removed

native/app/app.js
jeremywiebe: Is that still an issue given we are disabling service worker in Astro?

jeremywiebe:
@kerrmarin This looks good to me!!! :D   👍

PR #301, opened 2017-02-08
Astro: Merlins checkout app event showing sign in and showing shop

sshin-mobify:
Astro: Merlins checkout app event showing sign in and showing shop
from #300 create app event that shows the correct tab.

 **JIRA**: https://mobify.atlassian.net/browse/HYB-985
 **Linked PRs**: #300

## Changes
- create app event called `showSignIn` and `showShop` that causes tabBarController to select account or shop tab

## How to test-drive this PR
- open up checkout cart modal
- click signIn, the modal should close and open signIn tab
- click continue Shopping, the modal should close and open the shop tab

PR #313, opened 2017-02-10
Astro: Merlins Remove web header in checkout

sshin-mobify:
This removes default web header in the checkout flow. This is to match the looks of 
http://merlins-app.surge.sh/#artboard1

 **JIRA**:  https://mobify.atlassian.net/browse/HYB-985
 **Linked PRs**: #301

## Changes
- if runninh in Astro return null else return header
## How to test-drive this PR
- add an item to cart
- open the checkout modal
- press proceed to checkout
- on the top you shouldn't see a web header, you should see the checkout flow icons at the top

### DIFF
--- a/web/app/containers/checkout-header/container.js
+++ b/web/app/containers/checkout-header/container.js
@@ -3,34 +3,39 @@ import React from 'react'
 import Button from 'progressive-web-sdk/dist/components/button'
 import {HeaderBar, HeaderBarTitle} from 'progressive-web-sdk/dist/components/header-bar'
 import {Icon} from 'progressive-web-sdk/dist/components/icon'
+import {isRunningInAstro} from '../../utils/astro-integration'
 
 const CheckoutHeader = function(props) {
     const canSignIn = true
-    return (
-        <header className="t-checkout-header">
-            <HeaderBar className="t-checkout-header__bar">
-                <HeaderBarTitle className="u-flex-none u-padding-start u-text-align-start">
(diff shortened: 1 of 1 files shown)


web/app/containers/checkout-header/container.js
```diff
@@ -3,34 +3,39 @@ import React from 'react'
...
-        <header className="t-checkout-header">
-            <HeaderBar className="t-checkout-header__bar">
-                <HeaderBarTitle className="u-flex-none u-padding-start u-text-align-start">
-                    <h2 className="t-checkout-header__title u-heading-family u-text-uppercase">
-                        <span className="u-text-lighter">MERLIN'S</span> CHECKOUT
-                    </h2>
-                </HeaderBarTitle>
+    if (isRunningInAstro) {
+        return null
+    } else {
```
jeremywiebe: Given that you return in the `if` block, you probably don't need an `else` here.  A common pattern is to "fail fast" and then have the "happy path" continue to the end of the function. Having an `else` here just means you nest the "happy path" one level down... when really you don't need to.

See point #4 in this article: https://blog.codinghorror.com/flattening-arrow-code/  :)

sshin-mobify: I know what you mean, before I just had `if (!isRunningInAstro) return Header` and it was not loading. I fixed it by returning null then it worked. So I kept it like that, however not that I have `{!isRunningInAstro && Header}` it seems to work. Weird

PR #314, opened 2017-02-10
Add initial values to checkout shipping form

jnurse:
This PR adds initial values to the checkout shipping form and hides the email section for users that are logged in. 

 **Linked PRs**: #292 

## Changes
- Hide email section for logged in users
- Add initial values for fields based on JSON object found on the page
- Update the shipping address submission data's format

## How to test-drive this PR
- Run `npm run dev`
- login as a user that has no address saved yet
- add an item to your cart


DanielBrookRoberge:
Loading this branch on the home page I get the error
```
Cannot read property 'constructor' of undefined
```
which sounds like it's happening in the `isLoggedIn` check at app-parser.js:8

web/app/store/checkout/actions.js
```diff
@@ -1,9 +1,18 @@
...
 
-export const processShippingLocations = ({payload: {$response}}) =>
-    receiveShippingLocations(parseLocations($response))
+export const processCheckoutData = ({payload: {$response}}) => {
+    return (dispatch) => {
+        const magentoFieldData = extractMagentoShippingStepData($response).getIn(['children', 'shipping-address-fieldset', 'children'])
+        const initialValues = parseShippingInitialValues(magentoFieldData)
+
+        dispatch(receiveShippingLocations(parseLocations(magentoFieldData)))
+        dispatch(receiveShippingMethodInitialValues({shipping: {initialValues}}))
```
DanielBrookRoberge: Alternatively, and this may be relevant anyways, if you change the action definition to 
```
createAction('Receive Shipping Method Initial Values', 'shipping')
```
then this line can become `dispatch(receiveShippingMethodInitialValues({initialValues})`.

I must admit to this part being badly documented, but if we use additional parameters to `createAction` they will be the keys that each parameter to the action creator is merged under. The header container does this.

DanielBrookRoberge:
Approved! :+1:

DanielBrookRoberge:
actually, I'm getting an error if I'm logged out; `shippingMethods[0]` is undefined for me in the `fetchShippingMethods` function

jnurse:
@DanielBrookRoberge I'm not seeing an error when I'm logged out, but that sounds like the same issue I ran into a fixed yesterday. What's the URL it's trying to fetch the shipping methods with and the data it's using?

PR #322, opened 2017-02-11
Shipping known email

jnurse:
This PR adds some functionality to the email section on the checkout shipping page. Specifically on blur it will check if the email address is known, and show the sign in form if it is. It also adds submission handling for the sign in form.

 **Linked PRs**: #292 

## Changes
- Added onBlur handler to email address field
- Added call to API to check if email address is known
- Added submit handling for the sign in form

## How to test-drive this PR
- Run `npm run dev`
- add an item to the cart (as a logged out user)
- enter an email address that has an account associated with it 
- sign in on the form that appears

### DIFF
--- a/web/app/containers/checkout-shipping/actions.js
+++ b/web/app/containers/checkout-shipping/actions.js
@@ -1,8 +1,9 @@
 import {browserHistory} from 'react-router'
 import {createAction} from '../../utils/utils'
+import CheckoutShipping from './container'
 import checkoutShippingParser from './parsers/checkout-shipping'
 import shippingMethodParser from './parsers/shipping-method'
-import {addNotification} from '../app/actions'
+import {addNotification, fetchPage, removeNotification} from '../app/actions'
 import {getCustomerEntityID} from './selectors'
 import {getIsLoggedIn} from '../app/selectors'
 import {getShippingFormValues} from '../../store/form/selectors'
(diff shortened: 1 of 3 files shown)


web/app/containers/checkout-shipping/actions.js
```diff
@@ -28,6 +29,49 @@ export const onShippingEmailRecognized = () => {
...
+}
+
+export const submitSignIn = () => {
+    return (dispatch, getState) => {
+        const {
+            username,
+            password
+        } = getShippingFormValues(getState())
+
+        // This data has to be sent via AJAX, it doesn't work with makeJsonEncodedRequest
```
DanielBrookRoberge: Could this be done using a raw `fetch()` call?

web/app/containers/checkout-shipping/actions.js
```diff
@@ -28,6 +29,49 @@ export const onShippingEmailRecognized = () => {
...
+export const checkCustomerEmail = () => {
+    return (dispatch, getState) => {
+        const formValues = getShippingFormValues(getState())
+
+        makeJsonEncodedRequest('https://www.merlinspotions.com/rest/default/V1/customers/isEmailAvailable', {customerEmail: formValues.username}, {method: 'POST'})
+            .then((response) => response.text())
+            .then((responseText) => {
+                if (/false/.test(responseText)) {
+                    dispatch(onShippingEmailRecognized())
+                }
```
DanielBrookRoberge: A small thing and unlikely to happen much, but if you edit the email again to something without an account the password field does not go away.

web/app/containers/checkout-shipping/actions.js
jnurse: We spoke about this offline, but in case anyone looks back here and wonders what we found. Looks like a raw fetch or makeRequest here still don't work. I've compared the desktop request (using AJAX) to ours using makeJsonEncodedRequest and they're almost identical, but makeJsonEncodedRequest still doesn't work. So we'll leave this as is for now.

DanielBrookRoberge:
:+1: let's merge this!

PR #323, opened 2017-02-14
Analytic Middleware

wizardlyhel:
Provide an analytic middleware structure that allows easy multi-platform analytic implementation.

**Linked PR:** https://github.com/mobify/progressive-web-sdk/pull/499

## Changes
- Add redux-analytic npm package
- Introduce analytic middleware
- Fire pageview analytic using middleware
- Introduce a new action creator that accepts meta payload

## How to test-drive this PR
Setting up a local progressive-web-sdk dist build. Use https://github.com/mobify/progressive-web-sdk/pull/499 to test this change
```
npm run dev:build
```
**Scaffold project**
Inside `package.json`, change the path to progressive-web-sdk to the local version
```
     "progressive-web-sdk": "file:../../progressive-web-sdk"
```

In the scaffold, run the following so that you have a local copy of progressive-web-sdk
```
rm -r node_modules/progressive-web-sdk/
npm install
npm run dev
```
Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)

**Note:** Circle CI test will fail due to SDK branch required


web/app/containers/app/actions.js
```diff
@@ -34,7 +34,7 @@ export const removeAllNotifications = utils.createAction('Remove All Notificatio
  * @param {string} pageComponent - the component of the entered route
  * @param {string} currentURL - what's currently shown in the address bar
  */
-export const onRouteChanged = utils.createAction('On route changed', 'currentURL', 'pageComponent')
+export const onRouteChanged = utils.createActionWithMeta('On route changed', ['currentURL'], (...args) => (args[1]))
```
jeremywiebe: Is there a change there won't be 2 arguments ever?  Should that be something we protect against here?

web/app/main.jsx
```diff
@@ -16,11 +16,15 @@ import AppProvider from './app-provider'
 // added to the markup in `loader.js`
 import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
 
-import {initMobifyAnalytics} from 'progressive-web-sdk/dist/analytics'
+import {analyticsInitializer} from 'progressive-web-sdk/dist/analytics/analytic-manager'
+import * as ClientGA from './utils/analytics/client-ga'
```
jeremywiebe: Given that this is a sample, should we maybe just call this `client-analytics`?

web/app/utils/utils.js
```diff
@@ -17,6 +17,71 @@ export const createAction = (description, ...argNames) => {
...
+ *          }
+ *      }
+ * }))
+ *
+ * = Using convient meta payload creator for analytic =
+ * dispatch(actionVariable('abc', 'def', analyticMetaPayloadCreator('Pageview', {
+ *      name: 'PLP'
+ * })
+ *
+ * Note: The intention of this function is to replace createAction utilty function defined above
```
jeremywiebe: This looks like it could be a drop-in replacement...   is there any reason we shouldn't just make this method the new `createAction` method?

web/app/containers/app/actions.js
wizardlyhel: No no.
Payload arguments are now listed in an array. You can have as many as you like.
`'pageComponent'` is removed cause it was not used anywhere. Confirmed with @DanielBrookRoberge that it was no longer used after the refactor

web/app/utils/utils.js
wizardlyhel: For backwards compatibility, we must maintain `createAction`
`createActionWithMeta` have a different function signature and is to re-enable all features of `createReduxAction` that `redux-actions` originally comes with. In this case, it's the ability to pass in meta creator

web/app/main.jsx
wizardlyhel: ... yea

wizardlyhel:
@jeremywiebe Ready for another pass with changes from SDK PR

web/app/template.jsx
```diff
@@ -27,9 +29,7 @@ const template = (WrappedComponent) => {
                 Astro.trigger('pwa-navigate', {url})
             }
 
-            triggerMobifyPageView(route.routeName)
-
-            dispatch(onRouteChanged(url, WrappedComponent))
+            dispatch(onRouteChanged(url, analyticMetaPayloadCreator(analyticConstants.pageview, {name: route.routeName})))
```
DanielBrookRoberge: I still think this is more complicated than it needs to be; we shouldn't be building the analytics details where the action is dispatched, but where the action is defined. This way, we don't need to duplicate it wherever we dispatch the action and keep it in sync.

So if we do (above in actions.js)
```javascript
const onRouteChanged = createActionWithMeta('On route changed', ['currentURL'], (currentURL, name) => analyticMetaPayloadCreator(analyticConstants.pageview, {name}))
```
then this can be
```javascript
```

web/app/template.jsx
```diff
@@ -1,9 +1,11 @@
 import React, {PropTypes} from 'react'
 import {connect} from 'react-redux'
 import {onRouteChanged, fetchPage, removeAllNotifications} from './containers/app/actions'
-import {triggerMobifyPageView} from 'progressive-web-sdk/dist/analytics'
 import Astro from './vendor/astro-client'
 
+import {analyticMetaPayloadCreator} from './utils/utils'
```
DanielBrookRoberge: Since this is a regular function returning a value and not a higher-order function, I'd vote for something that's a verb, like `createAnalyticsMeta` (since it's not just creating the payload but the full meta object)

jeffkamo:
@wizardlyhel I think you got a +1 from Jer, can this PR be merged? (after resolving conflicts)

PR #325, opened 2017-02-14
Simpler method of offline detection

fractaltheory:
Instead of writing our own custom fetch handler (and subsequently, a cache handling mechanism), use an alternative, simpler method of detecting if we're offline.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1029
 **Linked PRs**: #265 

## Changes
- Revert changes to worker `main.js` file to state prior to #265 
- Add a new JSON file `static/js/offline-test.json` which is requested each time a page is requested
- Add worker handler for this request URL which will provide a success response of `{offline: true}` when the network fails
- Update logic to app to detect this (instead of a response header), while keeping original app logic the same (for whether to display offline banners/splash screen, etc.)

## TODO
- [x] Stand up a static asset on CDN which ignores query parameters to replace the `offline-test.json` file included as part of bundle
- [x] Modify request url to be the asset above, once deployed

## How to test-drive this PR
- Run `npm run dev`
- You should be able to just visit a few pages, switch to 'Offline' in devTools and then see the banner (on pages you've already cached, like the homepage) and the splash screen (on uncached pages)
- You should also be able to retry the connection from either the offline banner modal or the splash screen (once back online) and the app should recover
- If you're quite keen, you could push a bundle to the CDN, and using Charles, map `https://www.merlinspotions.com:443/service-worker-loader.js` to a local file which simply imports the `worker.js` file you just pushed to Cloud (which should verify that the CORS request we make doesn't pose an issue)

### DIFF
--- a/web/app/containers/app/actions.js
+++ b/web/app/containers/app/actions.js
@@ -59,7 +59,9 @@ export const checkIfOffline = () => {
     return (dispatch) => {
         // we need to cachebreak every request to ensure we don't get something
         // stale from the disk cache on the device
-        return utils.makeRequest(`${getBuildOrigin()}static/js/offline-test.json?${Date.now()}`)
+        return fetch(`//cdn.mobify.com/sites/progressive-web-scaffold/bundles/51/static/js/offline-test.json?${Date.now()}`, {
+            cache: 'no-store'
+        })
             .then((response) => response.json())
             .then((json) => {
                 if (json.offline) {
@@ -86,8 +88,6 @@ export const fetchPage = (url, pageComponent, routeName) => {
         return utils.makeRequest(url)
             .then(jqueryResponse)
             .then((res) => {
-                dispatch(checkIfOffline())
-
                 const [$, $response] = res
 
                 const currentURL = selectors.getCurrentUrl(getState())
@@ -109,6 +109,10 @@ export const fetchPage = (url, pageComponent, routeName) => {
                 }
                 dispatch(footerActions.process(receivedAction))
                 dispatch(navigationActions.process(receivedAction))
+
+                // Finally, let's check if we received a cached response from the
+                // worker, but are in fact 'offline'
+                dispatch(checkIfOffline())
             })
             .catch((error) => {
                 console.info(error.message)
--- a/web/app/containers/app/actions.js
+++ b/web/app/containers/app/actions.js
@@ -1,5 +1,4 @@
 import {jqueryResponse} from 'progressive-web-sdk/dist/jquery-response'
-import {getBuildOrigin} from 'progressive-web-sdk/dist/asset-utils'
 
 import * as utils from '../../utils/utils'
 import * as selectors from './selectors'
@@ -16,6 +15,7 @@ import * as navigationActions from '../navigation/actions'
 import * as productsActions from '../../store/products/actions'
 import * as categoriesActions from '../../store/categories/actions'
 
+import {OFFLINE_ASSET_URL} from './constants'
 import {closeModal} from '../../store/modals/actions'
 import {OFFLINE_MODAL} from '../offline/constants'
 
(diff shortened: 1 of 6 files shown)


web/app/containers/app/actions.test.js
```diff
@@ -17,68 +16,104 @@ afterAll(() => {
...
-    const fakeGet = jest.fn(() => 'offline')
-    const response = {
-        headers: {
-            get: fakeGet
+test('checkIfOffline dispatches setPageFetchError if it receives modified JSON from worker', () => {
+    const mockResponse = {
+        obj: {
+            offline: true
+        },
+        json: function() { // eslint-disable-line object-shorthand
```
DanielBrookRoberge: What eslint is expecting here is `json() {`, actually, not an arrow function.

web/app/containers/app/actions.test.js
```diff
@@ -17,68 +16,104 @@ afterAll(() => {
...
+
+            expect(fakeDispatch).toHaveBeenCalledTimes(1)
+            expect(fakeDispatch.mock.calls[0][0]).toEqual(setPageFetchError('Network failure, using worker cache'))
+        })
+})
+
+test('checkIfOffline clears offline modal and page fetch errors when it receives untouched JSON from network', () => {
+    const mockResponse = {
+        obj: {
+            offline: false
```
DanielBrookRoberge: Should this now be an empty object? (just in case)

web/worker/main.js
```diff
@@ -56,6 +53,35 @@ self.addEventListener('activate', (e) => {
...
+// App makes this asset request on each page fetch, expecting to see empty JSON
+// if network supplies successful response.
+// In the case of failure, we modify response to be `{offline: true}` which
+// indicates to app that we are offline.
+const checkIfOffline = (request) =>
+    fetch(new Request(request, {cache: 'no-store'}))
+    .catch(() => noCacheJSONResponse({offline: true}))
+
+// For enabling offline detection within the application
+// @TODO {MQ} - Modify this regular expression to match the final asset URL once
```
DanielBrookRoberge: Is this still TODO?

web/worker/main.js
```diff
@@ -56,6 +53,35 @@ self.addEventListener('activate', (e) => {
...
+    )
+}
+
+// App makes this asset request on each page fetch, expecting to see empty JSON
+// if network supplies successful response.
+// In the case of failure, we modify response to be `{offline: true}` which
+// indicates to app that we are offline.
+const checkIfOffline = (request) =>
+    fetch(new Request(request, {cache: 'no-store'}))
+    .catch(() => noCacheJSONResponse({offline: true}))
```
DanielBrookRoberge: that's so nice and simple. :+1:

fractaltheory:
@DanielBrookRoberge cleaned up based on your review, thanks for taking a look. Gonna go ahead and merge.

web/app/containers/app/actions.js
```diff
@@ -49,30 +51,46 @@ export const setPageFetchError = utils.createAction('Set page fetch error', 'fet
...
+ * Make a separate request that is intercepted by the worker. The worker will
+ * return a JSON object where `{offline: true}` if the request failed, which we
+ * can use to detect if we're offline.
+ */
+export const checkIfOffline = () => {
+    return (dispatch) => {
+        // we need to cachebreak every request to ensure we don't get something
+        // stale from the disk cache on the device - the CDN will ignore query
+        // parameters for this asset, however
+        return fetch(`${OFFLINE_ASSET_URL}?${Date.now()}`, {
```
johnboxall: I'm working on Lancome and verifying SW functionality, so I came looking here to see more about how it works... Dumb question: Why don't we just use `navigator.onLine`?

https://developers.google.com/web/updates/2011/06/navigator-onLine-in-Chrome-Dev-channel

fractaltheory: It's a good question actually - our thought here was that `onLine` is tied to the network adapter more than actual network conditions. So a simple network request will let us know if we can actually get some successful responses.

kschmidtdev: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine#Browser_compatibility - interesting that some Android WebViews don't support it properly so we ended up calling into native code to determine if an Android device is online: https://github.com/mobify/mobifyjs-burlington/blob/a8e0186ee48e0e60a232ba419524ba2e33af320a/native/app/app-plugins/networkStatePlugin.js

PR #326, opened 2017-02-14
Service worker loader comment and preview mode detection cleanup

Bielousov:
**JIRA**: N/A

## Changes
- Refactoring: preview mode detection code in service worker loader should return boolean
- Updated service worker loader intro comment

## How to test-drive this PR
- Run `npm run dev`
- See that service worker messages are shown as before in the console and that it is shown as registered in the Application tab of Chrome dev tools.


web/service-worker-loader.js
```diff
@@ -1,7 +1,8 @@
-// THIS IS A STRING VALUE NOT A BOOLEAN, BE CAREFUL
-const isPreview = /preview=(true|false)/.exec(self.location.search)[1]
+// MOBIFY PROGRESSIVE SERVICE WORKER LOADER
+// DO NOT MODIFY
```
jeremywiebe: When I see comments like this, I usually have two questions:
1. Why can't I modify it?
1. What if I feel I need to modify it?

Bielousov: Because this file is sent to client and the comment is for them.

jeremywiebe: Oh, I see now. I'm not 100% sure. 🤔 

Maybe just leave it the way it is.

DanielBrookRoberge: Maybe say `DO NOT MODIFY WITHOUT APPROVAL FROM MOBIFY`?

PR #329, opened 2017-02-15
Add formatting for phone input on checkout shipping page

jnurse:
This PR adds formatting to the phone number input on the checkout shipping page. 

 **Linked PRs**: n/a

## Changes
- Added normalizePhone util 
- Added normalize param to the phone number input

## How to test-drive this PR
- Run `npm run dev`
-  add an item to your cart
- fill out the form
- there should be formatting when you enter a phone number
- submit the form
- you see the payment page


web/app/utils/normalize-utils.js
```diff
@@ -0,0 +1,33 @@
+// Based on the example here: http://redux-form.com/6.0.1/examples/normalizing/
+export const normalizePhone = (value, previousValue) => {
```
jeremywiebe: This assumes a North American phone number. I'm ok with this as long as it's a conscious decision. :)

jnurse: Yeah the PWA doesn't really handle localization at all right now, so I figured it would be safe to assume a north american phone number. Happy to discuss it in more detail if anyone has any thoughts on this.

web/app/utils/normalize-utils.js
```diff
@@ -0,0 +1,33 @@
...
+    if (!value) {
+        return value
+    }
+
+    const onlyNums = value.replace(/\D/g, '')
+
+    const areaCode = onlyNums.slice(0, 3)
+    const exchange = onlyNums.slice(3, 6)
+    const lineNumber = onlyNums.slice(6, 10)
```
DanielBrookRoberge: It might be worth removing/ignoring a leading `1` here, since the area code is guaranteed to begin with a digit from 2-9.

web/app/utils/normalize-utils.js
jeremywiebe: @jnurse Cool.  If the scaffold isn't localized now, that can be something we think about when we do our first "international" app.  :)

web/app/utils/normalize-utils.js
jnurse: @DanielBrookRoberge do you mean stripping out the leading 1 when we replace all non-numbers on line 7?

DanielBrookRoberge: Yeah something like that.

PR #330, opened 2017-02-15
Cart stepper functionality

jnurse:
This PR fixes the stepper on the cart so it correctly updates the qty of an item.

 **Linked PRs**: n/a

## Changes
- Added onChange handler to stepper for cart items
- Updated updateItemQuantity action to not use FormData
- Moved getCart call into updateItemQuantity so we don't need to call both functions to update the qty

## How to test-drive this PR
- Run `npm run dev`
- Add an item to the cart
- Increase/decrease the qty for the item
- The cart should update correctly


web/app/store/cart/actions.js
```diff
@@ -21,6 +24,7 @@ export const getCart = () => (dispatch) => {
     const opts = {
         headers: baseHeaders
     }
+    dispatch(removeAllNotifications())
```
DanielBrookRoberge: Do we want to remove all notifications here, or just the cart quantity error one we set below?

web/app/containers/cart/partials/cart-product-list.jsx
```diff
@@ -99,14 +101,16 @@ const CartProductItem = ({product_name, product_image, idx, qty, product_price,
 
 CartProductItem.propTypes = {
     idx: PropTypes.number,
+    item_id: PropTypes.number,
```
DanielBrookRoberge: This appears to actually be a string; I'm getting a proptype failure.

web/app/containers/cart/partials/cart-product-list.jsx
```diff
@@ -41,7 +42,7 @@ const ProductSkeleton = () => (
 
 /* eslint-disable camelcase */
 
-const CartProductItem = ({product_name, product_image, idx, qty, product_price, onSaveLater}) => (
+const CartProductItem = ({product_name, product_image, item_id, idx, qty, product_price, onSaveLater, onQtyChange}) => (
```
bendvc: Why are we using underscore case? Although the style guide doesn't specifically say we have to use camelCase for params and object keys it does say that variables have to be camelCase
https://github.com/mobify/mobify-code-style/blob/master/javascript/README.md#use-camel-case-for-variable-names

web/app/containers/cart/partials/cart-product-list.jsx
```diff
@@ -60,6 +61,7 @@ const CartProductItem = ({product_name, product_image, idx, qty, product_price,
                     decrementIcon="minus"
                     initialValue={qty}
                     minimumValue={1}
+                    onChange={(newVal) => { onQtyChange(item_id, newVal) }}
```
bendvc: Doesn't look like onQtyChange is a required property, so potentially it could be null. Might want to add a default value like an empty function.

web/app/containers/cart/partials/cart-product-list.jsx
jnurse: @bendvc that's the format of the object we get from the server. We're not currently re-parsing it.

web/app/store/cart/actions.js
```diff
@@ -57,19 +61,28 @@ export const removeFromCart = (itemId) => {
...
+        const headers = {
+            ...baseHeaders,
+            'Content-Type': 'application/x-www-form-urlencoded',
+        }
 
-    const opts = {headers, body, method: 'POST'}
-    return fetch('/checkout/sidebar/updateItemQty/', opts)
-        .json()
+        const opts = {headers, body, method: 'POST'}
+        return makeRequest('/checkout/sidebar/updateItemQty/', opts)
```
bendvc: It would be nice to either start declaring these urls as const's higher up in the file, or possible in another file that we import, just to clean things up.

web/app/containers/cart/partials/cart-product-list.jsx
```diff
@@ -97,16 +99,22 @@ const CartProductItem = ({product_name, product_image, idx, qty, product_price,
     </ProductItem>
 )
 
+CartProductItem.defaultProps = {
+    onQtyChange: () => {}
```
DanielBrookRoberge: We'll get more accurate test coverage numbers (down the road) if you import `noop` from `progressive-web-sdk/dist/utils/utils` and use it here instead.

PR #331, opened 2017-02-15
Astro: Merlins fix account tab + onboarding from the merge + cart item added modal

sshin-mobify:
Fixes broken components from the dev merge into `astro-merlins-base`

## Changes
- login container import `isRunningInAstro` directly
- onboarding buttons trigger astro
- coverage is calculated and passed through

## How to test-drive this PR
- test the account page: switch between register tab and sign in tab
- from onboarding selecting register now and sign in should lead you to the account tab


web/app/containers/pdp/container.js
```diff
@@ -7,13 +7,20 @@ import PDPAddToCart from './partials/pdp-add-to-cart'
...
+    }
+
     return (
         <div className="t-pdp">
             <PDPHeading />
             <PDPCarousel />
             <PDPDescription />
             <PDPAddToCart />
-            <PDPItemAddedModal />
+            <PDPItemAddedModal coverage={coverage} />
```
jeremywiebe: The `PDPItemAddedModal` doesn't have a `coverage` property anymore.  What do you think about just moving this calculation into that modal component just before the `<Sheet ...` line?

https://github.com/mobify/progressive-web-scaffold/blob/develop/web/app/containers/pdp/partials/pdp-item-added-modal.jsx#L15

PR #333, opened 2017-02-15
Cart edit button

jnurse:
This PR fixes the cart item edit button so it correctly navigates to the PDP

 **Linked PRs**:n/a

## Changes
- Added href to cart item edit button
- Added PDP route for configure item URL

## How to test-drive this PR
- Run `npm run dev`
- Add an item to your cart
- Click "edit" for the item in you cart
- You should be taken to the PDP for the item


web/app/app-provider.jsx
```diff
@@ -23,6 +23,7 @@ const AppProvider = ({store}) => (
                 <Route component={PLP} path="supplies.html" routeName="productListPage" />
                 <Route component={PLP} path="new-arrivals.html" routeName="productListPage" />
                 <Route component={PLP} path="charms.html" routeName="productListPage" />
+                <Route component={PDP} path="checkout/cart/configure/id/*/product_id/*/" routeName="productDetailsPage" />
```
jeremywiebe: These are mostly design questions/comments, but if we re-use the `productDetailsPage`, the breadcrumbs don't reflect that we're in the cart/checkout flow. Also, there's no indication that we're editing the cart quantity since the "action" button reads "Add to Cart" and not "Update Cart".

jnurse: @jeremywiebe that's a good point, we'll probably also need to update what happens when you click the action button as well. I'll take another look at this later.

jnurse:
@jeremywiebe I've updated the parser to pull the initial quantity value and the CTA button text from desktop, so it now shows "Update Cart" and the correct qty based on what you have in your cart. As far as the breadcrumb goes, maybe that's something we can look into later.

web/app/app-provider.jsx
DanielBrookRoberge: We can definitely indicate the difference to the `PDP` component if we use a different `routeName` here, say `cartEditPage`.

web/app/containers/pdp/partials/pdp-add-to-cart.jsx
```diff
@@ -31,7 +31,9 @@ const PDPAddToCart = ({quantity, setQuantity, onSubmit, disabled}) => {
                 <label htmlFor="quantity">Quantity</label>
 
                 <div className="u-flexbox u-margin-bottom-lg u-margin-top">
-                    <Stepper {...stepperProps} />
+                    {quantity &&
```
DanielBrookRoberge: This means that we will not render the Stepper while loading, since `quantity` will be zero then. Do we want to render a skeleton? (we could also do `disabled={!quantity}`)

web/app/containers/pdp/selectors.js
```diff
@@ -32,7 +32,8 @@ export const getSelectedProduct = createGetSelector(
     Immutable.Map()
 )
 
-export const getItemQuantity = createGetSelector(getSelectedPdp, 'itemQuantity', 1)
+export const getItemQuantity = createGetSelector(getSelectedPdp, 'itemQuantity')
+export const getCTAText = createGetSelector(getSelectedPdp, 'ctaText')
```
DanielBrookRoberge: Do we want to have a default value for this?

web/app/containers/pdp/partials/pdp-add-to-cart.jsx
jnurse: A skeleton might be a good option. Setting `disabled={!quantity}` won't work though. I need to prevent it from rendering until we have the quantity to make sure the initialValue for the stepper gets set properly, otherwise whatever initialValue it gets before quantity exists will be used and the stepper's state won't update after receiving a new one.

web/app/app-provider.jsx
jnurse: Yeah we could, I'm wondering if we need to though. By parsing the CTA text and quantity value from the desktop content I was able to resolve the differences between the two pages, so unless we want to make additional design changes I'm not sure we need to make this change.

DanielBrookRoberge: I was thinking maybe we could get the breadcrumb changes like Jeremy was talking about this way, but I also understand if it's just not important right now.

web/app/containers/pdp/partials/pdp-add-to-cart.jsx
DanielBrookRoberge: That's a good point. Reminds me that maybe we should have just made Stepper stateless and maintained the value outside the component (say in the Redux store). Things to think about.

PR #336, opened 2017-02-16
Change PLP/PDP File Name's and Code Name References

bendvc:
We are going to drop the usage of acronyms in file/component names in favour of more descriptive names. This is better for readability and is more consistent with other template names

 **JIRA**: [WEB-1161](https://mobify.atlassian.net/browse/WEB-1161)

## Changes
- Change file and code references for pdp to product-details
- Change file and code references for plp to product-list
- Update code naming for tests
- Rename ProductList component so we don't have confusion, its new name is OrderDetails which is more inline to what the component actually is.

## How to test-drive this PR
- Run `npm run dev`


web/app/containers/pdp/constants.js
```diff
@@ -1 +0,0 @@
-export const PDP_ITEM_ADDED_MODAL = 'pdp-item-added-modal'
```
jeremywiebe: Is it possible to fix these as `git mv`s instead of delete/create actions?

web/tests/system/workflows/guest-checkout.js
```diff
@@ -1,26 +1,26 @@
 import process from 'process'
 import Home from '../page-objects/home'
-import PLP from '../page-objects/plp'
-import PDP from '../page-objects/pdp'
+import ProductList from '../page-objects/product-list'
+import ProductDetails from '../page-objects/product-details'
 // import Cart from '../page-objects/cart'
```
jeremywiebe: Unrelated, but any idea why these are commented out?

web/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## To be released
 - Add option to analyze bundle after build. Set `MOBIFY_ANALYZE` environment variable to `true` before running any build. Eg. `MOBIFY_ANALYZE=true npm run prod:build`
+- Rename plp and pdp containers to product-list and ProductDetails
```
bendvc: Not missing anything.. just eff'd up. Fixed that tho

web/tests/system/workflows/guest-checkout.js
bendvc: :man_shrugging:

web/app/app-provider.jsx
```diff
@@ -17,13 +17,13 @@ const AppProvider = ({store}) => (
...
-                <Route component={PLP} path="new-arrivals.html" routeName="productListPage" />
-                <Route component={PLP} path="charms.html" routeName="productListPage" />
-                <Route component={PDP} path="*.html" routeName="productDetailsPage" />
+                <Route component={ProductList} path="potions.html" routeName="productListPage" />
+                <Route component={ProductList} path="books.html" routeName="productListPage" />
+                <Route component={ProductList} path="ingredients.html" routeName="productListPage" />
+                <Route component={ProductList} path="supplies.html" routeName="productListPage" />
+                <Route component={ProductList} path="new-arrivals.html" routeName="productListPage" />
+                <Route component={ProductList} path="charms.html" routeName="productListPage" />
+                <Route component={ProductDetails} path="*.html" routeName="productDetailsPage" />
```
bendvc: I struggled with the same issue. In this case I think product details is ok as semantically it's correct in that it's showing a collection of individual details of a product.

Alternatively as you mentions in talking that just product would be better, and I agree, but it's a little far from industry standards. 

So lets just leave it as is for now.

web/app/containers/pdp/constants.js
bendvc: I'm not sure, might just be the way tower does things. But when I looked at the commit to merge into develop, it marked most of the ew files and 'r' for renamed. So it might fix itself in that merge.

jeremywiebe:
Resolve conflicts, but 👍  once they're resolved.

PR #337, opened 2017-02-17
Filter list of available regions in shipping page

fractaltheory:
Also display an input box if no regions are available to display.

## Changes
- Adds a set of new selectors to retrieve both the available regions, as well as the currently selected Country (id)
- Using those selectors, add a new selector that provides a filtered list of regions (based on the selected country id)
- Conditionally display input box with `region` name if no regions are available to show (also update submission action to include this form value)

## How to test-drive this PR
- Run `npm run dev`
- Add item to cart and visit `/checkout/cart/`
- Check out to proceed to shipping page
- Note that on initial load, only US states are shown
- Selecting a country like `Canada` will update list with provinces, while selecting something like `Cameroon` will display an empty input field


web/app/store/checkout/locations/selectors.js
```diff
@@ -1,9 +1,24 @@
...
+import {getCheckout, getForm} from '../../selectors'
 
 export const getLocations = createGetSelector(getCheckout, 'locations', Immutable.Map())
-
 export const getCountries = createGetSelector(getLocations, 'countries', Immutable.List())
-
 export const getRegions = createGetSelector(getLocations, 'regions', Immutable.List())
+
+export const getShippingForm = createSelector(getForm, (form) => Immutable.fromJS(form.shippingForm))
+export const getShippingFormValues = createGetSelector(getShippingForm, 'values', Immutable.Map())
```
jnurse: There should already be selectors for these keys already under store -> form -> selectors

web/app/store/checkout/locations/selectors.js
```diff
@@ -1,9 +1,24 @@
...
+export const getDefaultRegionEntry = createGetSelector(getRegions, 0)
+
+// Filter list of available regions based on what user has selected as country
+export const getRegionsForCountry = createSelector(getRegions, getSelectedCountryID, (regions, id) => {
+    return regions.filter((region) => region.get('country_id') === id)
+})
+
+// Insert default region option at top of list if we have regions to display
+export const getAvailableRegions = createSelector(getRegionsForCountry, getDefaultRegionEntry, (regions, defaultRegion) => {
+    return regions.size === 0 ? regions : regions.unshift(defaultRegion)
```
jnurse: [non-blocking] could `regions.unshift(defaultRegion)` be re-written as `[defaultRegion, ...regions]`

web/app/store/checkout/locations/selectors.js
fractaltheory: Nice: https://github.com/mobify/progressive-web-scaffold/pull/337/commits/2ebd8238d67adab54777e8e1949957083876c1cc

web/app/store/checkout/locations/selectors.js
fractaltheory: Well that would technically be possible, but it's simpler at the end when we call `selectorToJS` in `mapStateToProps` that all nested objects inside `regions` are Immutable too. Otherwise we'd have to convert them inside this selector and output a fully raw JS array of objects.

PR #340, opened 2017-02-17
Cart estimate shipping

jnurse:
This PR replaces the hardcoded shipping rate with correct values, and implements the estimate shipping modal submission

 **Linked PRs**: n/a

## Changes
- Refactored estimate shipping API call to be re-used on cart & checkout shipping pages
- Fetched shipping method & process checkout data when cart data is received
- Wrapped estimate shipping form in redux-form so we can access selected values
- Added action to submit estimate shipping form 

## How to test-drive this PR
- Run `npm run dev`
- Add an item to your cart
- click "calculate" on the cart page
- enter values in the estimate shipping form and submit
- a network request should go out with the values you entered (note all shipping is the same regardless of location right now, so shipping rate will not change)
- proceed to the checkout shipping page
- enter values for country, region and postal code
- confirm that a network request goes out with the new values entered (again shipping rate will not change)


fractaltheory:
I think this change would benefit from region filtering, implemented in another PR: https://github.com/mobify/progressive-web-scaffold/pull/337

web/app/containers/checkout-shipping/actions.js
```diff
@@ -18,6 +18,9 @@ export const process = ({payload: {$, $response}}) => {
     return receiveData(checkoutShippingParser($, $response))
 }
 
+
+export const fetchShippingMethods = () => fetchShippingMethodsEstimate('shippingForm')
```
fractaltheory: We could use the `SHIPPING_FORM_NAME` constant string here, right?

jnurse: Yup, good catch! 😄

web/app/store/checkout/shipping/actions.js
```diff
@@ -0,0 +1,40 @@
...
+            address.country_id = formValues.country_id
+            address.region_id = formValues.region_id
+            address.postcode = formValues.postcode
+            if (formValues.region) {
+                address.region = formValues.region
+                // Remove the region_id in case we have an old value
+                delete address.region_id
+            }
+        }
+        const getEstimateURL = `https://www.merlinspotions.com/rest/default/V1/${isLoggedIn ? 'carts/mine' : `guest-carts/${entityID}`}/estimate-shipping-methods`
```
fractaltheory: I think we've been trying to use paths that don't include the hostname, to allow for compatibility with staging environments.

web/app/store/checkout/shipping/actions.js
```diff
@@ -2,30 +2,37 @@
...
 
 
+
 export const fetchShippingMethodsEstimate = (formKey) => {
     return (dispatch, getState) => {
         const currentState = getState()
         const isLoggedIn = getIsLoggedIn(currentState)
         const formValues = getFormValues(formKey)(currentState)
         const entityID = getCustomerEntityID(currentState)
+        const registeredFieldNames = getFormRegisteredFields(formKey)(currentState).map(({name}) => name)
```
fractaltheory: `getFormRegisteredFields(formKey)(currentState)` can returned `undefined` which will throw an error when `map` is called.

fractaltheory:
Seems to work 👍

PR #345, opened 2017-02-20
Fixes the PWA navigation stuff

jeremywiebe:
Originally we started with using an Astro event to communicate the PWA navigation. I changed it to use an RPC method. At some point the changes to support the RPC on the web side got lost. This fixes the PWA side of in-app navigation.

 **JIRA**: [HYB-962](https://mobify.atlassian.net/browse/HYB-962)
 **Linked PRs**: N/A

## Changes
- Swift from using an Astro event to RPC method (this allows Astro to sit on the navigation while it swaps views before letting it continue)
- Expose `Astro.trigger` through `astro-integration` so that we only ever use `astro-integration` in the PWA.

## How to test-drive this PR
- Run `npm run dev`
- Run the app
- Navigate to some potions
- Navigate back using the header bar
Note: You'll know it's working if you get stacking. No stacking means it's broken!


web/app/app-provider.jsx
```diff
@@ -8,10 +8,30 @@ import {Cart, CheckoutConfirmation, CheckoutPayment, CheckoutShipping, Home, Log
 import CheckoutHeader from './containers/checkout-header/container'
 import CheckoutFooter from './containers/checkout-footer/container'
 
+import {getURL} from './utils/utils'
```
jansepar: I'm not sure about this but I'll ask: is this code that should live in template.jsx?

jeremywiebe: @jansepar We need it in both the template.jsx and here.  I'm open to other designs! :)

kerrmarin:
I'm closing this in favour of https://github.com/mobify/progressive-web-scaffold/pull/350 which contains only the commit that fixes the navigation

PR #348, opened 2017-02-21
Fix Lighthouse Test Integration

johnboxall:
(brief description / PR title)

## Changes
- (change1)

## How to test-drive this PR
- (necessary config changes)
- Run `npm run dev`


circle.yml
```diff
@@ -17,17 +17,15 @@ test:
             pwd: web
         - npm test -- --runInBand:
             pwd: web
-        # Disabled on January 27, 2017. For some reason, Lighthouse can't connect to Chrome as of Chrome v56.
-        # - sudo env "PATH=$PATH" npm run test:pwa-ci:
-        #     pwd: web
-        # Enable smoke-test when automated workflow tests have been implemented.
+        - sudo env "PATH=$PATH" npm run test:pwa-ci:
```
johnboxall: @ellenmobify would love to see a comment here about why we need `sudo`.

Also, would be cool to bundle the download chrome code above into a script like we do here: https://github.com/mobify/test-lighthouse-scores/blob/master/scripts/install-chrome-latest-stable.sh

It _may_ also be the case that we need to set `LIGHTHOUSE_CHROMIUM_PATH`: https://github.com/mobify/test-lighthouse-scores/blob/master/circle.yml#L11

web/scripts/lighthouse-local.sh
```diff
@@ -0,0 +1,15 @@
+OUTPUT_PATH=./lighthouse/audit-local.html
```
johnboxall: @ellenmobify this is the most complicated of these scripts – would love to see the broad strokes commented: "Spin up the devserver, proxy the website, changing the mobify tag to point to the devserver, start a instance of chrome that can load our proxy site (insecure certs) and then run a lighthouse test on that"

johnboxall:
@ellenmobify also, if you could delete https://github.com/mobify/progressive-web-scaffold/blob/develop/web/lighthouse/config.json as part of this PR, that'd be the bestest.

PR #351, opened 2017-02-22
Re-enable Lighthouse test on CI

ellenmobify:
Run Lighthouse tests on CircleCI again, with some changes

## Changes
- Move steps (starting server, starting proxy, running the test) into their own scripts

## How to test-drive this PR
- Checkout this branch
- `cd web`
- `npm i`
- `npm run test:pwa-prod` to test against the production site (https://www.merlinspotions.com)
- `npm run dev` in one tab and then `sudo npm run test:pwa-local` in another tab to test against the local build with Preview (via proxy) 

or 
- Rebuild https://circleci.com/gh/mobify/progressive-web-scaffold/tree/lighthouse-commands

The Lighthouse report will be opened in your default browser or can be found under the Artifacts tab on Circle.

*Notes* 
- We're making use of the `--allow-insecure-localhost` flag in order to test a service worker in dev environment (localhost). This has historically been buggy and was only recently fixed in Chrome 57 (still in beta) https://bugs.chromium.org/p/chromium/issues/detail?id=561820. If the Lighthouse score is lower than expected, it could be because the service worker is not being fetched due to the dev environment.


kerrmarin:
You can ignore the build failure by buddybuild - we're still setting it up :)

web/scripts/install-google-chrome-stable.sh
```diff
@@ -0,0 +1,7 @@
+# Install the latest stable version of Chrome.
+# https://circleci.com/docs/build-image-trusty/#google-chrome
+
+wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
+sudo apt-get update
+sudo apt-get --only-upgrade install google-chrome-stable
```
jasecode: Think my editor and this one's complaining about no newline at the end of the file.

circle.yml
```diff
@@ -5,10 +5,7 @@ dependencies:
     cache_directories:
         - "web/node_modules"
     pre:
-        - wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
-        - sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
-        - sudo apt-get update
-        - sudo apt-get --only-upgrade install google-chrome-stable
+        - ./scripts/install-google-chrome-stable.sh
```
jasecode: +1

web/scripts/install-google-chrome-stable.sh
```diff
@@ -0,0 +1,7 @@
+# Install the latest stable version of Chrome.
```
jeremywiebe: This should have a shebang to open the script. 
```bash
#!/usr/bin/env bash
```

web/package.json
```diff
@@ -116,9 +116,9 @@
     "test": "cross-env NODE_ENV=test jest",
     "test:all": "npm run lint && npm test -- --coverage",
     "test:e2e": "node ./node_modules/nightwatch/bin/runner.js -c ./tests/system/nightwatch-config.js --suiteRetries 1",
-    "test:pwa-prod": "lighthouse --config-path=./lighthouse/config.json --output=html --output-path=./lighthouse/audit-prod.html https://www.merlinspotions.com",
-    "test:pwa-ci": "concurrently --kill-others --success first --raw \"npm run dev\" \"sleep 35 && npm run test:pwa-local\" && ./lighthouse/check-score.js",
-    "test:pwa-local": "concurrently --kill-others --success first --raw \"npm run proxy\" \"chrome-debug --allow-insecure-localhost\" \"sleep 5 && lighthouse --config-path=./lighthouse/config.json --skip-autolaunch --output=html --output-path=./lighthouse/audit-local.html https://localhost\"",
+    "test:pwa-prod": "bash scripts/lighthouse-prod.sh",
```
jeremywiebe: Why do you have to invoke the script with `bash ...`?

web/scripts/lighthouse-ci.sh
```diff
@@ -0,0 +1,14 @@
+OUTPUT_PATH=./lighthouse/audit-local.html
```
jeremywiebe: This should have an opening shebang
```bash
#!/usr/bin/env bash
```

web/scripts/lighthouse-local.sh
```diff
@@ -0,0 +1,15 @@
+OUTPUT_PATH=./lighthouse/audit-local.html
```
jeremywiebe: This should have an opening shebang
```bash
#!/usr/bin/env bash
```

web/scripts/lighthouse-prod.sh
```diff
@@ -0,0 +1,9 @@
+OUTPUT_PATH=./lighthouse/audit-prod.html
```
jeremywiebe: This should have an opening shebang
```bash
#!/usr/bin/env bash
```

web/scripts/lighthouse-ci.sh
```diff
@@ -0,0 +1,14 @@
...
+URL=https://localhost
+
+concurrently --kill-others --success first --raw \
+	'npm run dev' \
+	'npm run proxy' \
+	'chrome-debug --allow-insecure-localhost' \
+	'sleep 40 && lighthouse \
+		--skip-autolaunch \
+		--output=html \
+		--output-path=./lighthouse/audit-local.html \
```
jeremywiebe: Did you mean to use `$OUTPUT_PATH` here?

web/scripts/lighthouse-ci.sh
```diff
@@ -0,0 +1,14 @@
...
+		https://localhost'
```
jeremywiebe: Should this be `$URL`?

web/package.json
ellenmobify: This was @johnboxall's suggestion and I'm not 100% clear on the rationale, either. John, can you weigh in on this?

johnboxall:
@ellenmobify @jeremywiebe just thinking about this a bit more – we recently made a change to Preview which allows it to be "bypassed" by passing a special HTTP Header.

If we could find a Chrome flag that allows us to add a special HTTP Header ... (or something else that an attacker would not be able to easily do, but we could easily do via scripting like adding a special value to the User-Agent) then we could remove the proxying which is error prone.

https://github.com/mobify/portal_app/pull/3012

cc @kerrmarin

johnboxall:
Also, now that https://github.com/GoogleChrome/lighthouse/pull/1761 has been merged, we should shortly be able to bump the Lighthouse version and remove the `chrome-debug` dance.

PR #385, opened 2017-03-03
Move release and changelog docs to root folder

kerrmarin:
I don't think it makes sense to have two separate docs for CHANGELOG and RELEASE for `web` and `native` folders. There's nothing in the `native` ones and IMHO it makes more sense to have all information/changes in one file.

## Changes
- Delete `native` CHANGELOG and RELEASE files, move `web` CHANGELOG and RELEASE files to root.


jeffkamo:
Is there anything block us from merging this? Personally I don't mind this change. @jeremywiebe thoughts?

PR #392, opened 2017-03-04
Integration manager

jansepar:
This PR is the main work-in-process `develop` tree for the Integration Manager. For historical reference, the web-data team based all initial Integration Manager (IM) work on this tree to shield the rest of the team from the drastic changes made introducing the IM. 

## Changes
- Introduce the Integration Manager
- Extract a Merlin's connector
- Build the Salesforce Commerce Cloud (SFCC) connector (this is not fully complete at time of merging!)

### Running against Merlins:

1. Open `app/main.jsx`.
2. Import the Merlin's connector:

```
import connector from './integration-manager/_merlins-connector'
// import connector from './integration-manager/_sfcc-connector'
```

3. Open Preview using the following link:

    https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=0

### Running against Salesforce Commerce Cloud:

1. Open `app/main.jsx`.
2. Import the SFCC connector:

```
// import connector from './integration-manager/_merlins-connector'
import connector from './integration-manager/_sfcc-connector'
```

3. Open Preview using the following link:

    https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1


web/app/template.jsx
```diff
@@ -35,6 +35,10 @@ const template = (WrappedComponent) => {
 
             dispatch(onRouteChanged(url, WrappedComponent))
 
+            if (WrappedComponent.fetcher) {
+                WrappedComponent.fetcher(url, dispatch)
```
jansepar: Not sure if `fetcher` is what we want here. My feeling is that these methods should be specific to what they are fetching - `fetchProducts`, `fetchProductDetails`, etc. Thoughts?

jnurse: Maybe I'm misunderstanding your comment but I like using `fetcher` here. If we make these fetch methods specific to what's being fetched then we'll end up with a big set of if/else statements to cover all the different templates and the data they need to get, and I'm not sure we want that logic stored in `template.jsx`. 

By using `fetcher` we can make sure that every route/template has a `fetcher` function but within it we can specify what sort of data we're fetching.

jansepar: Yeah fair enough, wouldn't want a set of if/elses in the template.jsx either. I think my comment was less about what it should be in `template.jsx`, and more about what our "interface" as defined by the Integration Manager. But on second look, it looks like we have things like `fetchPdpData` (https://github.com/mobify/progressive-web-scaffold/pull/392/files#diff-5bd6611334a42e8f4ab43bcf9d1adb4cR7), which is what I was looking for.

I guess I'd question the value of template.jsx calling this function for us automatically. To me, it's just adding a layer of indirection that I'm not (yet) convinced is necessary. It made sense to have fetching functionality in template.jsx previously because most of our sites always fetched the desktop version of the page we're on, but since we're not doing that with these refactors, I think we could ditch having `template.jsx` doing any of the action dispatching. Thoughts?

jansepar: I guess the problem is that in order to do that properly, we'd need to have the lifecycle methods available to our container, and we're doing it in template.jsx because the lifecycle methods are there already. But not sure that's a good reason to put it in template.jsx.

I think it would be nice to simply be able to go to a page-level component and see what is happening simply by looking at `componentDidMount`.

jansepar: One thing I think you could argue is, it's possible we may want to control how fetcher is called in the future, and by putting it in template.jsx, it gives us that flexibility (when template.js makes its way into the SDK anyways). But for that I'd apply YAGNI here:

https://github.com/mobify/developer-values#you-aint-gonna-need-it-yagni

web/app/integration-manager/responses.js
```diff
@@ -0,0 +1,46 @@
...
+            id: 'shippingEmailError',
+            showRemoveButton: true
+        }))
+    }
+}
+
+export const onAddToCartSucceess = () => (dispatch) => {
+    dispatch(addToCartComplete())
+    dispatch(openModal(PRODUCT_DETAILS_ITEM_ADDED_MODAL))
+}
```
DanielBrookRoberge: I don't have any big ideas for this unfortunately, but I really don't think that these thunk actions are the best place for this functionality; it's too tied into the details of the Merlin's UI. (I was confused to see `addToCartStarted` up in product-details/actions.js but `addToCartComplete` down here, for example)

One thought I've had is standardizing the Promise return from the command, so you can do 
```js
return myCommand(param)
    .then(() => {/* dispatch UI actions for success */}
    .catch(() => {/* dispatch UI actions for failure */}
```

web/app/template.jsx
```diff
@@ -23,6 +23,10 @@ const template = (WrappedComponent) => {
 
             dispatch(onRouteChanged(url, route.routeName))
 
+            if (WrappedComponent.fetcher) {
+                WrappedComponent.fetcher(url, dispatch)
```
DanielBrookRoberge: Definitely we came up with the name `fetcher` in a hurry. Maybe `getData` instead?

web/app/template.jsx
jeremywiebe: I was thinking about the `componentDidMount` idea. As I understand it `componentDidMount` won't fire when navigating from one route to another when the new route is the same component as the previous.

I think we talked about having a more "reacty" way to do this by having an "overridable" method in `template.jsx`. Similar to the `render` method, but specific to our SDK. 

Maybe just `fetch()`?

DanielBrookRoberge: I don't think we were talking about having the method in `template.jsx`, but instead in the component itself (so `container.jsx`). I don't want to duplicate the logic for determining whether you have to make a new fetch in all of the components, so I still think it should be called from inside the Template component, but what it calls can be a method rather than this ad-hoc addon.

jeremywiebe: We settled on having `template.jsx` call the fetching method (we've renamed it to an `initXyzPage` method) which is defined when you use the template on a route.

PR #396, opened 2017-03-06
Code Split Vendor from Main

jeffkamo:
**JIRA**: [WEB-1236](https://mobify.atlassian.net/browse/WEB-1236)

## Changes
- Use `CommonsChunkPlugin` to split vendor scripts from app [as per this](https://github.com/mobify/lancome-progressive/pull/829)
- Manage asynchronous download for `vendors.js` and `main.js` [as per this](https://github.com/mobify/lancome-progressive/pull/829)

| Before | After |
| ------- | ----- |
| **243.63** (gzipped) screen shot 2017-03-06 at 4 10 59 pm | **152kb** and **89.95** (gzipped)  screen shot 2017-03-06 at 4 09 46 pm |

> For reference, Lancome's `main.js` and `vendor.js` are 178kb and 166kb (gzipped) respectively.

| Timeline Before | Timeline After |
| ---------------- | -------------- |
| screen shot 2017-03-06 at 4 34 02 pm | screen shot 2017-03-06 at 4 00 44 pm |

## How to test-drive this PR
- run `MOBIFY_ANALYZE=true npm run build`, ensure the `node_modules` scripts are moved to `vendor.js`
- [Preview Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=0)
- Ensure `vendor.js` and `main.js` load asynchronously and in parallel in Network tab


jansepar:
We should definitely split astro-client out of there.

jansepar:
Also confused, why is astro-client in app not vendor?

jeffkamo:
@jansepar astro-client.js is hosted locally in `app/vendor/astro-client.js`, so that's probably why. Vendor.js is only `node_modules`

web/README.md
```diff
@@ -139,3 +139,11 @@ cd ../progressive-web-scaffold/web
 npm link progressive-web-sdk
 npm run dev
 ```
+
+## Analyze Bundle Size
+
+To visualize bundle script content, run:
+
+```
+MOBIFY_ANALYZE=true npm run build
```
jeremywiebe: I think the command needs to be `npm run prod:build` right?

jansepar:
There are two possible approaches to solving the "omg astro-client is too big" problem:

1. Only pull in the methods we care about, and don't bother loading astro-client at all (because we don't want that giant bluebird dependency).
2. We load astro-client async. The problem with that, is all of our Astro methods in https://github.com/mobify/progressive-web-scaffold/blob/develop/web/app/utils/astro-integration.js must become promise based.

jansepar:
Other options we talked about:

3. Generate a different file just for Astro
4. Move from Bluebird to ES6 Promises, and import `src/astro-client` from NPM instead of bringing in the vendor folder

jansepar:
👍

PR #397, opened 2017-03-07
Lighthouse tests with Mobify Preview

ellenmobify:
The previous PR (https://github.com/mobify/progressive-web-scaffold/pull/351) to re-enable the Lighthouse tests for CI did not accurately report results when testing a local build through Preview. This stemmed from the use of a proxy to allow Lighthouse to test a local build. 

The addition of https://github.com/mobify/portal_app/pull/3014 means we can now enable requests under a Preview environment without using a proxy. This means we can get accurate Lighthouse scores for local builds. Yay!

## Changes
- Use a custom user agent containing "MobifyPreview" so that Preview will accept our requests.
- Remove files and dependencies related to the proxy.

## How to test-drive this PR
- Run `npm run dev`
- In another tab, run `npm run test:pwa-local`

Or find the build in CI: https://circleci.com/gh/mobify/progressive-web-scaffold/2507. The report will be under the Artifacts tab.


web/scripts/lighthouse-ci.sh
```diff
@@ -1,21 +1,21 @@
...
-		--output-path=${OUTPUT_PATH} \
-		${URL}"
+# Finally, parse the HTML report for the Lighthouse score. 
+# CI will fail the build if the score is below a threshold.
+# See min_lighthouse_score in package.json
+
+# Change www.merlinspotions.com to the project's URL. 
+
+concurrently --kill-others --success --raw \
+	'chrome-debug --allow-insecure-localhost --user-agent="Chrome MobifyPreview"' \
```
johnboxall: @ellenmobify Lighthouse 1.6.0 merged the ability to pass Chrome flags directly to the `lighthouse` command:

https://github.com/GoogleChrome/lighthouse/releases/tag/1.6.0

You should be able to something like `lighthouse --chromeflags='--allow-insecure-localhost ...'

Which should also get rid of the need to use `concurrently` at all here.

web/scripts/lighthouse-local.sh
```diff
@@ -1,17 +1,19 @@
...
-		--skip-autolaunch \
-		--output=html \
-		--output-path=${OUTPUT_PATH} \
-		${URL}"
+# Change www.merlinspotions.com to the project's URL. 
 
-open $OUTPUT_PATH
+concurrently --kill-others --success --raw \
+	'chrome-debug --allow-insecure-localhost --user-agent="Chrome MobifyPreview"' \
+	"sleep 5 && lighthouse --skip-autolaunch --output=html \
```
johnboxall: @ellenmobify Lighthouse 1.6.0 includes the `--view` option which is a cross platform `open`.

You should be able to:
1) Remove `concurrently` by passing the chrome flags directly to LH
2) Pass the `--view` argument to Lighthouse to automatically open the report.

johnboxall:
@ellenmobify looks good – with the Lighthouse 1.6.0 release we can actually make these scripts even less complex and drop using concurrently.

web/scripts/lighthouse-local.sh
ellenmobify: Looks like the `--view` functionality will only look under a `reports` subdirectory.

web/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "babel-runtime": "6.11.6",
     "es6-promise": "4.0.5",
+    "http-server": "^0.9.0",
```
jeffkamo: Should this be moved back down to be alphabetical?

web/lighthouse/cert-db-password
```diff
@@ -0,0 +1 @@
+p4ssword
```
kschmidtdev: Normally we would put this in an environment variable https://mobify.atlassian.net/wiki/display/SYS/Best+Practices+For+Keeping+Secrets+in+Applications

PR #400, opened 2017-03-07
Bug/remove alert confirmation

kerrmarin:
Triggers two astro events when the confirmation page is rendered

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1025

Related PRs: https://github.com/mobify/progressive-web-scaffold/pull/398

## Changes
- Trigger disabling of the checkout alert
- Trigger cart-updating to have zero products

## How to test-drive this PR
- Run `npm run dev`
- Open in app, check out and click the "close" button


web/app/containers/checkout-confirmation/container.js
```diff
@@ -5,13 +5,26 @@ import CheckoutConfirmationDetails from './partials/checkout-confirmation-detail
...
 
-const CheckoutConfirmation = () => (
-    <div className="t-checkout-confirmation">
-        <CheckoutConfirmationSplash />
-        <CheckoutConfirmationDetails />
-        <CheckoutConfirmationQuestions />
-        <CheckoutConfirmationModal />
-    </div>
-)
+import {trigger} from '../../vendor/astro-client'
```
jeremywiebe: We should use the same-named function from `astro-integration`.  My thinking is that PWA's never talk directly to the `astro-client`.  This makes unit testing a bit easier too.

kerrmarin: Done :)

web/app/containers/checkout-confirmation/container.js
```diff
@@ -5,13 +5,26 @@ import CheckoutConfirmationDetails from './partials/checkout-confirmation-detail
...
-        <CheckoutConfirmationDetails />
-        <CheckoutConfirmationQuestions />
-        <CheckoutConfirmationModal />
-    </div>
-)
+import {trigger} from '../../utils/astro-integration'
+
+class CheckoutConfirmation extends React.Component {
+    componentDidMount() {
+        trigger('checkout:disable-alert')
```
jeremywiebe: Would it be possible to merge these two events into a single conceptual event, rather than being so explicit with Astro?  I'm thinking I'd rather see `checkout:completed` where in Astro we interpret that as we need.  Thoughts?

PR #404, opened 2017-03-08
Make cart full height in app

kerrmarin:
When in an app, the header and footer don't show, so we want to show the empty cart UI full-screen.

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1018

## Changes
- Add class for full viewport height
- Use that class on empty cart if running in Astro

## How to test-drive this PR
- Run `npm run dev`
- Run in app
- Go to cart, should be full screen
- Run in web
- Go to cart, should not be full screen


kschmidtdev:
Was there an update to `astro-client.js`?

kerrmarin:
@kschmidtdev not really sure what's going on there. I think there must be a mismatch between versions of some plugin that does the minification and it's changed. I've reverted the changes for it,

kerrmarin:
@jeffkamo I'll add instructions on how to preview an app in a new PR

PR #407, opened 2017-03-08
[Styleguide] Fix asset path and CSS URL handling

mlworks:
## Linked PRs
https://github.com/mobify/crabtree-evelyn-progressive/pull/168
https://github.com/mobify/lancome-progressive/pull/196

## Changes
- Styleguide will include local assets that's used by icons/CSS URLs
- Include icon sprite in styleguide
- Disable URL handling in CSS loader
- Update web's README to use the correct command for previewing styleguide

## How to test-drive this PR
- `npm run styleguide`
- [View styleguide](http://0.0.0.0:4000/)


mlworks:
Jeff's question was about whether we need to include the assetDir or not. We will need it because there's high chances where developers might use CSS background images and if we don't include that in the styleguide, those assets will not render.

PR #408, opened 2017-03-08
Automatically convert Immutable objects in selectors

DanielBrookRoberge:
This PR brings in the `reselect-immutable-helpers` library (which is the open source version of `selector-utils.js`) and uses its newest features to avoid having to wrap only some selectors in `selectorToJS`.

We now have a function `createPropsSelector` that wraps `createStructuredSelector`, wrapping each of the input selectors in `ensureJSSelector`, a function that converts Immutable objects to plain JS but still works if the object isn't Immutable. We can't tell ahead of time from the selector the type of the object, but we can tell when the selector is run. Source for these functions is at https://github.com/DanielBrookRoberge/reselect-immutable-helpers/blob/master/index.js#L31

## Changes
- Convert from `createStructuredSelector` to `createPropsSelector`
- Remove now-unnecessary uses of `selectorToJS`
- Remove duplicate `selectorToJS` definition from `selector-utils.js`

## How to test-drive this PR
- (necessary config changes)
- Run `npm run dev`
- make sure everything still works


web/app/containers/app/container.js
```diff
@@ -136,8 +135,8 @@ App.propTypes = {
     sprite: PropTypes.string,
 }
 
-const mapStateToProps = createStructuredSelector({
```
jeremywiebe: I like the idea of this!!  

Random idea... what if we provided our own `connect` higher order component and did this work in that? In that case we could do the `selectorToJS` work in there and then delegate to the real `connect` call.  That would mean that our `mapStateToProps` objects could just use a simple JS object with selectors.

DanielBrookRoberge: That's not a bad idea, but I think forcing every prop to have its own selector every time might be a barrier early in development. (the `const mapStateToProps = selectorToJS(getMyObject)` thing)

DanielBrookRoberge: Although, we could check whether it's a function or an object, similar to what react-redux does for `mapDispatchToProps`.

DanielBrookRoberge: Flip side though: providing our own `connect` completely nullifies any react-redux help people might get from Google etc. We swapped out the react-router `Router` component, but it provides exactly the same API; this wouldn't.

jeremywiebe: Very much agree with your last point. If we can't retain the same API that really hurts "googleability" which I think hurts our usability.

jeremywiebe:
This looks good. I think there are some documents in the SDK that need to be tweaked now that we don't use `selectorToJS` anymore though. The [template skeleton](https://github.com/mobify/progressive-web-sdk/blob/develop/generator-assets/template-skeleton/container.jsx) also looks like it needs to be updated.

PR #409, opened 2017-03-08
WEB-1246: Batch compress local image assets

mlworks:
**JIRA**: https://mobify.atlassian.net/browse/WEB-1246

## After Compression
saves 148kb,  ~36% savings
screen shot 2017-03-08 at 3 32 34 pm

## Changes
- Add `imagemin-pngquant` plugin to improve PNG assets compression
- Update `clean:images` script to include files stored in the root directory of `static/img`

## How to test-drive this PR
- Add some jpg/png images into the img directory
- Run `npm run clean:images` and see compression happen


web/package.json
```diff
@@ -106,7 +107,7 @@
     "add:page": "sdk-generate page",
     "build-sprites": "npm run clean:svgs && svg-sprite --symbol --symbol-dest='.' --symbol-sprite='sprite.svg' --shape-id-generator='pw-%s' --dest=./app/static/svg/sprite-dist ./app/static/svg/sprite-source/*.svg",
     "clean:svgs": "svgo --enable='removeUselessStrokeAndFill' -f ./app/static/svg/sprite-source",
-    "clean:images": "imagemin-power -v --recursive ./app/static/img/**/* --out-dir=.",
+    "clean:images": "imagemin-power -v --recursive --plugin=pngquant ./app/static/img/* ./app/static/img/*/** --out-dir=.",
```
jeremywiebe: If we quote the path we can get away with just using `\"./app/static/img/*/**\"` for the glob pattern and it looks like it finds all images including ones at the base directory we specify.
>  If an input argument is wrapped in quotation marks, it will be passed to node-glob for cross-platform glob support.

mlworks: @jeremywiebe thanks!! I was wondering how I could write it so it'll find all the images cleanly.

PR #414, opened 2017-03-10
WEB-1236: Optimize script loading

jeffkamo:
The latest in the Merlin's Performance Optimization changes committed upstream from work done on other projects!

 **JIRA**: [WEB-1243](https://mobify.atlassian.net/browse/WEB-1243)

## Changes
- Load scripts asynchronously in the same manner as [we did here](https://github.com/mobify/lancome-progressive/pull/849)
- Impact: not nearly as dramatic as in the above example. In both before/after, the site is blocked by `vendor.js` anyway.

| Before | After |
| ------- | ----- |
| _Notice the red box: that is the time waited before main.js starts to download due to lack of async loading_ screen shot 2017-03-09 at 3 54 07 pm | _Notice how all scripts now load at the same time_ screen shot 2017-03-09 at 5 03 03 pm |

## How to test-drive this PR
- Run `npm run dev`
- Verify that all four scripts load at the same time: `jquery.min.js`, `capture-latest.min.js`, `main.js`, and `vendor.js`


web/app/loader.js
```diff
@@ -84,54 +108,47 @@ if (isReactRoute()) {
...
+        // Load all dependencies asynchronously
+        asyncInitApp()
+        Promise.all([
+            jQueryPromise,
+            capturingPromise,
+            mainPromise,
+            vendorPromise
+        ]).then((result) => {
+            // asyncInitApp will now be able to resolve itself!
+            allAsyncDependenciesReady = true
```
jansepar: We actually don't need to *do* anything after all of these promises run. We just need to make sure that they are all loading in parallel, and that main doesn't execute until vendor is finished loading (which is what `webpackJsonpAsync` is for).

In looking into this, we realized there is possibly a bug in the following scenario:

In app/actions.js, `window.Progressive.initalCapturedDocHTML` could be undefined if the fetch code executes before the HTML is actually finshed loading.

To solve this, `requestCapturedDoc` should actually wait on a promise to guarantee that the page has finished loading.

I'd suggest that promise get created inside of the callback within Capture.init in here:

https://github.com/mobify/progressive-web-scaffold/pull/414/files#diff-fa44066d20629ee51e62e770692a8900R124

web/app/loader.js
```diff
@@ -84,54 +107,45 @@ if (isReactRoute()) {
...
+            onload: (resolveOnload) => {
                 window.Capture.init((capture) => {
-                    // NOTE: by this time, the captured doc has changed a little bit from original desktop.
-                    // It now has some of our own assets (e.g. main.css) but they can be safely ignored.
-                    window.Progressive.initialCapturedDocHTML = capture.enabledHTMLString()
-                    resolve()
+                    // NOTE: by this time, the captured doc has changed a little
+                    // bit from original desktop. It now has some of our own
+                    // assets (e.g. main.css) but they can be safely ignored.
+                    window.Progressive.initialCapturedDocHTML = new Promise((resolveCapture) => {
```
jansepar: So, this isn't the right location to create this promise. We want the promise to be created right away so that if capture.js loads after main.js, `initialCapturedDocHTML` won't be undefined.

I'd also rename this to `.capturedDocHTMLPromise`.

web/app/loader.js
```diff
@@ -84,54 +107,45 @@ if (isReactRoute()) {
...
+        })
+
+        const vendorPromise = makeScriptPromise({
+            id: 'progressive-web-vendor',
+            src: getAssetUrl('vendor.js')
+        })
+
+        // Load all dependencies asynchronously
+        asyncInitApp()
+        Promise.all([
```
jansepar: We don't actually need `Promise.all` here - you use `Promise.all` when you want to execute code after all async things finish, but in this case, we're not actually waiting for anything.

web/app/loader.js
```diff
@@ -84,54 +107,45 @@ if (isReactRoute()) {
...
-            body.appendChild(capturing)
         })
 
-        Promise.all([capturingPromise]).then(() => loadMain(body))
+        const mainPromise = makeScriptPromise({
+            id: 'progressive-web-main',
+            src: getAssetUrl('main.js')
+        })
+
+        const vendorPromise = makeScriptPromise({
```
jansepar: Since we're not actually waiting for something to happen at the end of all these scripts loading, these don't have to be promises.

web/app/loader.js
```diff
@@ -51,6 +53,24 @@ if (isReactRoute()) {
...
+        // Setting UTF-8 as our encoding ensures that certain strings (i.e.
+        // Japanese text) are not improperly converted to something else. We
+        // do this on the vendor scripts also just in case any libs we
+        // import have localized strings in them.
+        script.charset = 'utf-8'
+        script.async = true
+        script.id = id
+        script.src = src
+        script.onload = typeof onload === typeof function() {}
+            ? onload // resolve
```
jansepar: Don't need `resolve` since this isn't a promise anymore.

jansepar:
@jeffkamo this is looking good! 👍

PR #418, opened 2017-03-10
fix checkout header for payment and confirmation steps

gsaray:
Remove sign in button from payment step

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1245

## Changes
- remove sign in from payment and confirmation steps

## How to test-drive this PR
- Run `npm run dev`
- Add an item to cart
- go through the checout

### DIFF
--- a/web/app/router.jsx
+++ b/web/app/router.jsx
@@ -45,17 +45,17 @@ const Router = ({store}) => (
                 <Route component={ProductList} path="charms.html" routeName="productListPage" />
                 <Route component={ProductDetails} path="checkout/cart/configure/id/*/product_id/*/" routeName="cartEditPage" />
                 <Route component={ProductDetails} path="*.html" routeName="productDetailsPage" />
-                <Route component={CheckoutShipping} path="checkout/" routeName="checkingShipping" Header={CheckoutHeader} Footer={CheckoutFooter} />
+                <Route component={CheckoutShipping} path="checkout/" routeName="checkingShipping" Header={CheckoutHeader} Footer={CheckoutFooter} headerHasSignIn={true}/>
                 {/*
                     The URL for the payment page on desktop is /checkout/#payment,
                     but routing wasn't working properly when using this as the
                     route path so we specify a fetchUrl to make sure when we
                     fetch it's using the URL for the desktop page
                 */}
-                <Route component={CheckoutPayment} path="checkout/payment/" fetchUrl="/checkout/#payment" routeName="checkout-payment" Header={CheckoutHeader} headerHasSignIn={true}
+                <Route component={CheckoutPayment} path="checkout/payment/" fetchUrl="/checkout/#payment" routeName="checkout-payment" Header={CheckoutHeader} headerHasSignIn={false}
                     Footer={CheckoutFooter}
                     />
-                <Route component={CheckoutConfirmation} path="checkout/onepage/success/" routeName="checkout-confirmation" Header={CheckoutHeader} headerHasSignIn={true}
+                <Route component={CheckoutConfirmation} path="checkout/onepage/success/" routeName="checkout-confirmation" Header={CheckoutHeader} headerHasSignIn={false}
                     Footer={CheckoutFooter}
                     />
             </Route>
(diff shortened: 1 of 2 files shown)


web/app/router.jsx
```diff
@@ -52,8 +52,12 @@ const Router = ({store}) => (
                     route path so we specify a fetchUrl to make sure when we
                     fetch it's using the URL for the desktop page
                 */}
-                <Route component={CheckoutPayment} path="checkout/payment/" fetchUrl="/checkout/#payment" routeName="checkout-payment" Header={CheckoutHeader} Footer={CheckoutFooter} />
-                <Route component={CheckoutConfirmation} path="checkout/onepage/success/" routeName="checkout-confirmation" Header={CheckoutHeader} Footer={CheckoutFooter} />
+                <Route component={CheckoutPayment} path="checkout/payment/" fetchUrl="/checkout/#payment" routeName="checkout-payment" Header={CheckoutHeader} headerHasSignIn={true}
+                    Footer={CheckoutFooter}
+                    />
+                <Route component={CheckoutConfirmation} path="checkout/onepage/success/" routeName="checkout-confirmation" Header={CheckoutHeader} headerHasSignIn={true}
```
jeffkamo: This `headerHasSignIn={true}` logic is perhaps a bit backwards. I would expectr `true` to mean that the header should in fact have the sign in link visible. But the way this works at the moment is the opposite: `true` means the sign in links is not present.

I would recommend just reversing the logic so that true means the sign in link is visible, and false the link is not.

web/app/router.jsx
```diff
@@ -45,15 +45,19 @@ const Router = ({store}) => (
...
                     fetch it's using the URL for the desktop page
                 */}
-                <Route component={CheckoutPayment} path="checkout/payment/" fetchUrl="/checkout/#payment" routeName="checkout-payment" Header={CheckoutHeader} Footer={CheckoutFooter} />
-                <Route component={CheckoutConfirmation} path="checkout/onepage/success/" routeName="checkout-confirmation" Header={CheckoutHeader} Footer={CheckoutFooter} />
+                <Route component={CheckoutPayment} path="checkout/payment/" fetchUrl="/checkout/#payment" routeName="checkout-payment" Header={CheckoutHeader} headerHasSignIn={false}
+                    Footer={CheckoutFooter}
+                    />
+                <Route component={CheckoutConfirmation} path="checkout/onepage/success/" routeName="checkout-confirmation" Header={CheckoutHeader} headerHasSignIn={false}
+                    Footer={CheckoutFooter}
+                    />
```
jeffkamo: Minor: For the most part (we're not terribly consistent yet, but...) let's try and keep the end part of the tag at the same indent depth as the openning part. So...

```
<Route
    ...
/>
```

web/app/router.jsx
```diff
@@ -45,15 +45,19 @@ const Router = ({store}) => (
                 <Route component={ProductList} path="charms.html" routeName="productListPage" />
                 <Route component={ProductDetails} path="checkout/cart/configure/id/*/product_id/*/" routeName="cartEditPage" />
                 <Route component={ProductDetails} path="*.html" routeName="productDetailsPage" />
-                <Route component={CheckoutShipping} path="checkout/" routeName="checkingShipping" Header={CheckoutHeader} Footer={CheckoutFooter} />
+                <Route component={CheckoutShipping} path="checkout/" routeName="checkingShipping" Header={CheckoutHeader} Footer={CheckoutFooter} headerHasSignIn={true}/>
```
jeffkamo: Minor: A prop that is declared but past nothing is implied to be true. So you can actually do this:

```
<Route ... headerHasSignIn />
```

And that's the same as...

```
<Route ... headerHasSignIn={true} />
```

So my suggestion would be to stick to the prior example, and change the `headerHasSignIn` so the `false` instances can just exclude mentioning it at all. For example...

```
<Route ... />
```

Would be the same as...

```
<Route ... headerHasSignIn={false} />
```

jeffkamo:
Oh, and just be sure to fix that failing CI result before merging!

PR #422, opened 2017-03-13
Fix image width and modal

gsaray:
Cart product card’s columns not set as per mocks, should be % width so that the image gets bigger on larger devices

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1245

## Changes
- add a prop to product item for width% and make modal shrink to content

## How to test-drive this PR
- Run `npm run dev`
- go to PDP and add an item to cart

### DIFF
--- a/web/app/components/product-item/index.jsx
+++ b/web/app/components/product-item/index.jsx
@@ -13,7 +13,7 @@ const ProductItem = ({
     image,
     price,
     title,
-    widthPercentage
+    customWidth
 }) => {
     const classes = classNames('c-product-item', 'u-flexbox', 'u-row-reverse', className)
 
@@ -43,7 +43,7 @@ const ProductItem = ({
 
 
             {image &&
-                <div className="u-padding-end u-flex-none" style={{width: widthPercentage}}>
+                <div className="u-padding-end u-flex-none" style={{width: customWidth}}>
                     {image}
                 </div>
             }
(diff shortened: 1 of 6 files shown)


web/app/components/product-item/index.jsx
```diff
@@ -81,6 +82,11 @@ ProductItem.propTypes = {
      * Designates the ProductItem's unit price
      */
     price: PropTypes.node,
+
+    /**
+     * Designates the image width
+     */
+    widthPercentage: PropTypes.string
```
jeffkamo: `widthPercentage` seems pretty strict as far as a name goes, but it should be allowed to accept any unit. Can this be renamed to allow generic values as well?

web/app/components/product-item/index.jsx
```diff
@@ -42,7 +43,7 @@ const ProductItem = ({
 
 
             {image &&
-                <div className="u-padding-end u-flex-none">
+                <div className="u-padding-end u-flex-none" style={{width: widthPercentage}}>
```
jeffkamo: What happens to the `style` prop if `widthPercentage` is null or undefined (since it is an optional value)?

gsaray: there won't be any inline style if `customWidth` is not defined

jeffkamo:
👍 after CircleCI is fixed!

PR #432, opened 2017-03-14
WEB-1250: Reduce initial image asset downloads

mlworks:
**JIRA**: https://mobify.atlassian.net/browse/WEB-1250

## Changes
- Add local lazy load image component so it'll automatically load the content when you scroll to it

**Before**
> Please ignore the assets MP_banner_x.jpg and MerlinsPotions_Logo_Light.png, those are leaked assets from Chrome

screen shot 2017-03-14 at 3 33 22 pm

**After**
> Please ignore the assets MP_banner_x.jpg and MerlinsPotions_Logo_Light.png, those are leaked assets from Chrome

screen shot 2017-03-14 at 3 39 53 pm

## How to test-drive this PR
- Run `npm run dev`


web/app/components/lazy-load-image/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+    render() {
+        const {
+            className,
+            image,
+            placeholder
+        } = this.props
+
+        const classes = classNames(componentClass, className, {
+            // 'c--modifier': bool ? true : false
```
jeffkamo: If this isn't being used, let's delete it

web/app/components/lazy-load-image/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    checkVisible() {
+        const topPosition = this.el.getBoundingClientRect().top
+
+        if (topPosition <= window.innerHeight + window.scrollY - this.props.threshold) {
+            this.setState({
+                visible: true
+            })
+        }
```
jeffkamo: Actually, would it make sense that once the image is visible, the scroll event was also removed (instead of waiting for unmount)? Might as well just not run anything, if not necessary. I'm imaging product pages with hundreds or more products on them... The normal Lazy Load component should probably be used in that case, but just in case this component is abused #futureproof.

web/app/components/lazy-load-image/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+}
+
+
+LazyLoadImage.propTypes = {
+    /**
+     * Adds values to the `class` attribute of the root element
+     */
+    className: PropTypes.string,
+
+    image: PropTypes.node,
```
jeffkamo: This could use a comment description.

web/app/components/lazy-load-image/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+    placeholder: PropTypes.node,
```
jeffkamo: This could also use a comment description.

web/app/components/lazy-load-image/index.jsx
jeffkamo: And, should this be required? It's rather an explicit part of the component afterall... or should this actually accept image props? Like the src and alt text?

It seems a bit silly that you have to import the Image and LazyLoadImage component to use it.

Or, should this component be decoupled from the image component and should lazy load any kind of generic content? This component prop type doesn't prevent the user from passing it non-image nodes, afterall, so anyone could make it work with anything.

web/app/containers/footer/partials/footer-social-icons.jsx
```diff
@@ -13,8 +15,19 @@ const FooterSocialIcons = () => {
...
+                        <LazyLoadImage
+                            image={
+                                <Image
+                                    src={getAssetUrl(icon)}
+                                    alt={title}
+                                    height="32"
+                                    width="32"
+                                />
+                            }
+                            placeholder={<span className="u-visually-hidden">{title}</span>}
```
jeffkamo: This might actually be confusing to a blind user. There's no explanation why there is text that is suddenly replaced with an image. In fact, if the user has the span highlighted and the image swaps out, they could loose focus and become lost... I'm not sure the placeholder is very useful as far as accessibility is concerned.

Or perhaps an "image loading..." message would be better.

web/app/containers/home/partials/home-categories.jsx
```diff
@@ -9,19 +9,25 @@ import Image from 'progressive-web-sdk/dist/components/image'
...
+            image={
+                <Image
+                    src={getAssetUrl(`static/img/categories/${alt.trim().replace(/\s+/g, '-')
+                            .toLowerCase()}@2x.png`)}
+                    alt={alt}
+                    height="60"
+                    width="60"
+                />
+            }
+            placeholder={<SkeletonBlock height="60" width="60" />}
```
jeffkamo: Since this is the same as the `SkeletonBlock` above, maybe the placeholder element could be in a `const` and reused in both places.

web/app/components/lazy-load-image/index.jsx
mlworks: Yea maybe renaming this component to `LazyLoadContent`? That way anything can go inside this component.

mlworks: +1 to required, I'm keeping this component generic in terms of contents that it receives, so I opted to not make props for src/alt text.

jeffkamo:
Good to go once `content` is changed to `children`

PR #423, opened 2017-03-13
[Integration Manager] Reorganization of Home and Product Details

jeremywiebe:
Continuation of initial work to figure out the integration manager. This PR moves more "Merlins" HTML logic (parsing) into the connector. 

 **JIRA**: (link to JIRA ticket)
 **Linked PRs**: (links to corresponding PRs, optional)

## Changes
- (change1)

## How to test-drive this PR
- (necessary config changes)
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- (how to access the new / changed functionality -- fixtures, URLs)

PR #438, opened 2017-03-15
[Integration Manager] Reorganize demandware connector

jnurse:
This PR updates the demandware connector to match the reorganized merlins connector.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1256
 **Linked PRs**: #423

## Changes
- Move home and PDP commands for demandware connector into their own folder

## How to test-drive this PR
- Set up tag injector for the demandware sandbox (http://mobify-tech-prtnr-na03-dw.demandware.net/on/demandware.store/Sites-2017refresh-Site/default/Home-Show)
- Run `npm run dev`
- Preview [the demandware site](https://preview.mobify.com/?url=http%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- The homepage should work correctly

### DIFF
--- a/web/app/integration-manager/_demandware-connector/commands.js
+++ b/web/app/integration-manager/_demandware-connector/commands.js
@@ -1,141 +1,15 @@
 import {makeRequest} from 'progressive-web-sdk/dist/utils/fetch-utils'
-import {urlToPathKey} from '../../utils/utils'
 import {receiveCartContents} from '../../store/cart/actions'
-import {receiveHomeData, receiveNavigationData} from '../responses'
-import {receiveProductDetailsProductData, receiveProductDetailsUIData} from '../product-details/responses'
-import {parseProductDetails, parseBasketContents} from './parser'
+import {parseBasketContents, getCurrentProductID} from './parser'
 
-const SITE_ID = 'Sites-2017refresh-Site'
-const API_TYPE = 'shop'
-const API_VERSION = 'v17_2'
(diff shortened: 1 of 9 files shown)


web/app/integration-manager/_demandware-connector/commands.js
```diff
@@ -1,141 +1,15 @@
...
-        })
+import * as homeCommands from './home/commands'
+import * as productDetailsCommands from './product-details/commands'
 
-}
 
-
-export const addToCart = () => (dispatch) => {
+const addToCart = () => (dispatch) => {
     return initDemandWareSession()
```
jansepar: Are we making a fetch to get the Demandware session every time?

If so, one thought is that we could turn `initDemandwareSession` into a promise that returns the result of the call to get the session with `then` without fetching everytime.

Or alternatively like your TODO says, put it in the store.

jnurse: Yes right now we're initializing a session each time, but my thought was we would address this when we tackle the rest of the global demandware data like if the user is logged in etc.

jansepar: Kk if we have a plan in place for addressing this then I won't block the PR being merged.

PR #442, opened 2017-03-15
Service worker in SDK

DanielBrookRoberge:
This PR switches the Merlins worker to use the central worker code in the SDK. Currently that is in PR https://github.com/mobify/progressive-web-sdk/pull/549/; until we merge, release, and update to the new release, this will fail on Circle

 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/549/

## Changes
- Use the SDK worker 

## How to test-drive this PR
- Install the associated SDK branch
- Run `npm run dev`
- See that a new service worker is registered with version 0.2.0 (as printed in the console)
- Reload the home page twice
- See that all three caches are populated
- Go offline
- See that the home page loads again.


web/worker/main.js
```diff
@@ -1,103 +1,12 @@
 /* global PROJECT_SLUG, DEBUG */
 /* eslint-env worker, serviceworker */
 
-import toolbox from 'sw-toolbox'
+import worker from 'progressive-web-sdk/dist/worker/main'
 
-const cachebreaker = /b=([^&]+)/.exec(self.location.search)[1]
 const CAPTURING_URL = 'https://cdn.mobify.com/capturejs/capture-latest.min.js'
```
jnurse: This looks like it's the same URL the worker in the SDK is using as a default. Is this meant to be an example of how you can pass your own capturing URL into the worker?

DanielBrookRoberge: Yes. I'm not 100% sure I still want to keep the default in the SDK, either.

web/worker/main.js
```diff
@@ -1,103 +1,12 @@
...
-    e.waitUntil(
-        self.clients.claim()
-            .then(() => caches.keys())
-            .then((cacheKeys) => cacheKeys.filter(
-                (key) => cacheNames.indexOf(key) === -1 && !key.endsWith('$$$inactive$$$')))
-            .then((keysToDelete) => keysToDelete.map((key) => caches.delete(key)))
-            .then((promises) => Promise.all(promises))
-    )
+worker({
+    capturingURL: CAPTURING_URL,
```
jeremywiebe: Given we have this defaulted, would it be clearer to not "override" the capturing URL here?

web/worker/main.js
jeremywiebe: I _think_ we want to keep the default in the SDK ... but I'd vote to remove the "override" here.

web/worker/main.js
```diff
@@ -1,103 +1,12 @@
...
+    slug: PROJECT_SLUG,
+    debug: DEBUG
```
jeremywiebe: I believe this is `isDebug` in the SDK.

PR #443, opened 2017-03-16
Web 1284  web 1285  web 1286  fonts

jeffkamo:
**JIRA**: [WEB-1284](https://mobify.atlassian.net/browse/WEB-1284), [~~WEB-1285~~](https://mobify.atlassian.net/browse/WEB-1285), [~~WEB-1286~~](https://mobify.atlassian.net/browse/WEB-1286)

## Changes
- Add `webfontloader` library
- Defer font loading to a bit later in the app life cycle (when App container mounts)
- Fix FOUT by displaying system fonts until Oswald is loaded, THEN show Oswald fonts, as per [Zack Leatherman's method here](https://www.zachleat.com/web/comprehensive-webfonts/#fout-class)

### Animated!

Notice how in _After_ the `Shop` text is always visible. In both cases, the category name is placeholder if the text is not available.

| Before | After |
| ------ | ------ |
| fout-before | fout-after |

### First Paint of Text

Notice how there text in the _Before_ is invisible. At the same point in time, the _After_ screenshot has text perfectly visible!

| Before | After |
| ------ | ------ |
| earliest-text-before | earliest-text-after |

### Network

Notice how the loading of fonts has been deferred to a later point in time. In the _Before_ screen, you can see that the Google API CSS file starts to downloaded after `main.css` finishes downloaded (that's where we were importing the fonts originally) and the font files themselves download before content is rendered (we can assume this is the case because image assets haven't started downloading yet).

In the _After_, the Google API CSS downloads at the same time as we fetch the SVG asset. The font files themselves end up downloading just after image assets.

I assert that the overall impact this will have is that large impact assets will load first, and lesser impact assets (fonts) will load slightly later. I hope this means visual progression as measured by Web Page Test (i.e. page speed index) will improve by this change.

| Before | After |
| ------ | ------ |
| network-before | network-after |

## How to test-drive this PR
- Run `npm run dev`
- Verify that fonts load as intended


web/app/styles/base/_typography.scss
```diff
@@ -7,8 +7,8 @@
 //
 // 1. Linter workaround
 
-$base-typography__fonts: 'https://fonts.googleapis.com/css?family=Oswald:200,400'; // 1
-@import url($base-typography__fonts);
+// $base-typography__fonts: 'https://fonts.googleapis.com/css?family=Oswald:200,400'; // 1
```
mlworks: Let's remove these while we are at it.

PR #451, opened 2017-03-16
Windows instructions for avoiding SSL errors. Bundle push instructions update.

jvoll:
**JIRA**: none
 **Linked PRs**: none

## Changes
- Adds instructions for preventing SSL errors on Windows machines. During training we have to do this every time to reduce the number of errors/warnings first time users see in their consoles when running the scaffold for the first time.
- Updates instructions to use `npm` scripts for saving credentials and deploying bundles. The old instructions involved using `grunt` and checking that the `mobify-client` node module is installed (which we don't seem to install by default).

## How to test-drive this PR
- Read the instructions. I've followed the SSL instrcutions before with Windows 7/10 users and I just tried them out on a Windows 10 VM with success.
- Have not tried the bundle instructions today but I've seen them work in the past...except on Windows where our script will error (scheduled for work next sprint: https://mobify.atlassian.net/browse/WEB-1282)

Thanks @jeremywiebe for originally putting together these instructions for me (I mostly just stole those)!!!

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -7,13 +7,18 @@ npm install
 npm run dev
 ```
 
-If you will be deploying bundles to Mobify Cloud, then follow these steps:
+## Deploying Bundle to Mobify Cloud
 
-- Make sure you have the `mobify-client` npm module installed
-- Authorize your computer to push bundles by:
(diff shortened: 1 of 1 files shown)


DanielBrookRoberge:
:+1:

PR #455, opened 2017-03-17
Lighthouse URL handling

ellenmobify:
DRY up the Lighthouse scripts using a single URL variable.

## Changes
- Set siteUrl in package.json
- Use the siteUrl value in the Lighthouse scripts
- Also allow a URL argument to be passed in eg. `npm run test:pwa-prod https://www.something.com`

## How to test-drive this PR
- `npm run test:pwa-prod` or `npm run test:pwa-prod https://www.someurl.com`
- `npm run dev` in one tab and then `npm run test:pwa-local`
- Rebuild on CI: https://circleci.com/gh/mobify/platform-scaffold/tree/lighthouse-url-vars


web/scripts/lighthouse-local.sh
```diff
@@ -19,6 +21,6 @@ lighthouse \
 	--output=html \
 	--output-path=${OUTPUT_PATH} \
 	--disable-device-emulation=true \
```
johnboxall: @ellenmobify 🤔 this is going to be tricky for us to solve.

We want Mobile Emulation ... but we want to add a value to the User Agent...

* https://github.com/GoogleChrome/lighthouse/blob/ade2d88fc7d4e52546ad2000eb098d6dbad6689e/lighthouse-core/lib/emulation.js#L105
* https://github.com/GoogleChrome/lighthouse/blob/91ad83a37c25e213a6ef8e9166e86bdc6fc5dbdd/lighthouse-core/gather/driver.js#L686

We might want to file a bug w/ the Lighthouse Team and maybe we'll luck out and they'll extend things to allow us to override the UA even when emulation is enabled.

web/scripts/lighthouse-prod.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 OUTPUT_PATH=./reports/audit-prod.html
-URL=https://www.merlinspotions.com
+# Can also pass in URL as first argument. If none given, fallback to what's specified in package.json's siteUrl key.
+URL=${1-$npm_package_siteUrl}
```
johnboxall: @ellenmobify this comment explaining what's going on here probably isn't necessary as these are fairly standard bash-isms.

PR #460, opened 2017-03-20
Use package.json's siteUrl

ellenmobify:
There are a couple places where we've hardcoded a URL:
* The preview links from running `npm run dev`
* The `site.js` file that configures environments for end-to-end tests

## Changes
- Instead of a hardcoded URL, use the value from `package.json`'s 'siteUrl' 

## How to test-drive this PR
- Checkout this branch
- Change the value for `siteUrl` in `package.json`
- `npm run dev` and observe that the Preview link has the site URL set to your given URL
- Or `npm run smoke-test` and observe that the Preview page has the site URL set to your given URL

screen shot 2017-03-20 at 3 43 52 pm


web/dev-server/logger.js
```diff
@@ -33,7 +37,7 @@ const logger = {
 
         console.log(
             chalk.bold('Preview URL: ') + // eslint-disable-line prefer-template
-            chalk.magenta(`https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A${port}%2Floader.js&disabled=0&domain=&scope=0`) + '\n' +
+            chalk.magenta(`https://preview.mobify.com/?url=${encodedUrl}&site_folder=https%3A%2F%2Flocalhost%3A${port}%2Floader.js&disabled=0&domain=&scope=0`) + '\n' +
```
jeremywiebe: Nice.  Would you be able to also extract the `site_folder` value and `urlencode()` it above like you've done on line 28 with the `url` also?  Makes the construction of the url so much clearer.

web/package.json
```diff
@@ -90,6 +90,7 @@
     "svg-sprite": "1.3.6",
     "svgo": "0.7.1",
     "text-loader": "0.0.1",
+    "urlencode": "1.1.0",
```
jeremywiebe: Sorry, missed this earlier.  Does `encodeURIComponent` not work? I'd like to limit how many dependencies we have whenever possible.

ellenmobify: Didn't know about `encodeURIComponent`. It does the job! 😄

PR #462, opened 2017-03-21
[Integration manager] login for merlins connector

jnurse:
This PR adds the login page actions to the merlins connector (form submission and fetch page content). 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1257
 **Linked PRs**: n/a

## Changes
- Moved login form submission action to the merlins connector
- Moved login page process action to the merlins connector

## How to test-drive this PR
- Run `npm run dev`
- navigate to: https://www.merlinspotions.com/customer/account/login/
- the login and create account pages should load correctly and form submissions should work correctly


web/app/integration-manager/_merlins-connector/login/commands.js
```diff
@@ -0,0 +1,58 @@
...
+        })
+        .catch((error) => {
+            if (error.name !== SubmissionError) {
+                reject(new SubmissionError({_error: 'Failed to login due to network error.'}))
+            }
+        })
+}
+
+export const submitLoginForm = (href, formValues, resolve, reject) =>
+    submitForm(href, formValues, '.form-login', resolve, reject)
```
jnurse: yes it does

web/app/integration-manager/_merlins-connector/login/commands.js
```diff
@@ -0,0 +1,58 @@
...
+            if (isFormResponseInvalid($response, formSelector)) {
+                const error = {
+                    _error: 'Username or password is incorrect'
+                }
+                return reject(new SubmissionError(error))
+            }
+            if (isRunningInAstro) {
+                jsRpcMethod('user:loggedIn', [])()
+            }
+            window.location.href = '/customer/account'
```
jnurse: I spoke with Mitch, and he'd like to navigate to the account page in this case. Since the account page on Merlin's hasn't been added to the PWA we may want to leave this as is for now. The URL will be different between Merlin's and demandware, and we can't just programatically navigate to the route since there is no route for the account page.

jnurse: Actually I just had a thought maybe we can pass the account page URL back to the app and it navigates to whatever URL was passed back.

web/app/containers/login/actions.js
```diff
@@ -133,7 +94,8 @@ export const submitSignInForm = (formValues, resolve, reject) => {
             formValues[input.name] = input.value
         })
 
-        return sendForm(href, formValues, '.form-login', resolve, reject)
+        return login(href, formValues, resolve, reject)
+            .then(handleLoginSuccess)
```
DanielBrookRoberge: Probably we should store the href and hidden inputs in the connector, since they're irrelevant to the UI. 

We can also avoid having to pass the `resolve` and `reject` functions from redux form if we throw in the command on error and do a `.catch((error) => reject(error))` here.

web/app/containers/login/actions.js
```diff
@@ -96,28 +74,11 @@ const validateRegisterForm = (formValues) => {
...
-        .catch((error) => {
-            if (error.name !== SubmissionError) {
-                reject(new SubmissionError({_error: 'Failed to login due to network error.'}))
-            }
-        })
+const handleLoginSuccess = (href) => {
+    if (isRunningInAstro) {
+        jsRpcMethod('user:loggedIn', [])()
+    }
+    window.location.href = href
```
jnurse: In this case no, because the account page we navigate to has not been implemented in the PWA yet. I've made a few additional fixes to this as part of the work for the demandware login, so that it will first check if the URL is a route, and only do window.location.href if it's not.

DanielBrookRoberge: Ok, good to know.

PR #468, opened 2017-03-27
Update readme for Native 

kerrmarin:
Adds some extra documentation for the native apps

## Changes
- Add steps to get set up with native app development

## How to test-drive this PR
- Read the changes, 👍 👎


johnboxall:
Thanks @kerrmarin – This is a good first pass at expanding the instructions.

Non-blocking feedback:

The expanded steps still don't meet our baseline requirements. If you follow these steps, the app will not run correctly.

The [default configuration](https://github.com/mobify/platform-scaffold/blob/develop/native/app/config/baseConfig.js#L12) assumes you're running the PWA locally – if you are not, the app will not load. Can we mention this (or improve the error handling scenario for it?)

https://github.com/mobify/developer-values#️-write-great-documentation

I'd separate the steps to "setup" the project and the steps to "run" it.

Beyond that, I believe there is a lot more information here worth mentioning:

* How should you configure Xcode? _I've found I need to configure the `scaffold` as the target._
* How do you actually go about developing the app? _What's the workflow? After you change a file will it automatically reload? If not, how should you reload it?_

If we had an explicit Astro setup article on astro.mobify.com we could link to that.

We should also pillage relevant content from:

https://github.com/mobify/astro-scaffold

kerrmarin:
Hey @johnboxall thanks for the feedback. I'll update this PR with more info as soon as I get some time 👍

native/README.md
```diff
@@ -1,17 +1,35 @@
...
+- Clone this repository
+- Navigate to the `native` folder in your terminal
+- Run `npm run deps`
 
+## Run
+To run the app, follow these steps.
+
+### iOS
+To run the app for iOS, follow these steps:
+- Open the `scaffold.xcworkspace` file in the
```
jeremywiebe: "... in the..." where?

native/README.md
```diff
@@ -1,17 +1,35 @@
...
 
+- Clone this repository
+- Navigate to the `native` folder in your terminal
+- Run `npm run deps`
 
+## Run
+To run the app, follow these steps.
+
+### iOS
+To run the app for iOS, follow these steps:
```
jeremywiebe: This sentence feels like we're re-stating exactly what is said in the above sentence under the **Run** heading. Maybe reword or trim it down entirely.  "follow these steps..."

native/README.md
```diff
@@ -1,17 +1,35 @@
...
-\* We recommend using [nvm](https://github.com/creationix/nvm#installation) to manage your node and npm versions. 
+We recommend using [nvm](https://github.com/creationix/nvm#installation) to manage your node and npm versions. 
 
-#Generator
+## Setup
 
-You can use the [generator](scripts/generator.sh) to generate a new Progressive Mobile app.
+Before you get started, take some time to look through [the Getting Started Guide](http://astro.mobify.com/latest/guides/before-you-begin/). Once you're familiar with progressive apps, follow these steps to set up this project to start developing:
 
+- Clone this repository
```
jeremywiebe: I wonder if it'd be easier to just present this as a series of steps in a bash shell. 

```sh
# Clone the repo
$ git clone git@github.com:mobify/platform-scaffold.git

$ cd platform-scaffold/native

# Install dependencies
$ npm run deps
```

native/README.md
```diff
@@ -1,17 +1,35 @@
...
+- Run `npm run deps`
 
+## Run
+To run the app, follow these steps.
+
+### iOS
+To run the app for iOS, follow these steps:
+- Open the `scaffold.xcworkspace` file in the 
+- Make sure you have the `scaffold` target selected.
+- Build and run the app (`CMD+R`)
```
jeremywiebe: John pointed me to a sweet way to present the keyboard shortcuts:
```html
<kbd>⌘R</kbd>
```
Which comes out as: <kbd>⌘**R**</kbd>

native/README.md
kerrmarin: the only issue I see with that is that if people don't use SSH keys and always clone via https they might get an error. We could add both methods though (one commented out)?

jansepar:
👍 this is definitely better than what we have right now. Hope you don't mind @kerrmarin but I'm merging this so that it makes it's way into the next Platform Scaffold release.

jansepar:
Annnnd it works :D merging

PR #472, opened 2017-03-28
Don't show preloader if running in Astro

kerrmarin:
In Astro,  until we decide whether we want to show a preloader and what that preloader will look like, we'll disable the preloader.

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1040

## Changes
- Don't display preloader if in Astro

## How to test-drive this PR
- Run `npm run dev`
- Preview in app by running app (preview is on by default)
- You shouldn't see any header preloader when opening the cart, for example

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -45,7 +45,9 @@ const asyncInitApp = () => {
 }
 
 if (isReactRoute()) {
-    displayPreloader(preloadCSS, preloadHTML, preloadJS)
+    if (!isRunningInAstro) {
+        displayPreloader(preloadCSS, preloadHTML, preloadJS)
+    }
 
     // Create React mounting target
     const body = document.getElementsByTagName('body')[0]


jeffkamo:
Instead of not showing the preloader at all, would it bet better to...

a) Load and display an App-only preloader? Or...
b) Add environment classes to the preloader, so the CSS can be customized accordingly?

jeffkamo:
Should we make such a fix now rather than wait?

kerrmarin:
Hey @jeffkamo, in Astro we already have the ability to show a native loader that covers the screen. The issue for us is that we see a flash of the second header-bar for about 1/4 of a second just before the content renders and we want to avoid that. Because of the small impact, I'm not sure there's much value in adding an app-only preloader, but I've added a ticket for that for @henryh15 to prioritize: https://mobify.atlassian.net/browse/HYB-1041

jeffkamo:
Oh I see, I wasn't aware the app had it's own mechanism for displaying a loading state. I guess that makes sense!

In that case 👍

kerrmarin:
Thanks for the review @jeffkamo 💯

PR #473, opened 2017-03-28
Make reducers more explicit

DanielBrookRoberge:
The `mergePayloadForActions` function can be confusing and splits the reducer code between two different styles. This PR removes it in favour of the simpler `mergePayload` function.

The changes in this PR will prevent generated reducers from working until the below SDK PR is merged and released.

 **JIRA**: N/A
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/567

## Changes
- Remove all uses of the utility function
- Remove the function itself

## How to test-drive this PR
- Preview Merlin's
- See that nothing has changed.


web/app/utils/reducer-utils.js
```diff
@@ -1,12 +1,6 @@
 import Immutable from 'immutable'
 export const mergePayload = (state, {payload}) => state.mergeDeep(payload)
 
-export const mergePayloadForActions = (...actions) => {
-    const handlers = {}
-    actions.forEach((action) => { handlers[action] = mergePayload })
-    return handlers
-}
-
 const isList = Immutable.List.isList
```
jnurse: We should probably move these utils into the SDK at some point.

DanielBrookRoberge: Agreed!

PR #466, opened 2017-03-24
Add AstroConnector component to allow interaction between webviews

kerrmarin:
To coordinate different webviews in the app we need to be able to notify them of changes that will affect the store, this is an initial implementation of a `NativeConnector` component that will serve to listen for/trigger notifications to the native parts of the app.

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1025
 **Linked PRs**:  https://github.com/mobify/astro/pull/790

## Changes
- Move all app events to `AppEvents.js`
- Add `NativeConnector` component
- Connect app to native connector for events

## How to test-drive this PR
- Change `package.json` to point to linked PR
- Run `npm run deps` in the `native` folder
- Change `baseConfig.js` in `native/config` to enable preview if disabled
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Go to product
- Go back
- Go to product
- Add to cart
- Go to cart
- Remove from cart
- Close cart
- Go back
- Cart icon should not show a red dot with a 1

PR #474, opened 2017-03-28
Update redux store if a webview logs in

kerrmarin:
This depends on https://github.com/mobify/platform-scaffold/pull/466

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1007
 **Linked PRs**: https://github.com/mobify/platform-scaffold/pull/466

## Changes
- Refresh the cart when a user logs in

## How to test-drive this PR
- Change `package.json` to point Astro to`feature/topplugin-rpc` 
- Run `npm run deps` in the `native` folder
- Change `baseConfig.js` in `native/config` to enable preview if disabled
- Run `npm run dev`
- Go to account
- Log in with an account with items in their cart
- Go to a different tab
- Cart icon should show the same number of items as the accoutn tab


native/app/controllers/tabController.js
```diff
@@ -44,14 +40,22 @@ TabController.init = async function(tabItem) {
...
+    navigationView.on('cart:count-updated', (data) => {
+        AppEvents.trigger(AppEvents.updateCart, data)
     })
 
     navigationView.on('open:cart-modal', () => {
         headerController.showCartModal()
     })
 
+    AppEvents.on(AppEvents.cartNeedsUpdate, async () => {
+        navigationView.trigger('cart:needs-update')
```
kschmidtdev: Is there a reason that `'cart:needs-update'` is used explicitly here instead of `AppEvents.cartNeedsUpdate` again? To be clear that navigationView is getting the event? There is some repetition so curious if it's worth renaming or reusing.

jeremywiebe: @kschmidtdev There's two types of events in play here. Web->Astro and Astro->Astro.  We've typically used raw strings for web->astro to make it clear that you can't depend on the string constant for consistency.

kerrmarin:
@kschmidtdev this PR depends on changes to the other PR so most of the changes in this PR will "disappear" as soon as #466 merges in. i.e. I branched off of 466 to work on this

native/app/controllers/tabController.js
kschmidtdev: That makes sense - any chance we can use different strings then ;)

kschmidtdev:
@kerrmarin that makes sense, just was confused about the `package.json` part of the instructions 😄

kerrmarin:
Doh! I copied and pasted the instructions from #466... I'll update it

kerrmarin:
@kschmidtdev this now reflects the changes that should be included in this PR

native/app/controllers/tabController.js
jeremywiebe: @kschmidtdev That's a good idea actually.  I wonder if we should work to use a convention where the event name contains some prefix to signal that it's "special" also.  

Something like `bridge:<event-name>`

kerrmarin:
@jeremywiebe @kschmidtdev do you want me to do that as part of this PR or a different one?

native/app/controllers/tabController.js
kerrmarin: @jeremywiebe @kschmidtdev do you want me to do that as part of this PR or a different one?

jeremywiebe:
@kerrmarin I'd say a different one.

PR #476, opened 2017-03-29
Remove `astro-client` from `main.js`

jeffkamo:
**JIRA**: [WEB-1261](https://mobify.atlassian.net/browse/WEB-1261)

## Changes
- Utilize the astro-detect.js module to reduce bloat to main.js
- Load astro-client only when needed (when in astro!)
- Delete vendor/astro-client.js and pull in astro as an NPM module instead
- Delete the script that copy/pasted astro-client.js into vendor
- Cleanup our usages of the astro-integration utilities by wrapping them in promises, etc.

## How to test-drive this PR

**Web**
- Run `npm run dev`
- Verify that the app works as intended

**App**
- Coming Soon

## Todo

- [x] Update `mobify-progressive-web-app` version to latest Astro release BEFORE we merge PR!
- [x] Verify changes on iOS app
- [x] Verify changes on Android app

## Impact

Main.js shrunk from 70.93kb → **47.97kb**

Savings of **22.96kb**

| Before | After |
| --- | --- |
| before | after |


web/package.json
```diff
@@ -15,6 +15,7 @@
     "es6-promise": "4.0.5",
     "immutable": "3.8.1",
     "lodash.frompairs": "4.0.1",
+    "mobify-progressive-app-sdk": "http://github.com/mobify/astro#develop",
```
kschmidtdev: Astro is a private repo - this should probably point to a NPM release. We may also want to update `RELEASE.md` with something to remind to check on updating the version

jeffkamo: Good point, I left this as is since a new version of Astro has not yet been released.

Will add a to-do to ensure this is corrected before merged!

jeffkamo:
Blocked until a new version of Astro is released, which requires this PR being merged:
https://github.com/mobify/astro/pull/790

kerrmarin:
@jeffkamo beta3 has been released and scaffold updated to point to beta3 👍

web/app/containers/header/container.test.js
```diff
@@ -1,18 +1,30 @@
...
 import React from 'react'
+import * as AstroIntegration from '../../utils/astro-integration'
 
 import ConnectedHeader from './container'
 const Header = ConnectedHeader.WrappedComponent
 
 describe('The header', () => {
-    test('Should not render header at all if running in an Astro app', () => {
-        const header = shallow(<Header isRunningInAstro={true} />)
+    const originalisRunningInAstro = AstroIntegration.isRunningInAstro
```
jeremywiebe: Super nitpick, but "Is" should be capitalized in the variable name.  (Same with header tests above)

web/app/containers/header/container.test.js
```diff
@@ -1,18 +1,30 @@
...
+
+    afterEach(() => {
+        // Value of `isRunningInAstro` might have been overwritten in test
+        AstroIntegration.isRunningInAstro = originalisRunningInAstro
+    })
+
+    test('1 Should not render header at all if running in an Astro app', () => {
```
jeremywiebe: What's the reasoning behind numbering the tests? Numbering becomes a maintenance burden if you add tests in between or before existing ones.

web/app/loader.js
```diff
@@ -17,6 +19,7 @@ initCacheManifest(cacheHashManifest)
 const IS_PREVIEW = /mobify-path=true/.test(document.cookie)
 
 const CAPTURING_CDN = '//cdn.mobify.com/capturejs/capture-latest.min.js'
+const ASTRO_CLIENT_CDN = '//assets.mobify.com/astro/astro-client-0.18.0.min.js'
```
jeremywiebe: This should be `1.0.0-beta3` (https://github.com/mobify/astro/releases/tag/1.0.0-beta3)

native/scripts/update-web-astro-client.sh
```diff
@@ -20,5 +20,3 @@ if [ ! -e "$ROOT/node_modules/mobify-progressive-app-sdk/dist/astro-client.js" ]
         npm run build:astro_client
     popd
 fi
-
-cp -f "$ROOT/node_modules/mobify-progressive-app-sdk/dist/astro-client.js" ../web/app/vendor/astro-client.js
```
jeremywiebe: I'm not sure if we want to go this route. We decided to use the `native` side's npm module so we wouldn't have two versions of Astro being used (one in `/native` and one in `/web`).  Now we're going back to that.  What are the benefits of this change?

web/app/loader.js
```diff
@@ -3,7 +3,9 @@ import {displayPreloader} from 'progressive-web-sdk/dist/preloader'
 import cacheHashManifest from '../tmp/loader-cache-hash-manifest.json'
 import {isRunningInAstro} from './utils/astro-integration'
 
-window.Progressive = {}
+window.Progressive = {
+    AstroPromise: {}
```
jeremywiebe: I think we should initialize this to `Promise.resolve({})` so that it actually is a Promise even when not running in Astro.

web/app/containers/header/container.test.js
jeffkamo: Oops, that was there from troubleshooting something. Will delete.

web/app/loader.js
jeffkamo: Speaking of which, @jeremywiebe is there a better way to manage this dependency so that it matches what's in package.json?

native/scripts/update-web-astro-client.sh
jeffkamo: Fair point, although what's changed is that the `astro-client` itself as a dependency is being loaded via CDN instead of via the local vendor path. That decision was made on the assumption that we wanted to load `astro-client` conditionally instead of loading it as part of the bundle (main.js).

At the moment, the only static dependency is `astro-detect.js`. Depending on it in the way you described above (having it in `/native` and not also `/web`) makes sense, and I can change that copy/paste task to do so for `astro-detect.js` instead.

Does that work?

web/app/loader.js
jeffkamo: Good point! Will do!

native/scripts/update-web-astro-client.sh
jeremywiebe: Yes, I think that'd be good.  Having astro as an npm dep on the web side means a fairly large number of new npm modules being installed when we do `npm install` in `/web`  I'd like to restrict that as much as possible.  Thanks.

web/app/loader.js
jeremywiebe: Glad to pair this out, but DBR suggested we use a `DefinePlugin` in `base.loader.js`.  That basically does a text replacement of some symbol with a value.  Meaning we can read the `native/package.json` in `base.loader.js` and get the Astro version and then use the `DefinePlugin` to replace `ASTRO_VERSION` with the actual version.

jeffkamo:
Verbal 👍  from @jeremywiebe

PR #478, opened 2017-03-30
[Integration manager] merlins checkout shipping

jnurse:
This PR updates some of the checkout shipping commands for Merlins in the connector. This template was already partially moved into the connector, but this PR cleans up a few parts to make the demandware connector work easier. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1307
 **Linked PRs**: n/a

## Changes
- Moved checkout shipping commands into the checkout folder of the connector
- Moved UI fail/success actions back to the app and throw errors in connector commands that fail

## How to test-drive this PR
- Run `npm run dev`
- Add an item to the cart
- navigate to the checkout shipping page
- confirm the page still works as expected


web/app/containers/checkout-shipping/actions.js
```diff
@@ -1,3 +1,57 @@
...
+import {submitShipping as submitShippingCommand, checkCustomerEmail as checkCustomerEmailCommand, checkoutSignIn} from '../../integration-manager/checkout/commands'
+import {getShippingFormValues} from '../../store/form/selectors'
+import {addNotification, removeNotification} from '../app/actions'
 
 export const showCompanyAndApt = createAction('Showing the "Company" and "Apt #" fields')
+export const receiveCheckoutShippingData = createAction('Receive Checkout Shipping Data')
+
+const onShippingEmailRecognized = () => {
+    return (dispatch) => {
+        dispatch(receiveCheckoutShippingData({customerEmailRecognized: true}))
```
DanielBrookRoberge: This could use a `setCustomerEmailRecognized` action or similar.

web/app/containers/checkout-shipping/actions.js
```diff
@@ -1,3 +1,57 @@
...
+
+export const onShippingLoginError = (errorMessage) => {
+    return (dispatch) => {
+        dispatch(addNotification({
+            content: errorMessage,
+            id: 'shippingEmailError',
+            showRemoveButton: true
+        }))
+    }
+}
```
DanielBrookRoberge: If we don't need to do anything here besides dispatch a single action, it can just be a plain function that returns the `addNotification` action object.

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -0,0 +1,168 @@
...
+
+export const fetchShippingMethodsEstimate = (formKey) => {
+    return (dispatch, getState) => {
+        const currentState = getState()
+        const isLoggedIn = getIsLoggedIn(currentState)
+        const formValues = getFormValues(formKey)(currentState)
+        const entityID = getCustomerEntityID(currentState)
+        const registeredFieldNames = getFormRegisteredFields(formKey)(currentState).map(({name}) => name)
+        // Default values to use if none have been selected
+        const address = {country_id: 'US', region_id: '0', postcode: null}
```
DanielBrookRoberge: This could be defined outside of the function, maybe?

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -0,0 +1,168 @@
...
+            })
+    }
+}
+
+export const fetchCheckoutShippingData = (url) => (dispatch) => {
+    return dispatch(fetchPageData(url))
+        .then(([$, $response]) => {
+            // entity_id is used for API calls
+            const customerEntityID = parseCheckoutEntityID($response)
+            dispatch(receiveEntityID({customerEntityID}))
```
DanielBrookRoberge: If we define this action with `customerEntityID` as a parameter then we won't need to build an object here (which is a little inconsistent with other actions like this)

web/app/integration-manager/_merlins-connector/checkout/commands.js
jnurse: We can definitely do that, but if we do we'll need to make sure we're using a copy of that data here, so that we always start with the default values and don't modify them.

PR #479, opened 2017-03-30
[Integration manager] demandware shipping

jnurse:
This PR adds the ability to submit shipping information in the checkout.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1304
 **Linked PRs**: n/a

## Changes
- Renamed region and country fields in checkout forms
- Added DW connector command to get shipping method estimates
- Added DW connector command to get shipping data if user has already saved their shipping info
- Added DW connector command to submit shipping data

## How to test-drive this PR
- Run `npm run dev`
- Preview [the demandware site](https://preview.mobify.com/?url=http%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Add an item to your cart
- navigate straight to https://mobify-tech-prtnr-na03-dw.demandware.net/checkout/ (don't go to the cart page since it hasn't been set up yet)
- confirm the page works as expected
- entering valid information and proceeding to the next page should work (the next page will be missing content)


web/app/containers/checkout-shipping/actions.js
```diff
@@ -41,7 +44,19 @@ export const submitSignIn = () => (dispatch, getState) => {
 export const submitShipping = () => (dispatch, getState) => {
     const currentState = getState()
     const formValues = getShippingFormValues(currentState)
-    return dispatch(submitShippingCommand(formValues))
+    const {firstname, lastname} = splitFullName(formValues.name)
+    const address = {
+        firstname,
+        lastname,
+        ...formValues
+    }
```
DanielBrookRoberge: This will include the `name` field, do you think that would cause problems?

web/app/integration-manager/_demandware-connector/checkout/commands.js
```diff
@@ -0,0 +1,140 @@
...
+                    if (responseJSON.fault) {
+                        throw new SubmissionError({_error: 'Unable to save shipping data'})
+                    }
+                })
+        })
+}
+
+export const checkCustomerEmail = () => () => Promise.resolve(true)
+
+export const checkoutSignIn = () => (dispatch) => dispatch(noop)
```
DanielBrookRoberge: Is there a reason why these two are different? Is one a placeholder for future work?

web/app/integration-manager/_merlins-connector/app/commands.js
```diff
@@ -31,5 +32,6 @@ export const fetchPageData = (url) => (dispatch) => {
 }
 
 export const initApp = () => (dispatch) => {
+    dispatch(setCheckoutURL(CHECKOUT_URL))
```
DanielBrookRoberge: Do we expect the checkout URL to change during the course of the app? The problem with putting it in the Redux store is that it won't work right unless the router is set up to match. 

That said, that opens a huge can of worms about how the URL scheme will vary by connector; this seems like a useful way to deal with it for now.

web/app/integration-manager/_demandware-connector/checkout/commands.js
jnurse: Both aren't really valid commands for the demandware app (at least not currently) but we need checkCustomerEmail to be a promise that returns a value (so it doesn't break other functionality), where as checkoutSignIn doesn't really ever get called. There's probably a better way of handling that, TBH I set up the checkoutSignIn noop to get rid of an error and forgot to revisit it.

web/app/containers/checkout-shipping/actions.js
jnurse: No we need all three for demandware (firstname, lastname and fullname). Both demandware and merlin's need first and last name split, so I thought it made more sense to split them here before sending them to the connector. Chances are any other connectors we make will need them split as well.

web/app/integration-manager/_merlins-connector/app/commands.js
jnurse: We don't expect the URL to change, I definitely considered having this as a constant only, but I thought since the URL is required by throughout the app's UI, as well as potentially actions within the app this might be easiest. 

Another option would be to set up something like an app config where we can define some of these app URLs, and maybe a few other variables or flags related to the app's behaviour. Then we could swap out which config we use, maybe when we register a connector?

PR #482, opened 2017-03-30
Container code splitting

mlworks:
**JIRA**: https://mobify.atlassian.net/browse/WEB-1311
 **Owners**: @cdegit @mlworks

## Before
main.js -> 48.93.kb gzip
vendor.js -> 151.89kb gzip
screen shot 2017-04-04 at 1 37 14 pm

## After
main.js -> 31.35kb gzip
vendor.js -> 146.12kb gzip
screen shot 2017-04-04 at 1 30 04 pm

## Changes
- Split up containers from `main.js` using `react-loadable` (1.07kb gzipped) and lazy load them
- Update SDK to allow react-loadable to work
- Update eslint parser for babel

## How to test-drive this PR
- Run `npm run dev`
- Preview Merlin's Potions, instead of using `https://localhost:8443/loader.js` use your computer IP as that's how the bundled containers will be loaded


web/app/containers/app/container.js
```diff
@@ -40,6 +54,18 @@ class App extends React.Component {
                 families: ['Oswald:200,400']
             }
         })
+
+        // Lazy load other containers when browser is at the end of frame
+        // to prevent jank
+        requestIdleCallback(() => {
+            UnwrappedCart.preload()
```
cdegit: Should we kick these off all at once or should we spread the preloading across multiple frames?

cdegit:
You'll need the change I made here: https://github.com/mobify/babista-progressive/pull/104/files#diff-622b5b56a595aa778be5c81c63ecf14c in order to get the files to load from the correct location for a bundle. It must be inside webpack's entry file (main.js) in order to work.

cdegit:
If I go directly to https://www.merlinspotions.com/potions.html I get the following errors: 
```
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in. Check the render method of `Loadable`.

Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in. Check the render method of `Loadable`.
```

web/app/containers/app/actions.js
```diff
@@ -12,14 +14,18 @@ import appParser from './app-parser'
 import {ESTIMATE_FORM_NAME} from '../cart/constants'
 import {SHIPPING_FORM_NAME} from '../checkout-shipping/constants'
 
-import Cart from '../cart/container'
-import CheckoutShipping from '../checkout-shipping/container'
-import CheckoutPayment from '../checkout-payment/container'
-import CheckoutConfirmation from '../checkout-confirmation/container'
+import {
+    UnwrappedCart,
```
cdegit: (non-blocking) It might be helpful to explain why these are called 'Unwrapped' and why we're using this version of the container here

web/app/containers/app/container.js
cdegit: (non-blocking) We might want to prioritize this with the earlier pages in the core flow first. They all seem to load pretty quickly anyways, but it might be worthwhile

web/app/utils/utils.js
```diff
@@ -131,3 +131,18 @@ export const getCookieValue = (cookieName) => {
...
+ * Currently requestIdleCallback is only supported in Chrome,
+ * we'll have to provide a fallback for iOS Safari
+ * https://developers.google.com/web/updates/2015/08/using-requestidlecallback
+ * http://caniuse.com/#feat=requestidlecallback
+ */
+export const requestIdleCallback = (fn) => {
+    if ('requestIdleCallback' in window) {
+        return window.requestIdleCallback(fn)
+    } else {
+        return fn()
```
cdegit: I think we still want to put this inside a setTimeout, otherwise on iOS we'll be starting all of the other file loads synchronously.

cdegit: As someone using this polyfill I would expect it to always run code async, not sync.

PR #487, opened 2017-04-03
MSG-461: Page views for Pusheen

fractaltheory:
Current mechanism for increment page view cookie is on a page load - which essentially doesn't occur in a SPA environment. This sets up a basic module for firing a `customEvent` on window, using some agreed-upon event names, that Duncan will listen for (listener set up in the .init method of the target client script)

 **JIRA**: [MSG-461](https://mobify.atlassian.net/browse/MSG-461)
 **Linked PRs**: https://github.com/mobify/duncan/pull/59

## Changes
- Add a `Duncan` module that simply fires a `customEvent` when called
- Hook this up to the `dispatchRouteChange` handler so it's fired on route change

## TODOs
- Eventually replace the local module with an SDK one

## How to test-drive this PR
- Preview
- Navigate around, noticing that `mobify.webpush.activeVisit` cookie increments on route change


jeremywiebe:
This feels alot like analytics. What do you think of using the existing analytics middleware to hook in these Duncan events instead of adding custom code into `template.jsx`?

https://github.com/mobify/platform-scaffold/blob/develop/web/app/utils/analytics/client-analytics.js

fractaltheory:
@jeremywiebe  Cool, that seems like a good spot - but maybe I should create a Duncan distributor rather than including it inside `client-analytics` which is not where I'd expect to find it. What do you think?

jeremywiebe:
@fractaltheory  Yes, definitely make a Duncan distributor.  That `client-analytics` module is basically a stub for partners to begin from.

fractaltheory:
@jeremywiebe took a stab at a Duncan distributor if you want to take a look.

web/app/utils/push-messaging/duncan.js
```diff
@@ -0,0 +1,42 @@
...
+const pageView = (force = false) => {
+    if (hasIgnoredFirstCall || force) {
+        DuncanEvents.dispatch('PAGE_VIEW')
+    }
+
+    hasIgnoredFirstCall = true
+}
+
+export default {
+    _core: DuncanEvents,
```
fractaltheory: It's not currently used anywhere, and you're right that it's meant to be seen as private. I exposed it to provide flexibility in the case a developer needed to extend it without needing to say, push out a new SDK release. e.g. Add an event type so that it's valid and can be fired, or override dispatch functionality, etc.

If you think that's added complexity and YAGNI, it would be fine to just remove that from the export.

jeremywiebe: If that's the intent, then can you add a comment to that effect here?  I'm ok leaving this here, but I'd prefer our pattern be that if we start hitting these cases that we patch using `_core` and then upstream so that we don't have a growing bulk of Duncan behaviour implemented one-off in various builds.

PR #490, opened 2017-04-04
Created a PrivacyPage component as part of a training exercise.

rsexton404:
(brief description / PR title)

 **JIRA**: none
 **Linked PRs**: none

## Changes
- Added PrivacyPage container
- Added a new function to parser-utils called getTextArrayFrom to grab an array of text elements from the DOM

## How to test-drive this PR
- Run `npm run dev`
- Click on Privacy and Cookie Policy link
- Page should load via PWA instead of desktop site when navigating away and coming back
- Intro paragraph should be extracted from desktop site
- List of cookies should be extracted from desktop site
- Look for the paragraph and cookie list in the app state using React dev tools


web/app/containers/privacy-page/actions.js
```diff
@@ -0,0 +1,13 @@
...
+export const receiveData = createAction('Receive PrivacyPage data')
+
+export const process = ({payload}) => {
+    const {$, $response} = payload
+    const parsed = privacyPageParser($, $response)
+    return receiveData(parsed)
+}
+
+// This action will change the `title` key in the local private state
+export const changeTitle = createAction('Change PrivacyPage title', 'title')
```
jvoll: I don't think this is being used anywhere, maybe delete

web/app/containers/privacy-page/container.jsx
```diff
@@ -0,0 +1,58 @@
+import React, {PropTypes} from 'react'
+import {connect} from 'react-redux'
+import {createPropsSelector} from 'reselect-immutable-helpers'
+import SkeletonText from 'progressive-web-sdk/dist/components/skeleton-text'
+
+import {getTitle, getText, getIntro, getCookieList} from './selectors'
+// import * as privacyPageActions from './actions'
```
jvoll: 🔫  (in dev-speak this emoji means you can remove that empty line or commented out stuff). Obviously comments like this wouldn't come in the "early review" stage but they do come when the author tells the reviewer that the PR is "done" (or "Ready for Review").

web/app/containers/privacy-page/container.jsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const mapStateToProps = createPropsSelector({
+    cookieList: getCookieList,
+    intro: getIntro,
+    text: getText,
+    title: getTitle
+})
+
+const mapDispatchToProps = {
```
jvoll: Since this page doesn't actually need to dispatch any actions you can actually remove this totally and remove it from your connect function as well I believe.

web/app/containers/privacy-page/reducer.js
```diff
@@ -0,0 +1,11 @@
+import {handleActions} from 'redux-actions'
+import Immutable from 'immutable'
+import {mergePayload} from '../../utils/reducer-utils'
+import * as privacyPageActions from './actions'
+
+const initialState = Immutable.Map()
+
+export default handleActions({
+    [privacyPageActions.receiveData]: mergePayload,
+    [privacyPageActions.changeTitle]: mergePayload
```
jvoll: this can go too

web/app/containers/templates.js
```diff
@@ -12,6 +12,7 @@ import UnwrappedLogin from './login/container'
 import UnwrappedProductDetails from './product-details/container'
 import UnwrappedProductList from './product-list/container'
 import UnwrappedOffline from './offline/container'
+import UnwrappedPrivacyPage from './privacy-page/container'
```
jvoll: I don't think you and I ever really talked through why this is necessary. Let me know if you have any questions around why we do this.

web/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "progressive-web-scaffold",
-  "projectSlug": "progressive-web-scaffold",
+  "name": "accenture-training-merlins",
```
jvoll: obviously we would never merge this (just for the record). but I know we did this for pushing the bundle to cloud.

web/app/containers/privacy-page/container.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+const containerClass = 't-privacy-page'
+const titleClass = `${containerClass}__title`
+
+const PrivacyPage = ({title, intro, text, cookieList}) => (
+    <div className={containerClass}>
+        { title
+            ? <h1 className={titleClass}>{title}</h1>
+            : <SkeletonText lines={1} type="h1" width="100px" />
+        }
```
jeffkamo: The format we've been using for ternary operators is this:

```jsx
{title ?
    <h1 ... />
:
    <SkeletonText ... />
}
```

This will keep your tab structure more consistent.

web/app/containers/privacy-page/container.jsx
```diff
@@ -0,0 +1,58 @@
+import React, {PropTypes} from 'react'
+import {connect} from 'react-redux'
+import {createPropsSelector} from 'reselect-immutable-helpers'
+import SkeletonText from 'progressive-web-sdk/dist/components/skeleton-text'
+
+import {getTitle, getText, getIntro, getCookieList} from './selectors'
+// import * as privacyPageActions from './actions'
+
+const containerClass = 't-privacy-page'
+const titleClass = `${containerClass}__title`
```
jeffkamo: We're starting to move away from this pattern of using variables for classes when they aren't necessary. Instead, just add the string to the className prop.

This pattern is actually not very helpful for maintainence. It's harder to follow than it seems, so we suggest we avoid it.

Updates to the SDK are coming to remove the pattern from our generators and components.

web/app/containers/privacy-page/container.jsx
```diff
@@ -0,0 +1,58 @@
...
+        }
+        { intro
+            ? <p>{intro}</p>
+            : <SkeletonText lines={4} type="p" width="400px" />
+        }
+        { cookieList
+            ? <div>
+                <h2>All of the cookies</h2>
+                <ul>
+                    {[...cookieList].map((cookie, index) =>
```
jeffkamo: Does this need the spread operator? Can't you just do `cookieList.map()`?

web/app/containers/privacy-page/parser.js
```diff
@@ -0,0 +1,9 @@
+import {getTextFrom, getTextArrayFrom} from '../../utils/parser-utils'
+
+export const privacyPageParser = ($, $html) => {
+    return {
+        title: getTextFrom($html, 'h1.page-title'),
+        intro: getTextFrom($html, 'div.message.info'),
+        cookieList: getTextArrayFrom($, $html, 'th'),
+    }
+}
```
jeffkamo: This is fine, although you _can_ also write it this way using ES6 shortcuts:

```jsx
export const privacyPageParser = ($, $html) => ({
    title: getTextFrom($html, 'h1.page-title'),
    intro: getTextFrom($html, 'div.message.info'),
    cookieList: getTextArrayFrom($, $html, 'th'),
})
```

jeffkamo:
I also have some thoughts that we should consider discussing in the near future. Parsing: when do parse everything vs. parse as little as possible!

Some pages make sense for parsing because there is predictable content that we want. Pages like product description pages, product lists, registration pages, checkout, etc. work well in this regard.

However, this is not true for all pages. Some content types are unpredictable, and parsing them is an exercise in futility at worst, maintenance nightmares at best. Privacy page is one content type I would suggest is not the right fit for large amounts of parsing. Other examples include License/Legal pages, maybe blog posts, etc. etc.

I won't suggest that you make any changes to this PR's parsing, but instead just give you a heads up that we must discuss this topic at some point to ensure we aren't spreading dangerous practices ;)

jvoll:
Thanks @jeffkamo. I've added a topic around parsing to the next Web dev sync. I've almost certainly been spreading the "parse everything" message as I don't have a full understanding myself of what strategies we should use.

On a somewhat meta note... wonder if we should stop using the privacy page for this exercise, though, the purpose is mostly to write things up and less for them to "parse everything". Maybe we just need to call it out with a caveat. Hrrrrrmmmmm....

web/app/containers/privacy-page/container.jsx
rsexton404: Cool. We'll also need to remove this from our Best Practices Guide in the docs.

jeremywiebe:
✅  Lessons learned
✅  brains expanded
✅  knowledge acquired

Closing.

PR #495, opened 2017-04-05
Re-enable the service worker since it’s become preview-aware

kschmidtdev:
Also restore control of preview toggling to the PreviewController

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1010
 **Linked PRs**: #304

## Changes
- Remove disable service worker in Astro check
- Restore control of preview toggling to the PreviewController

## How to test-drive this PR
- Run `npm run dev`
- Enable port forwarding to your Android device with chrome://inspect (localhost 8443)
- Run the Android app
- Shake to enable preview prompt
- Enable preview
- Kill app
- Make some local changes to your bundle you can easily verify
- Run app again
- Observe Preview display comes up
- Observe via Chrome Debugging tools a service worker exists
- Preview
- Observe your local changes are visible

### DIFF
--- a/native/app/app.js
+++ b/native/app/app.js
@@ -66,7 +66,7 @@ window.run = async function() {
             previewController.presentPreviewAlert()
         })
 
-        const previewEnabled = baseConfig.previewEnabled
+        const previewEnabled = await previewController.isPreviewEnabled()
         if (previewEnabled) {
             runAppPreview()
         } else {
--- a/native/app/config/baseConfig.js
+++ b/native/app/config/baseConfig.js
@@ -9,16 +9,13 @@ const colors = {
     whiteColor: '#ffffff'
 }
 
-const previewEnabled = true
-
 const baseConfig = {
     baseURL: 'https://www.merlinspotions.com',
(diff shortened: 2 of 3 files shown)


native/app/app.js
```diff
@@ -66,7 +66,7 @@ window.run = async function() {
             previewController.presentPreviewAlert()
         })
 
-        const previewEnabled = baseConfig.previewEnabled
```
kerrmarin: Should we remove the `baseConfig.js` `previewEnabled` property?

kschmidtdev: Good catch!

jeremywiebe: Are we sure though?  Isn't this partly so that we can have a build-time flag to control whether preview is even available?  For app store builds we want to make sure Preview can't be invoked at all.

kerrmarin: Yea, but that's what the `previewDisabledRelease` astro build variant is for, right?

kschmidtdev: Yes, this code isn't executed at all when you aren't using a 'preview-enabled' build variant

PR #498, opened 2017-04-05
[Integration Manager] Build in more of the Login component

DanielBrookRoberge:
Subsequent to discussions about hard-coding versus parsing content, this PR modifies the Login container to explicitly specify the form fields, similar to the various checkout forms. The increased simplicity this brought to the components allowed substantial code removal.

 **JIRA**: N/A

## Changes
- Build in the shape of the signin and register forms
- No longer parse that content from the HTML in the Merlin's connector
- Convert the ui -> login branch of the store to a simple collection of flags
- Move the signin vs register page determination from the connector to the component
- General minor refactors of the code

## How to test-drive this PR
- Run the tests
- Run `npm run dev`
- See that the login page loads and runs as before (but faster-looking!)
- Do the same with the Demandware site.


web/app/containers/login/partials/common.jsx
```diff
@@ -124,32 +28,41 @@ LoginFieldLabel.propTypes = {
...
-    tooltip: PropTypes.oneOfType([
-        PropTypes.object,
-        PropTypes.bool
-    ])
+    tooltip: PropTypes.node
 }
+
+const RememberMeTooltipContent = ({openModal}) => (
+    <a href="#remember-me" onClick={openModal}>
+        {'What\'s this?'}
```
jeffkamo: Why not just pass this as a normal HTML child rather than a JS string?

web/app/containers/login/partials/register-panel.jsx
```diff
@@ -1,43 +1,24 @@
...
-            <PanelDescription description={description} />
+            <h3 className="u-margin-bottom u-color-brand u-text-font-family u-text-normal">
+                New Customers
+            </h3>
+            <p>
+                {'Creating an account has many benefits: check out faster, keep more than one address, track orders, and more'}
+            </p>
             <div className="u-margin-top">
-                <PanelRequiredText requiredText={requiredText} />
+                {'* Required Fields'}
```
jnurse: Same as Jeff's comment above. Is there a reason why the text here and just above in the `<p>` tag are JS strings and not just plain text?

web/app/containers/login/partials/common.jsx
DanielBrookRoberge: Mainly because the raw, unpaired `'` character confuses things like Github syntax highlighting and my editor.

web/app/containers/login/partials/register-panel.jsx
DanielBrookRoberge: I was worried about special characters causing problems. That's probably not a real issue.

web/app/containers/login/partials/common.jsx
jeffkamo: Aha, fair enough!

jeffkamo:
@jnurse there was talk during Q2 OKR planning about addressing i18n and whatnot, but not sure if it got descoped or not. CC @henryh15

DanielBrookRoberge:
Henry's busy in a meeting right now; I'll wait to get confirmation on the internationalization stuff before merging.

henryh15:
Yeah we've had to cut a bunch of things on the FE team due to team constraints and unfortunately internationalization was one of the de-scoped items.

PR #499, opened 2017-04-05
Use relative paths for components in the READMEs

DanielBrookRoberge:
The component README files were generated with incorrect import statements for the Javascript of the component. This fixes them so that no tool thinks we're missing a dependency called `components`.

 **JIRA**:  N/A


DanielBrookRoberge:
Definitely typoed the sed command there. Fixed!

PR #501, opened 2017-04-06
Add Nightwatch Duncan page count test

fractaltheory:
We use a new way of incrementing the page count in a PWA, so let's test that it works properly.

 **Linked PRs**: https://github.com/mobify/platform-scaffold/pull/487

## Changes
- Add `Webpush` page object
- Hook up page visit counting to the guest checkout steps
- Fix a product detail page selector for 'item added' modal

## How to test-drive this PR
- Run `npm run dev`
- Run `npm run test:e2e`
- See that tests pass


web/tests/system/page-objects/webpush.js
```diff
@@ -0,0 +1,35 @@
...
+                        done()
+                    })
+                }
+            })
+            // eslint-enable prefer-arrow-callback
+
+    return this
+}
+
+Webpush.prototype.checkVisitCount = function(count) {
```
jeremywiebe: Do we have a convention around naming for this type of method?  I'd expect a method that asserts a state to be `assertVisitCount`.

web/tests/system/page-objects/webpush.js
```diff
@@ -0,0 +1,35 @@
...
+    this.browser
+        .getCookie(activeVisitCookie, function callback(result) {
+            this.assert.equal(result.value, count)
```
jeremywiebe: Is `this` correct at this point? Why not use an arrow function here?

web/tests/system/page-objects/webpush.js
```diff
@@ -0,0 +1,35 @@
...
+
+const Webpush = function(browser) {
+    this.browser = browser
+}
+
+Webpush.prototype.setup = function() {
+    this.browser
+        .waitForElementPresent(loaderScript)
+            // eslint-disable prefer-arrow-callback
+            .executeAsync(function callback(done) {
```
jeremywiebe: Any reason we can't use an arrow function here?

fractaltheory: Nope, fixed: https://github.com/mobify/platform-scaffold/pull/501/commits/35db3707bc5b5eb8e26185b3408f4c53cd1c6d8c

web/tests/system/page-objects/webpush.js
fractaltheory: `this` is correct if we use a function expression rather than an arrow function.

web/tests/system/page-objects/webpush.js
fractaltheory: Changed: https://github.com/mobify/platform-scaffold/pull/501/commits/35db3707bc5b5eb8e26185b3408f4c53cd1c6d8c

fractaltheory:
@jeremywiebe thanks for the feedback, I believe it's all been addressed now and this is ready for a re-review.

jeremywiebe:
Talked to @fractaltheory  Merging before this gets stale again. :)

PR #502, opened 2017-04-06
WEB-1290: banner showing on every step of checkout

gsaray:
"Welcome Back" banner when going through checkout on Merlin's with a previously registered email showing up on every step of checkout

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1290

## Changes
- call removeAllNotifications when route changes

## How to test-drive this PR
- Run `npm run dev`
- Add an item to cart
- go to shipping and enter "mobifyqa@gmail.co" for email 
- continue to payment

### DIFF
--- a/web/app/template.jsx
+++ b/web/app/template.jsx
@@ -26,7 +26,7 @@ const template = (WrappedComponent) => {
             if (!route.suppressFetch) {
                 dispatch(fetchPage(url, WrappedComponent, route.routeName, route.fetchUrl))
             }
-            this.props.dispatch(removeAllNotifications())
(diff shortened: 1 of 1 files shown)


web/app/template.jsx
```diff
@@ -26,6 +26,7 @@ const template = (WrappedComponent) => {
             if (!route.suppressFetch) {
                 dispatch(fetchPage(url, WrappedComponent, route.routeName, route.fetchUrl))
             }
+            this.props.dispatch(removeAllNotifications())
```
mlworks: I think we should be able to just do `dispatch(removeAllNotifications())`

PR #503, opened 2017-04-06
[Integration manager] demandware cart commands

jnurse:
This PR adds commands for the cart page to the demandware connector, it also adds product images to the cart data in the store. 

**Note:** This PR does not include work for the wishlist "save for later" functionality because it requires setting up the ability to transfer a cart on login, so I'll open another PR for that functionality. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1271
 **Linked PRs**: n/a

## Changes
- Added command to remove item from the cart
- Added command to update an item quantity in the cart
- Added command to "fetch" the page content
- Updated the getCart command to also fetch product images (only fetches an image if it's not in the store) 

## How to test-drive this PR
- Run `npm run dev`
- Preview [the demandware site](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Add an item to your cart
- navigate to the cart page and confirm the page is working as expected


web/app/containers/product-details/selectors.js
```diff
@@ -22,7 +22,7 @@ export const getSelectedProductDetails = createGetSelector(
     Immutable.Map()
 )
 
-export const getProductDetailsByPathKey = (pathKey) => createGetSelector(getProductDetails, pathKey, Immutable.Map())
+export const getProductDetailsByPathKey = (pathKey) => createGetSelector(getProducts, pathKey, Immutable.Map())
```
DanielBrookRoberge: These parametric selectors don't really benefit from using Reselect, since we're building a new selector function every time we use it. So it's probably more lightweight to just do 
```js
export const getProductDetailsByPathKey = (pathKey, state) => 
    getProducts(state).get(pathKey, Immutable.Map())
```

Either way, we should probably note somewhere that this should only be used in actions and not in components, since it would break all of the performance optimizations that exist there.

web/app/containers/cart/partials/cart-estimate-shipping.jsx
```diff
@@ -49,7 +49,7 @@ export const CartEstimateShippingModal = ({closeModal, isOpen, countries, stateP
                         :
                             <ReduxForm.Field component={Field} className="pw--has-select" name="regionId" label="State/Province">
                                 <select>
-                                    {stateProvinces.map(({label, value}) => <option value={value} key={value}>{label}</option>)}
+                                    {stateProvinces.map(({label, value}) => <option value={value} key={label}>{label}</option>)}
```
jeremywiebe: For Salesforce, at least, this doesn't look like it works exactly. The `value` field in the list of states for the US contains contains duplicates which causes React to complain. 

screen shot 2017-04-11 at 3 25 26 pm

web/app/containers/product-details/selectors.js
jeremywiebe: @DanielBrookRoberge Do you think there's a huge performance loss if we just use Reselect for everything that looks like a selector? Thinking if we can get away with it we just stay consistent despite it possibly being slightly slower.

web/app/containers/product-details/selectors.js
```diff
@@ -65,4 +65,6 @@ export const getFirstProductCarouselItem = createGetSelector(
 )
 export const getFirstProductImage = createGetSelector(getFirstProductCarouselItem, 'img')
 
-export const getUenc = (pathKey) => createGetSelector(getProductDetailsByPathKey(pathKey), 'uenc')
+export const getProductCarouselItemsByPathKey = (pathKey) => createGetSelector(getProductDetailsByPathKey(pathKey), 'carouselItems', Immutable.List())
+export const getFirstProductCarouselItemByPathKey = (pathKey) => createGetSelector(getProductCarouselItemsByPathKey(pathKey), 0, Immutable.Map())
```
jeremywiebe: I wonder if these should be `getDefaultXyz` instead of `getFirstXyz` to allow for connectors that may not use the first item by default. ?  YAGNI?

web/app/integration-manager/_demandware-connector/cart/commands.js
```diff
@@ -24,17 +27,100 @@ export const createBasket = () => {
         /* eslint-enable camelcase */
 }
 
+export const getProductImage = (item, currentState) => {
```
jeremywiebe: Do we need to export this?  It doesn't look used outside of this file?

web/app/integration-manager/_demandware-connector/cart/commands.js
```diff
@@ -24,17 +27,100 @@ export const createBasket = () => {
...
+
+export const removeFromCart = (itemId) => (dispatch, getState) => {
+    return createBasket()
+        .then((basketID) => {
+            return makeDemandwareRequest(`${API_END_POINT_URL}/baskets/${basketID}/items/${itemId}`, {method: 'DELETE'})
+                .then((response) => {
+                    if (response.ok) {
+                        return response.json()
+                    }
+                    throw new Error('Unable to remove item')
```
jeremywiebe: Can we add the `itemId` to this error message and log the response as a `console.warn` here? Or do we have conventions about what to log?

web/app/integration-manager/_demandware-connector/cart/commands.js
```diff
@@ -24,17 +27,100 @@ export const createBasket = () => {
...
+                body: JSON.stringify({
+                    quantity: itemQuantity
                 })
+            }
+            return makeDemandwareRequest(`${API_END_POINT_URL}/baskets/${basketID}/items/${itemId}`, requestOptions)
+            .then((response) => {
+                if (response.ok) {
+                    return response.json()
+                }
+                throw new Error('Unable to update item')
```
jeremywiebe: Same comment about error message and logging as above.

web/app/containers/cart/partials/cart-estimate-shipping.jsx
jeremywiebe: Update. I didn't realize these were constants we provided. I think the solution is just to collapse these Armed Forces entries as follows:
 
screen shot 2017-04-11 at 3 53 58 pm

web/app/containers/product-details/selectors.js
DanielBrookRoberge: The non `ByPathKey` versions of these are going away in the product schema refactor. So probably we can leave it and change it when we land the schema stuff.

web/app/containers/cart/partials/cart-estimate-shipping.jsx
jnurse: I pulled those values/labels from the state dropdown on the desktop instance. I can update it to follow your suggestion, but I was thinking it made more sense to follow desktop's convention since there's probably a reason behind how they've split them up.

web/app/containers/product-details/selectors.js
DanielBrookRoberge: I've realized since writing this comment that these components can't be used badly inside `createPropsSelector` for component props without a bunch of other things going on. So I'm not especially opposed to this anymore; we might hold on to a bit more memory than otherwise but we don't want to prematurely optimize.

web/app/integration-manager/_demandware-connector/cart/commands.js
jnurse: I'm not sure we have any conventions, at least not that I know of. Right now this error ends up being shown as a notification to the user, so if we leave it in a notification we might not want to add the itemId here, since it will be gibberish to the user.

web/app/integration-manager/_demandware-connector/cart/commands.js
jnurse: Good catch 😅

web/app/containers/product-details/selectors.js
jnurse: Added a comment 😄

web/app/integration-manager/_demandware-connector/cart/commands.js
jeremywiebe: Maybe just `console.error()` this then with the `response`?  Just thinking about how you'd track down problems adding/removing to cart.

jeremywiebe: Aaaaand, Jen just pointed out that the info would all be in the Network tab so logging is a bit redundant.  TIL.  :)

PR #506, opened 2017-04-07
Less step of the release process

jeffkamo:
Less step of the release process


jeffkamo:
missed merging this before changes happened in develop. Closing, and doing a new release!

PR #509, opened 2017-04-07
WEB-1328: Order summary always showing as $0.00

gsaray:
Order summary always showing as $0.00

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1328

## Changes
- check key to subtotal

## How to test-drive this PR
- Run `npm run dev`
- Add an item and go to cart


web/app/utils/utils.js
```diff
@@ -146,3 +146,25 @@ export const requestIdleCallback = (fn) => {
         return setTimeout(() => fn(), 1)
     }
 }
+
+
+export const getLocationData = (formValues, registeredFieldNames) => {
```
jeremywiebe: In Mobify's PWA's `getXyz` usually refers to a redux-select selector.  This looks more like a parser, so maybe `parseLocationData()`?

web/app/utils/utils.js
```diff
@@ -146,3 +146,25 @@ export const requestIdleCallback = (fn) => {
...
+        const getRegisteredFieldValue = (fieldName) => {
+            return registeredFieldNames.includes(fieldName) ? formValues[fieldName] : undefined
+        }
+        address.country_id = getRegisteredFieldValue('country_id')
+        address.region_id = getRegisteredFieldValue('region_id')
+        address.postcode = getRegisteredFieldValue('postcode')
+        if (formValues.region) {
+            address.region = getRegisteredFieldValue('region')
+            // Remove the region_id in case we have an old value
+            delete address.region_id
```
jeremywiebe: I don't understand this.  On line 161 we assign this but then we remove it right away. Does `region` make `region_id` unnecessary?  

If so, could we maybe put the assignment of `region_id` in an `else` block instead of assigning and then `delete`ing it?

jnurse: Yes `region` makes `region_id` unnecessary, but we need to delete the key because even if we don't assign a new value to `region_id` on line 161, the object contains a `region_id` by default (see line 153). We still could move the `region_id` assignment to an `else` block, but we'll need to leave the line that deletes it.

PR #511, opened 2017-04-10
No longer parse plp text

DanielBrookRoberge:
The text on the PLP when there were no available products was parsed out of the PLP HTML. This PR removes that in favour of a built-in string. 

 **JIRA**: N/A

## Changes
- Use a string in the code rather than a parsed string when there are no items
- No longer pick out the string in the Merlin's parser.

## How to test-drive this PR
- Run `npm run dev`
- Go to an empty category (e.g. Charms)
- See that there is still text
- See that the text is not stored in the categories branch of the Redux store.


DanielBrookRoberge:
Weird, it was working last week. I'll look into it.

PR #515, opened 2017-04-10
WEB-1230: Implement log out functionality

mlworks:
**JIRA**: https://mobify.atlassian.net/browse/WEB-1230

## Changes
- Update local nav-item component to accept custom `onClick` behaviour
- Add `signOut` action which will display notification, update cart items and update the navigation menu

## How to test-drive this PR
- Run `npm run dev`
- Log into an account
- Navigate to Potions PLP
- Open menu and click on sign out
- Reopen menu and you should see sign in link again while Potions item is still highlighted


web/app/containers/app/actions.js
```diff
@@ -207,3 +207,30 @@ export const fetchSvgSprite = () => {
...
+                dispatch(addNotification({
+                    content: 'You are now signed out',
+                    id: 'signedOutNotification'
+                }))
+                dispatch(cartActions.getCart())
+
+                // Update navigation menu
+                // Need to request current location so when we are on Potions PLP
+                // the potions nav item will render as 'active'
+                makeRequest(window.location.href)
```
jnurse: Is this section needed? The correct nav item is showing for me without this piece.

mlworks: Oh yea I have to parse the navigation again because without this section, the sign out button doesn't get updated to sign in button.

web/app/components/nav-item/index.jsx
```diff
@@ -27,6 +28,16 @@ NavItemWithIcon.propTypes = NavItem.propTypes
 
 
 /**
+ * Project-specific NavItem which displays an icon on the left.
+ */
+export const NavItemWithOnClick = (props) => (
+    <ListTile {...props} children={props.title} className="c-nav-item c--with-icon" startAction={<NavItemIcon name="user" />} />
```
jnurse: Can you remove this children prop, and make {props.title} an actual child of the component?

web/app/containers/app/actions.js
jnurse: Fair enough. Do you think we'll ever need to chain this sign out action with other actions, and execute those other actions after this request has finished? If that's the case can you return this makeRequest call?

PR #517, opened 2017-04-11
Add support for Samsung native browser

cdegit:
## Changes
- Add babel polyfill
- Fix a bug that was breaking the PDP on Samsung

## How to test-drive this PR
- Run `npm i`
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1) in the Samsung native browser
- smoke test the site and make sure everything works

## Note
Checkout appears to be broken on develop and this branch. If you refresh the page while you're at `https://www.merlinspotions.com/checkout/payment/`, `extractMagentoShippingStepData` will fail


web/app/containers/product-details/partials/product-nearest-stores.jsx
```diff
@@ -40,7 +40,7 @@ class ProductNearestStores extends React.Component {
         })
 
         const merlinsPotionsScript = document.createElement('script')
-        merlinsPotionsScript.append(merlinsPotionsAsync)
+        merlinsPotionsScript.innerHTML = merlinsPotionsAsync
```
jansepar: Is this compatible cross browser? I'd double check Safari.

web/webpack/base.main.js
```diff
@@ -13,6 +13,7 @@ const analyzeBundle = process.env.MOBIFY_ANALYZE === 'true'
 const config = {
     devtool: 'cheap-source-map',
     entry: [
+        'babel-polyfill',
```
jansepar: Does this have a bundle size impact?

drewzboto:
If we do merge this we also need:
- Updated to support matrix: https://docs.mobify.com/platform/1.0/
- Update GS to let them know their contracts will change
- A strategy on how to keep regression testing on this OS/Browser combo

web/app/containers/product-details/partials/product-nearest-stores.jsx
cdegit: Yup, this works in Safari.

web/webpack/base.main.js
cdegit: Yes. Babel-polyfill itself isn't that large, but it pulls in more from core-js which is unfortunately pretty big:

Before:
screen shot 2017-04-11 at 2 10 49 pm
Vendor.js: 147.25 KB gzipped

After: 
screen shot 2017-04-11 at 2 08 02 pm
Vendor.js: 173.18 KB gzipped

jansepar:
So, yeah, I'm definitely a -1 on this change if it makes the build 26KB larger. Is there any alternative way of resolving this?

jansepar:
We need a test in CI to check the build size and fail if it's over a certain limit :)

DanielBrookRoberge:
We could make yet another bundle file with polyfills; the loader should be able to tell whether one is needed and conditionally load the polyfill bundle.

This could also include the `es6-promise` polyfill, which isn't needed very many places anymore.

web/app/containers/product-details/partials/product-nearest-stores.jsx
jeremywiebe: We have to be careful when we say "Safari" because we also have to take into account iOS and Android embedded web views back to the supported levels we have for Astro (which right now is Android 4.4 and iOS 8 still, I think).

cdegit:
@jansepar @DanielBrookRoberge I've split the polyfills (core.js and whatwg-fetch) out into their own chunk. I removed the polyfill coming from the SDK, as the core.js shim covers it. 

I've also moved the SDK into its own chunk, as a couple of components were being imported multiple times within different chunks.

screen shot 2017-04-18 at 5 18 21 pm

Vendor: 143.17 KB gzipped

DanielBrookRoberge:
Does the core.js shim include Promises? This looks pretty good if that's the case.

cdegit:
Hmm, actually it looks like this isn't applying the polyfills early enough. 🤔

cdegit:
Okay, looks like the polyfill is being applied at the right time now!

@DanielBrookRoberge Yeah, the core.js shim includes promises.

jeremywiebe:
@cdegit I wonder if we have to avoid use of `import()` in the polyfill for the exact reason you said (Promise being one of the things we potentially need to polyfill).  

[This article](https://philipwalton.com/articles/loading-polyfills-only-when-needed/) points to a more manual process using `<script>` tags, which got me thinking... what if we moved this polyfilling stuff into the `loader.js` instead of `main.jsx`? Then we could use the `loadScript` function that already does this (we'd possibly need to add a parameter to allow us to make it `sync` instead of `async`).  

@jansepar Can you see any issues with that?

cdegit:
Yeah, I'm not 100% sure how transpiling those imports works. I'll give that a try and see how it works.

kschmidtdev:
If this ends up polyfilling Promise then it will help out older Android WebViews for Apps! /cc @henryh15 @kerrmarin

cdegit:
@jeremywiebe Okay, we're no longer polyfilling promises with promises 😄

web/app/loader.js
```diff
@@ -60,15 +60,17 @@ if (isReactRoute()) {
...
+    const loadScript = ({id, src, onload, async = true}) => {
         const script = document.createElement('script')
 
         // Setting UTF-8 as our encoding ensures that certain strings (i.e.
         // Japanese text) are not improperly converted to something else. We
         // do this on the vendor scripts also just in case any libs we
         // import have localized strings in them.
         script.charset = 'utf-8'
-        script.async = true
+        if (async) {
```
jeremywiebe: Could this not be shortened to just this?
```js
script.async = async
```

web/webpack/base.loader.js
```diff
@@ -12,10 +12,14 @@ const readNativeAstroVersion = () => {
 
 module.exports = {
     devtool: 'cheap-source-map',
-    entry: './app/loader.js',
+    entry: {
+        loader: './app/loader.js',
+        'core-polyfill': 'core-js',
```
jeremywiebe: One last thing... I wonder if we should bundle `core-js` and `whatwg-fetch` together.  Basically if you need _any_ polyfills, you have to load them all. The upside is that you only have a single request to get all polyfills needed.  Thoughts?

cdegit: I thought about that, but `whatwg-fetch` is needed by a lot of browsers, and `core-js` is needed by fewer. Given how much larger the `core-js` polyfill is, I didn't really want to load it on browsers that didn't need it.

web/app/loader.js
cdegit: I believe `async` is one of those attributes where if its set at all, it's assumed to be true regardless of the value.

Good call, I'll change that.

cdegit: Nevermind, looks like `script.async = isAsync` works!

web/app/loader.js
```diff
@@ -146,6 +146,23 @@ if (isReactRoute()) {
...
+            loadScript({
+                id: 'progressive-web-core-polyfill',
+                src: getAssetUrl('core-polyfill.js'),
+                isAsync: false
+            })
+        }
+
+        if (!global.fetch) {
+            loadScript({
+                id: 'progressive-web-core-polyfill',
```
jeremywiebe: This id is identical to the previous one.  Is that on purpose?

cdegit: Nope! Good catch

jansepar:
Ah, just looked at the changes, looks like we're downloading the polyfills on the fly, awesome! You've probably already thought of this but just to confirm, if main loaded before the polyfills, would this stilll work?

cdegit:
The polyfills are now being loaded synchronously before main.js, so I don't believe we should run into that problem.

web/app/loader.js
```diff
@@ -60,15 +60,15 @@ if (isReactRoute()) {
...
+    const loadScript = ({id, src, onload, isAsync = true}) => {
         const script = document.createElement('script')
 
         // Setting UTF-8 as our encoding ensures that certain strings (i.e.
         // Japanese text) are not improperly converted to something else. We
         // do this on the vendor scripts also just in case any libs we
         // import have localized strings in them.
         script.charset = 'utf-8'
-        script.async = true
+        script.async = isAsync
```
jansepar: So, interesting thing is that this flag is a bit of a lie. You can't dynamically load scripts sychronously like this. The only option would be to document.write them, but because our script is loaded async from the tag, `doc.writing` into the page is dangerous.

cdegit: @jansepar I've refactored this to instead wait for the polyfills to load before initializing the app.

web/app/loader.js
```diff
@@ -16,159 +22,156 @@ const isReactRoute = () => {
...
+        loadAsset('meta', {
+            name: 'theme-color',
+            content: '#4e439b'
+        });
+
+        // load the worker if available
+        // if no worker is available, we have to assume that promises might not be either.
+        (('serviceWorker' in navigator)
+         ? loadWorker()
+         : {then: (fn) => setTimeout(fn)}
```
benlast: You could now replace this line with `: Promise.resolve()`...

web/app/loader.js
```diff
@@ -16,159 +22,156 @@ const isReactRoute = () => {
...
+                })
+            }
+        }, 150)
+    }
+
+    hasInitialized = true
+}
+
+// Apply polyfills
+const neededPolyfills = getNeededPolyfills()
```
benlast: Could this block maybe just be `getNeededPolyfills().forEach((polyfill) => polyfill.load(attemptToInitializeApp))`?

jeremywiebe: @benlast We still need the `else` path for when there are no polyfills as there won't be a load callback to call `attemptToInitializeApp` otherwise.

web/app/loader.js
```diff
@@ -3,172 +3,171 @@ import {getAssetUrl, loadAsset, initCacheManifest} from 'progressive-web-sdk/dis
 import {displayPreloader} from 'progressive-web-sdk/dist/preloader'
 import cacheHashManifest from '../tmp/loader-cache-hash-manifest.json'
 import {isRunningInAstro} from './utils/astro-integration'
-
-window.Progressive = {
-    AstroPromise: Promise.resolve({})
-}
-
+import {loadScript} from './utils/utils'
```
DanielBrookRoberge: This should go in a separate `loader-utils.js` file, so that we don't pull in more code than we have to in the loader.

web/app/utils/loader-utils.js
```diff
@@ -0,0 +1,17 @@
+export const loadScript = ({id, src, onload, isAsync = true}) => {
```
jansepar: Nice change - I'd actually much prefer if this lived in the SDK but wouldn't consider it a blocker since that wasn't exactly the point of this PR :)

web/app/utils/polyfills.js
```diff
@@ -0,0 +1,30 @@
+import {getAssetUrl} from 'progressive-web-sdk/dist/asset-utils'
+import {loadScript} from './loader-utils'
+
```
jansepar: Since this *is* the point of this PR - how do you feel about this moving to the SDK?

jansepar: Actually just looking at this file more... This doesn't really feel like a "utility". It's quite hard coded specifically for our build. I do wonder then if it should be in the utils folder at all...

cdegit: Yeah, I expect this file will end up being pretty build specific. For a build that needs i18n for example, you could add that polyfill here. I was thinking it could also live in just the /app folder, but I didn't want to clutter it up too much

PR #519, opened 2017-04-11
[Integration Manager] Fix interaction between Loadable and Template HoCs.

DanielBrookRoberge:
By separating the Template HoC from the actual template components, the React Loadable work made the `fetcher` functions not work. This PR rearranges things so that they will work again. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-56

## Changes
- Wrap the template components in their own `container.js`, as long as they've been converted to the Integration Manager from the old `fetchPage` action. (the IM changes broke the circular dependency)
- Write a wrapper component for Loadable to simplify `templates.jsx` and the lazy-load of the templates in the App component.

## How to test-drive this PR
- Run `npm run dev`
- See that all of the non-Home templates work again.

### DIFF
--- a/web/app/containers/cart/container.js
+++ b/web/app/containers/cart/container.js
@@ -1,5 +1,6 @@
 import React, {PropTypes} from 'react'
 import {connect} from 'react-redux'
+import template from '../../template'
 import {createPropsSelector} from 'reselect-immutable-helpers'
 import {getAssetUrl} from 'progressive-web-sdk/dist/asset-utils'
 import classNames from 'classnames'
@@ -117,4 +118,4 @@ const mapStateToProps = createPropsSelector({
     hasItems: getCartHasItems
 })
 
-export default connect(mapStateToProps)(Cart)
+export default template(connect(mapStateToProps)(Cart))
--- a/web/app/containers/login/container.test.js
+++ b/web/app/containers/login/container.test.js
@@ -6,7 +6,7 @@ import React from 'react'
 import * as AstroIntegration from '../../utils/astro-integration'
 
 import ConnectedLogin from './container'
-const Login = ConnectedLogin.WrappedComponent
+const Login = ConnectedLogin.WrappedComponent.WrappedComponent
 
 describe('The Login', () => {
     const originalIsRunningInAstro = AstroIntegration.isRunningInAstro
--- a/web/app/router.jsx
+++ b/web/app/router.jsx
@@ -6,6 +6,7 @@ import {Provider} from 'react-redux'
 
 // Containers
 import App from './containers/app/container'
+// These templates are code-split out of the main bundle.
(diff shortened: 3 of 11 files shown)


web/app/containers/login/container.test.js
```diff
@@ -6,7 +6,7 @@ import React from 'react'
 import * as AstroIntegration from '../../utils/astro-integration'
 
 import ConnectedLogin from './container'
-const Login = ConnectedLogin.WrappedComponent
+const Login = ConnectedLogin.WrappedComponent.WrappedComponent
```
jeremywiebe: Ewwww  :)

web/app/router.jsx
```diff
@@ -6,7 +6,9 @@ import {Provider} from 'react-redux'
 
 // Containers
 import App from './containers/app/container'
-import {Cart, CheckoutConfirmation, CheckoutPayment, CheckoutShipping, Home, Login, ProductList, ProductDetails} from './containers/templates'
+import {Cart, CheckoutConfirmation, CheckoutPayment, CheckoutShipping, Login, ProductList, ProductDetails} from './containers/templates'
```
jeffkamo: Should we call out this line explicitly as code splitted templates? Might help clarify the comment below about the homepage, etc.

DanielBrookRoberge: Yeah sounds good. Longer term I'm thinking of renaming `containers/templates.js` for something more explicitly about code splitting (since it won't be doing template wrapping in the Integration Manager code when it's done)

web/app/containers/login/container.test.js
DanielBrookRoberge: Yeah it's not the nicest. Oh well, we get good things for the price of this.

web/app/containers/cart/container.js
```diff
@@ -117,4 +118,4 @@ const mapStateToProps = createPropsSelector({
     hasItems: getCartHasItems
 })
 
-export default connect(mapStateToProps)(Cart)
+export default template(connect(mapStateToProps)(Cart))
```
jeremywiebe: Random thought/suggestion.  What if we introduced `index.js[x]` files in these container folders and did the wrapping in there?  That way our imports for the wrapped components could just be `import container from 'container/cart'` and for tests we could do `import container from 'container/cart/container'`.  

I'm concerned about the amount of wrapping we're doing before exporting the actual component.  The `.UnwrappedComponent` bits feel like they could become hard to maintain if we always need to remember how wrapped a component is.  If we introduce the above pattern then if we import the `index.js` file we treat it as a black box, but if we want to test, we import the `container.js` file and know it's not wrapped.

jeremywiebe: Discussed IRL... looks like this won't _quite_ work as then `mapDispatchToProps` and `mapStateToProps` would have to be in the `index.js`.  Let's keep thinking about some solution for this in a future PR.

PR #525, opened 2017-04-11
Rename containers to .JSX file extension

jvoll:
**JIRA**: small part of https://mobify.atlassian.net/browse/WEB-1224
 **Linked PRs**: none

## Changes
- Rename all files that contain JSX to have `.jsx` file extension
- Necessary config changes to make webpack happy as a result of this change
- Fix an unrelated bug in PLP where we don't detect the "no products for category" situation correctly and instead just load indefinitely

## How to test-drive this PR
- Run `npm run dev`
- Verify everything works as per usual
- Go to a PLP with no products in the category such as "new arrivals". Previously this would have showed skeletons forever, now it should show the no results message.


web/webpack/base.onboarding.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
     module: {
         rules: [
             {
-                test: /\.js$/,
+                test: /\.jsx?$/,
```
jvoll: Sure, I'll make that change in the place I copied this from too 😉

PR #528, opened 2017-04-11
[Integration Manager] Product Schema Refactor

DanielBrookRoberge:
This PR refactors the app to follow the Product schema as discussed yesterday. It includes the typechecking work, and typechecks the product response payloads. 

This is a wide-ranging refactor, touching the commands and parsers for both PDP and PLP, the components and selectors for both PDP and PLP, and some of the PDP actions. In the process, I found some bugs that I fixed

Future work:
 - Move many product-related selectors into the `store/products` directory in preparation for moving to the SDK
- Stub out type checking in production

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-10

## Changes
- Implement Product schema typechecking with `runtypes`
- Modify connectors to return data that matches the schema
- Modify components and selectors to use the new data format
- Modify PDP actions to use the new selectors
- Modify the variation command to use the new data.
- Never store `null` in sessionStorage for the authorization value
- Fix validation of variations in the add to cart action

## How to test-drive this PR
- Preview on both Merlin's and Demandware connectors
- Check that the PDP and PLP render correctly
- Check that the variation form on the PDP works correctly
- Check that the add to cart button correctly validates the variation form.


web/app/integration-manager/products/types.js
```diff
@@ -0,0 +1,74 @@
...
+
+const VariationCategory = Runtypes.Record({
+    id: Identifier,
+    label: Text,
+    values: Runtypes.Array(Option)
+})
+
+const Variation = Runtypes.Record({
+    id: Identifier,
+    values: Runtypes.Dictionary(Identifier, Identifier)
```
jeremywiebe: Given that "typedef"s are so easy to create, could we further typedef `Identifier` to have things like `ProductIdentifier` and `VariationCategoryIdentifier` so that something like this dictionary is obvious what is a valid key and what's a valid value?

PR #532, opened 2017-04-13
Move product selectors out of the product details branch

DanielBrookRoberge:
This PR separates the selectors for the product information in the store from the selectors for the product details UI data. 

 **JIRA**: N/A (but found while working on https://mobify.atlassian.net/browse/WEBDATA-10)

## Changes
- Move product selectors to `app/store/products/selectors.js` so they aren't so tied to the ProductDetails container.

## How to test-drive this PR
- Preview and see that nothing has changed.

### DIFF
--- a/web/app/containers/product-details/actions.js
+++ b/web/app/containers/product-details/actions.js
@@ -3,8 +3,8 @@ import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
 import {SubmissionError} from 'redux-form'
 import {createPropsSelector} from 'reselect-immutable-helpers'
 
-import * as selectors from './selectors'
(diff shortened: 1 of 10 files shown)


web/app/containers/product-details/actions.js
```diff
@@ -5,6 +5,7 @@ import {createPropsSelector} from 'reselect-immutable-helpers'
 
 import * as selectors from './selectors'
```
jeremywiebe: What do you think about updating these imports also to avoid importing the entire namespace and just the named exports we use?

DanielBrookRoberge: I've definitely been doing that piecemeal as we work on this. I can definitely do that here too.

PR #534, opened 2017-04-17
[Integration Manager] Recover from stale session data

DanielBrookRoberge:
If a tab with the site is left open for long enough, the data in `sessionStorage` will no longer be recognized by the backend. This PR detects the error conditions and retries without the stale data. 

 **JIRA**: N/A (discovered during https://mobify.atlassian.net/browse/WEBDATA-10)

## Changes
- If the auth token refresh fails, start over with a new token
- If the cart is missing, create a new one

## How to test-drive this PR
- Leave a tab open with the Demandware Merlin's long enough for the token to get stale
- Reload and see that there are no auth errors in the console.


web/app/integration-manager/_demandware-connector/cart/utils.js
```diff
@@ -71,10 +71,13 @@ export const parseAndReceiveCartResponse = (responseJSON) => (dispatch, getState
 
 export const requestCartData = () => {
     return createBasket()
-        .then((basketID) => {
-            const options = {
-                method: 'GET'
+        .then((basketID) => makeDemandwareRequest(`${API_END_POINT_URL}/baskets/${basketID}`, {method: 'GET'}))
+        .then((response) => {
+            if (!response.ok) {
```
jeremywiebe: When we talked I think this is the spot we decided we should check the status code just to sure.

web/app/integration-manager/_demandware-connector/cart/utils.js
```diff
@@ -71,10 +71,13 @@ export const parseAndReceiveCartResponse = (responseJSON) => (dispatch, getState
...
+                // Our basket has expired, clear and start over
+                deleteBasketID()
+                return requestCartData()
```
jnurse: I agree, in fact I think I briefly tried setting up something similar at one point and got stuck in said loop. 😿

jeremywiebe: Might be a great place for the [Circuit Breaker](https://martinfowler.com/bliki/CircuitBreaker.html) pattern! :D

DanielBrookRoberge: I think I'll just switch to an explicit 404 check and add a `noRetry` flag for the second time through.

web/app/integration-manager/_demandware-connector/cart/utils.js
```diff
@@ -69,12 +69,18 @@ export const parseAndReceiveCartResponse = (responseJSON) => (dispatch, getState
         .then((basketData) => dispatch(receiveCartContents(basketData)))
 }
 
-export const requestCartData = () => {
+export const requestCartData = (noRetry) => {
```
jeremywiebe: The call sites for `requestCartData` don't seem to be passing any parameters when they call this function. Is this using some ES6 to default in some way?

DanielBrookRoberge: No, just old-school Javascript (since `undefined` is falsy)

web/app/integration-manager/_demandware-connector/cart/utils.js
```diff
@@ -69,12 +69,18 @@ export const parseAndReceiveCartResponse = (responseJSON) => (dispatch, getState
...
-                method: 'GET'
+        .then((basketID) => makeDemandwareRequest(`${API_END_POINT_URL}/baskets/${basketID}`, {method: 'GET'}))
+        .then((response) => {
+            if (response.status === 404) {
+                if (noRetry) {
+                    throw new Error('Cart not found')
+                }
+                // Our basket has expired, clear and start over
+                deleteBasketID()
+                return requestCartData()
```
jeremywiebe: I think we want `requestCartData(true)` here right?

DanielBrookRoberge: Yup! Will fix.

PR #536, opened 2017-04-17
[Integration Manager] Don't parse for the footer anymore

DanielBrookRoberge:
This PR builds in the navigation links for the footer. While doing this I noticed that the newsletter form data in the Redux store was unused so I removed that too. 

 **JIRA**: N/A

## Changes
- Build in the links for the footer
- Remove footer parser and corresponding actions, selectors, and reducer lines.
- No longer disable the newsletter form while waiting for the data.

## How to test-drive this PR
- Run `npm run dev`
- See that the footer is still full of links.


web/app/containers/footer/partials/footer-navigation.jsx
```diff
@@ -1,39 +1,33 @@
-import React, {PropTypes} from 'react'
-import {createPropsSelector} from 'reselect-immutable-helpers'
-import * as selectors from '../selectors'
-import {connect} from 'react-redux'
+import React from 'react'
```
jeremywiebe: `React` doesn't seem to be used in this file any longer.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
```diff
@@ -1,40 +1,32 @@
...
-    newsletter: selectors.getNewsletter
-})
-
 const mapDispatchToProps = {
     onSubmit: actions.signUpToNewsletter
 }
 
 export default connect(
-    mapStateToProps,
+    null,
```
jeremywiebe: What does this `null` represent?

web/app/containers/footer/partials/footer-navigation.jsx
DanielBrookRoberge: The `React` object needs to be in scope wherever there is JSX, since it transpiles to `React.createElement` calls. It's not automatic to be independent of the import mechanism in use.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
DanielBrookRoberge: The null means no `mapStateToProps` function; we need it since there's still a `mapDispatchToProps` parameter.

DanielBrookRoberge:
The submit action seems to be broken on the `integration-manager` branch, too; guess now's the time to fix it up

DanielBrookRoberge:
@jeremywiebe Submission is fixed.

web/app/containers/footer/partials/footer-navigation.jsx
jeremywiebe: TIL!  Thanks.

PR #537, opened 2017-04-17
[Integration Manager] Category Schema Refactor

DanielBrookRoberge:
This PR refactors the Category branch of the store to match the discussed schema, and cleans up the API a little for increased user friendliness. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-10

## Changes
- Add the schema and conform the two connectors to it
- Change the UI to expect the new schema
- Split the `receiveCategoryContents` action and corresponding types into `categoryInfo` and `categoryContents`.

## How to test-drive this PR
- Preview with both connectors.
- See that the PLP displays correctly, and that the transition to the PDP is still smooth.


web/app/integration-manager/_demandware-connector/categories/commands.js
```diff
@@ -1,11 +1,36 @@
...
+        parentId
+    }))
+
+    if (parentId) {
+        dispatch(fetchParentCategoryData(parentId))
+    }
+}
+/* eslint-enable camelcase, no-use-before-define */
+
+const fetchParentCategoryData = (id) => (dispatch) => {
```
DanielBrookRoberge: The parent part is that we don't fetch the contents, just the header information (so a `CategoryInfo` but not a `CategoryContents`)

web/app/store/categories/reducer.js
```diff
@@ -1,12 +1,14 @@
...
 import {handleActions} from 'redux-actions'
 import {mergePayload} from '../../utils/reducer-utils'
-import {receiveCategory} from '../../integration-manager/categories/responses'
+import {receiveCategory, receiveCategoryInformation, receiveCategoryContents} from '../../integration-manager/categories/responses'
 
 const initialState = Immutable.Map()
 
 const categoryReducer = handleActions({
-    [receiveCategory]: mergePayload
+    [receiveCategory]: mergePayload,
```
DanielBrookRoberge: Yeah you've got the gist of this. I realized while working on the Demandware connector that they have different lifecycles, so splitting them out like this not only helped with type checking but I think clarified some different concerns.

In particular, once we've got paginated results working we'll be getting more products which will change the CategoryContents but not the CategoryInfo.

web/app/integration-manager/_demandware-connector/categories/commands.js
jeremywiebe: Could we name it `fetchCategoryInfo` then to more tightly map between the method and what it's fetching?  `fetchCategoryInfo() -> CategoryInfo`

PR #539, opened 2017-04-18
WEB-1336: Add Filter

jeffkamo:
**JIRA**: [WEB-1336](https://mobify.atlassian.net/browse/WEB-1336)

## Changes
- Add filter functionality to the product list page

## Todo
- [x] Add the UI part of filters
- [x] Verify typography and spacing
- [x] Verify all requirements have been met from the Jira ticket
- [x] Write tests... #feelShameAsAMinorAction
- [x] Merge #543 
- [x] Fix merge conflicts
- [x] Finish applying feedback

## How to test-drive this PR
- Run `npm run dev`
- Visit a product list page like the [Potions page](https://www.merlinspotions.com/potions.html)
- Verify the Filter functionality works as intended
- Visit a page with the `?filter_is_open` query string present: https://www.merlinspotions.com/potions.html?filter_is_open
- Verify that the page opens with the filter modal open by default


web/app/store/categories/parsers/product-list.js
```diff
@@ -1,6 +1,40 @@
...
 import {urlToPathKey} from '../../../utils/utils'
 
+const priceParser = (value) => {
+    const values = value.split('-')
+    const floor = values[0] || 0
+    const ceiling = values[1] || Infinity
+    return {floor, ceiling}
+}
+
+const filtersParser = () => {
```
jnurse: Since we're not getting any of this data from the desktop site's HTML I wonder if it makes sense to have this in the parser file? How do you think the filters should behave if we moved this into the integration manager? Should it still work the same (or close to it)?

web/app/store/categories/parsers/product-list.js
```diff
@@ -1,6 +1,40 @@
 import {parseTextLink, getTextFrom} from '../../../utils/parser-utils'
 import {urlToPathKey} from '../../../utils/utils'
 
+const priceParser = (value) => {
```
jnurse: This function looks good, but I think might be unnecessary while we're hard coding all the content for filters. This feels like an example of [YAGNI](https://github.com/mobify/developer-values#you-aint-gonna-need-it-yagni) what do you think?

web/app/utils/filter-utils.js
```diff
@@ -0,0 +1,92 @@
...
+ *
+ *    const tokens = filters.reduce(toTokens, [])
+ *    // => [{ruleset: 'price', criteria: 'blue'}, ...]
+ *    //    list of tokenized filters
+ *
+ * @param {array} list - the list that has only active tokens appended to it.
+ * @param {object} currentFilter - the current filter object
+ * @returns {array} - The final list of filter tokens in the format of {ruleset, criteria}
+ */
+export const toTokens = (list, currentFilter) => {
```
jnurse: This function seems to be doing two things, filtering out the non-active filters and building out these tokens I wonder if we could instead create a selector that gets only active filters? Then this function would just build out the tokens? 

It might require a few changes to the structure of the filter object.

web/app/utils/filter-utils.js
```diff
@@ -0,0 +1,92 @@
+const NON_NUMERIC_REGEX = /\D/
+
+const RULESETS = {
+    price: (item, range) => {
+        const price = parseInt(item.price.replace(NON_NUMERIC_REGEX, ''))
```
jnurse: You might want to use parseFloat here instead, that way prices like $15.99 will correctly give you a float of 15.99 rather than an int of 15.

web/app/utils/filter-utils.js
```diff
@@ -0,0 +1,92 @@
+const NON_NUMERIC_REGEX = /\D/
```
jnurse: Technically this regex will just give you the first non-numeric character, you might want to adjust this so it will match all non-numeric characters except for the '.' character which you'll need to keep in place. Happy to help figure out what that regex would look like if you'd like.

web/app/utils/filter-utils.js
```diff
@@ -0,0 +1,92 @@
...
+ *
+ *    const items = [{}, {}, ...]
+ *    const tokens = getFilters.toJS().reduce(toTokens, [])
+ *    items.filter(byTokens(tokens)) => items matching token criteria
+ *
+ * @param  {array} tokenList - list of filter tokens (see toToken function
+           above). It's param `currentItem` is likely an object that will
+           eventually be consumed by a React component.
+ * @returns {function} - a callback function, intended for filter methods
+ */
```
jnurse: Nice function descriptions 👏

web/app/store/categories/parsers/product-list.js
```diff
@@ -1,6 +1,40 @@
...
+                    count: '2',
+                    criteria: priceParser('10-20'),
+                    label: '$10.00 - $19.99'
+                }, {
+                    // active: true,
+                    count: '2',
+                    criteria: priceParser('20-30'),
+                    label: '$20.00 - $29.99'
+                },
+            ]
```
jnurse: I wonder if this is the best structure for the filter data, or if there's a better way to keep track of all this data. Maybe that's something we can discuss when you get a chance.

web/app/store/categories/parsers/product-list.js
jeffkamo: I was going to eventually replace this hard coded example with actually parsing the desktop.

I imagine the format of the price will change from site to site, so I'm not sure if it makes sense to include it in the integration manager.

web/app/store/categories/parsers/product-list.js
jeffkamo: +1, definitely open to discussing this format more.

jeffkamo:
@mitchquik Sorting and Filtering have now been combined. Whenever you have a chance to give this a design pass, that would be rad 😄

web/app/containers/product-list/selectors.js
```diff
@@ -23,24 +24,43 @@ export const getProductListContentsLoaded = createHasSelector(
...
+        }))
+
+        return activeFilters
+    }
+)
+export const getFilteredProductListProducts = createSelector(
     getProductListProducts,
+    getActiveFilters,
+    (products, filters) => {
+        const filterTokens = filters.toJS().reduce(toTokens, [])
```
jnurse: If we only have active filters now, is it still necessary to have this reducer? We could just leave the filter objects as is and pass the extra data to `filter(byTokens(..))`

web/app/containers/product-list/partials/product-list-filter-modal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                                for the bubbled events from the children button elements */}
+                            {/* eslint-disable
+                                jsx-a11y/click-events-have-key-events,
+                                jsx-a11y/onclick-has-focus,
+                                jsx-a11y/onclick-has-role,
+                                jsx-a11y/no-static-element-interactions */}
+                            <div
+                                className="t-product-list__filter-modal-items"
+                                role="presentation"
+                                onClick={(e) => {
```
jeffkamo: Fair enough, that would probably be simpler. Is this event delegation premature optimization?

jnurse: Yeah I think so.

web/app/store/categories/actions.js
```diff
@@ -1,11 +1,26 @@
...
+export const changeFilter = createAction('Change Filter')
 export const changeSortOption = createAction('Change Sort Option')
 
+export const changeFilterTo = (filterQuery) => (dispatch, getStore) => {
+    const currentState = getStore()
+    const categoryData = getSelectedCategory(currentState).toJS()
+
+    categoryData.filters.forEach((filter) =>
+        filter.kinds.forEach((kind) => {
+            kind.active = kind.query === filterQuery
```
jeffkamo: Ya it's kind of awkward. Since with the current feature availability (only one filter can be active) there's no opportunity for the user to disable specific filters. I'm inclined to come back to fleshing out this functionality at a future date in interest of preventing blocking Gazelle on her work for the Search results page.

web/app/containers/product-list/partials/product-list-filter-modal.jsx
```diff
@@ -0,0 +1,131 @@
...
+                            <span className="u-text-extra-lighter">Filter Results By</span>
+                        </h1>
+                    </HeaderBarTitle>
+
+                    <HeaderBarActions>
+                        <IconLabelButton iconName="close" label="" onClick={closeModal}>Close</IconLabelButton>
+                    </HeaderBarActions>
+                </HeaderBar>
+
+                <Accordion initialOpenItems={[0]}>
```
jeffkamo: That sounds like reasonable possibility. Although in the interest of time, and to prevent us from blocking Gazelle, we may have to come back to flesh out this functionality at a later date. We don't currently know when platform will have the time to expand Merlin's filtering capabilities.

web/app/store/categories/parsers/product-list.js
```diff
@@ -1,6 +1,50 @@
...
+const priceFilterParser = ($, $html) => {
+    const $priceOptions = $html.find('.filter-options .item')
+
+    return [{
+        label: 'Price',
+        ruleset: 'price',
+        kinds: $priceOptions.map((idx, kind) => {
+            const $kind = $(kind)
+            let query = $kind.find('a')[0].search
+            const price = query.split('=')[1]
```
jeffkamo: This is pretty specific to Merlin's magento implementation, and I am assuming that their API for price filtering requires the `=`. That is of course indeed a wild assumption on my part...

web/app/store/categories/parsers/product-list.js
```diff
@@ -1,6 +1,50 @@
...
+            const $count = $kind.find('.count').remove()
+            const count = $count.text().trim()
+
+            query = query.replace(REGEX_DASH, 'to').replace(REGEX_NON_ALPHA_NUM, '')
```
jeffkamo: It was to keep the query (and therefore the filters' key and IDs) alpha numeric. Maybe that isn't necessary, but that was my thought at the time.

web/app/containers/product-list/partials/product-list-filter-modal.jsx
```diff
@@ -120,7 +119,7 @@ const mapStateToProps = createPropsSelector({
 })
 
 const mapDispatchToProps = {
-    changeFilter: (e) => changeFilterTo(e.target.closest('[id]').id),
+    changeFilter: (id) => changeFilterTo(id),
```
jnurse: Since you're no longer passing different information to changeFilterTo, you don't need to add a function here. It can just be written as:

`changeFilter: changeFilterTo`

web/app/containers/product-list/selectors.js
```diff
@@ -39,11 +39,11 @@ export const getActiveFilters = createSelector(
...
 
-        filters.map((filter) => filter.get('kinds').forEach((kind) => {
-            if (kind.get('active')) {
-                activeFilters = activeFilters.push(kind)
-            }
-        }))
+        filters.forEach((filter) => {
+            activeFilters = activeFilters.concat(
+                filter.get('kinds').filter((kind) => kind.get('active'))
+            )
```
jnurse: [Non-blocking] If you wanted to keep using `activeFilters.push()` something like this should work:

`activeFilters.push(...filter.get('kinds').filter((kind) => kind.get('active'))`

Up to you though, what you have right now is totally fine 👌

web/app/store/categories/parsers/product-list.js
jnurse: I think parsers are always going to be specific to the implementation, and since we just have the one filter type right now this seems like a reasonable assumption. Maybe when more filter types are added to Merlin's we can re-visit this.

web/app/store/categories/parsers/product-list.js
jnurse: AFAIK keys and IDs can have dashes. If you're cool with getting rid of it, it might save others a bit of confusion when looking through the code later.

web/app/utils/filter-utils.js
```diff
@@ -17,69 +17,42 @@ const RULESETS = {
...
+ * byFilters - An array filter callback to shrink down a list of items
  * based on the provided filter tokens. Example usage:
  *
  *    const items = [{}, {}, ...]
  *    const tokens = getFilters.toJS().reduce(toTokens, [])
- *    items.filter(byTokens(tokens)) => items matching token criteria
+ *    items.filter(byFilters(tokens)) => items matching token criteria
  *
- * @param  {array} tokenList - list of filter tokens (see toToken function
+ * @param  {array} filterList - list of filter tokens (see toToken function
```
jnurse: This comment should probably be updated a bit more, since we no longer have a toToken function.

web/app/store/categories/parsers/product-list.js
```diff
@@ -1,6 +1,50 @@
...
+        label: 'Price',
+        ruleset: 'price',
+        kinds: $priceOptions.map((idx, kind) => {
+            const $kind = $(kind)
+            let query = $kind.find('a')[0].search
+            const price = query.split('=')[1]
+            const $count = $kind.find('.count').remove()
+            const count = $count.text().replace(REGEX_NON_NUM, '')
+
+            query = query.replace(REGEX_DASH, 'to').replace(REGEX_NON_ALPHA_NUM, '')
```
jeffkamo: @jnurse ah! In trying to remove this `.replace(REGEX_DASH, 'to')` I remember why I did it.

It's not because of the key or id, it's to prevent the second `replace()` from removing the dash. That way, we avoid `price20-30` from becoming `price2030` which is perhaps safe, but does make it vague (is that 20-30, 0-2030 or 2030-Infinity?)

jnurse: Ok makes sense, might be worth adding a comment explaining why we're replacing those characters.

PR #543, opened 2017-04-19
Web 1337  sort

gsaray:
sort filter implementation

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1337

## Changes
- Implement sort utils and use it in selectors to sort products 

## To Do
- have to finalize the UI after filter is added 

## How to test-drive this PR
- Run `npm run dev`
- go to PLP and sort

### DIFF
--- a/web/app/containers/product-list/selectors.js
+++ b/web/app/containers/product-list/selectors.js
@@ -42,18 +42,14 @@ export const getProductListProducts = createSelector(
 export const getSortedListProducts = createSelector(
     getProductListProducts,
     getSort,
-    (products, sort, activeSort) => {
+    (products, sort) => {
         const arrayOfProducts = products.toJS()
-        const options = sort.toJS().options
+        const options = sort.get('options')
 
-        if (options) {
-            options.forEach((option) => {
-                if (option.selected === true) {
-                    activeSort = option.value
-                }
-                return activeSort
-            })
+        if (!options) {
+            return arrayOfProducts
         }
-        return arrayOfProducts.sort(sortLib[activeSort])
+        const activeSort = options.find((option) => option.get('selected'))
+        return arrayOfProducts.sort(sortLib[activeSort.get('value')])
     }
 )
--- a/web/app/containers/product-list/selectors.js
+++ b/web/app/containers/product-list/selectors.js
@@ -54,8 +54,6 @@ export const getSortedListProducts = createSelector(
                 return activeSort
             })
         }
-
-
         return arrayOfProducts.sort(sortLib[activeSort])
     }
 )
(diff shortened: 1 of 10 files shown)


web/app/containers/product-list/selectors.js
```diff
@@ -30,8 +31,29 @@ export const getHasProducts = createSelector(
...
 )
+
+export const getSortedListProducts = createSelector(
+    getProductListProducts,
+    getSort,
+    (products, sort, activeSort) => {
+        const arrayOfProducts = products.toJS()
+        const options = sort.toJS().options
+
+        if (options) {
```
jnurse: Instead of checking that options exist we could move this code below the if statement, and instead check if options doesn't exist and [return early](https://github.com/mobify/mobify-code-style/tree/develop/es5).

web/app/containers/product-list/selectors.js
```diff
@@ -30,8 +31,29 @@ export const getHasProducts = createSelector(
...
+            options.forEach((option) => {
+                if (option.selected === true) {
+                    activeSort = option.value
+                }
+                return activeSort
```
jnurse: I'm not sure forEach is the best approach here. Returning within a forEach doesn't really do anything, plus you probably want to filter the options (or something similar) instead. In fact Immutable lists have a function called `.find()` that might work better here.

web/app/store/categories/actions.js
```diff
@@ -1,7 +1,20 @@
 import {createAction, urlToPathKey} from '../../utils/utils'
 import productListParser from './parsers/product-list'
+import {getCurrentPathKey} from '../../containers/app/selectors'
+import {getSelectedCategory} from '../../containers/product-list/selectors'
 
 export const receiveCategory = createAction('Receive Category Data')
+export const changeSort = createAction('Change Sort Option')
+
+export const changeSortIn = (sortValue) => (dispatch, getStore) => {
```
jnurse: I'm not sure I understand the reasoning behind calling this function `changeSortIn`, maybe this function should be called `changeSort` and the plain action above could be `receiveSortOptions`?

web/app/utils/sort-utils.js
```diff
@@ -0,0 +1,30 @@
...
+            return 1
+        }
+        return 0
+    },
+
+    // sort by price
+    price: (a, b) => {
+        // replacing $ sign with empty string to compare the price
+        const priceA = (a.price.replace(DOLLAR_SIGN, ''))
+        const priceB = (b.price.replace(DOLLAR_SIGN, ''))
```
jnurse: We store prices as strings, but we probably want to convert them to floats so the `priceA - priceB` comparison works correctly.

PR #544, opened 2017-04-20
[Integration Manager] Fix add to cart - better handling of form_key

jeremywiebe:
This PR fixes the Add to Cart feature so that it (hopefully) is less finicky.

*Background*
The `form_key` is a session-level magento token that must be submitted with each form. The Magento desktop creates this value and puts it in a cookie on the client-side. This PR makes the Merlin's connector do the same thing. 

## Changes
- Create and set a `form_key` cookie during app "startup" in the Merlin's connector
- Remove the app-wide `web/app/containers/app/app-parser.js` as it's no longer used (there's one in the Merlin's connector now)
- Created a `submitForm` method for Merlin's to wrap `makeFormEncodedRequest` and handle the `form_key` management
- Remove `getFormKey` from `web/app/containers/app/selectors.js` as that's now in the Merlin's connector also
- Move login eventing for Astro into the `NativeConnector`
- Wrap `CheckoutConfirmation` in its `container.jsx` instead of in the `templates.js` file
- 

## How to test-drive this PR
- Run `npm run dev`
- Add a few items to the cart
- Proceed all the way through checkout


web/app/integration-manager/_merlins-connector/cart/commands.js
```diff
@@ -33,16 +34,17 @@ export const getCart = () => (dispatch) => {
...
     const formValues = {
         ...formInfo.get('hiddenInputs').toJS(),
         qty
     }
-    return makeFormEncodedRequest(formInfo.get('submitUrl'), formValues, {method: formInfo.get('method')})
-        .then(() => {
-            return dispatch(getCart())
-        })
+
+    formValues.form_key = getFormKey(getState())
```
jnurse: Since we need the state in two different locations, could you store it in a variable at the top of the function instead of calling getState() twice?

web/app/integration-manager/_merlins-connector/cart/commands.js
```diff
@@ -57,8 +59,8 @@ export const addToCart = (key, qty) => (dispatch, getStore) => {
  *   busts a cache. removeFromCart() will call getCart() once the request to remove the item has completed
  */
 export const removeFromCart = (itemId) => {
-    return (dispatch, getState) => {
-        return makeFormEncodedRequest(REMOVE_CART_ITEM_URL, {item_id: itemId, form_key: getFormKey(getState())}, {method: 'POST'})
```
jnurse: I'm not sure I follow why we're not pulling the formKey off the state here, but we are in the getCart command.

web/app/integration-manager/_merlins-connector/cart/commands.js
jeremywiebe: Oops, this should be deleted.  `submitForm()` manages adding the `form_key` which is why I didn't do it in `removeFromCart` below.

web/app/integration-manager/_merlins-connector/cart/commands.js
jeremywiebe: The new `submitForm` method manages the `form_key` so that we don't have to remember to do it everywhere we submit a form.

web/app/containers/app/actions.js
```diff
@@ -62,11 +60,6 @@ export const onPageReceived = createAction('On page received', [
 
 export const setFetchedPage = createAction('Set fetched page', ['url'])
 
-export const receiveData = createAction('Receive App Data')
-export const process = ({payload: {$response}}) => {
```
jeremywiebe: Oops, I completely missed that.  I'll restore them for now and your Checkout Confirmation PR can do the last work to remove them.

PR #546, opened 2017-04-21
add shipping rate and tax in payment step

gsaray:
add shipping rate and tax in payment step
 **JIRA**: https://mobify.atlassian.net/browse/WEB-1328

## Changes
- add shipping rate and tax in payment step

## How to test-drive this PR
- Run `npm run dev`
- Add an item to cart and continue to payment


web/app/containers/cart/_cart.scss
```diff
@@ -117,6 +117,18 @@ $cart__custom-ease-out: cubic-bezier(0.35, 0.35, 0.32, 1);
 }
 
 
+// Estimate Shipping Modal
+// ---
+
+.t-cart__estimate-shipping-modal {
+    .pw-inline-loader__loading-dot {
```
jeffkamo: Instead of overriding a component style with this template style (which has high specificity), can you update the inline loader theme to have this style instead?

web/app/containers/cart/partials/cart-estimate-shipping.jsx
```diff
@@ -63,9 +65,15 @@ export const CartEstimateShippingModal = ({closeModal, isOpen, countries, stateP
...
                     <FieldRow>
-                        <Button className="c--secondary u-width-full u-text-uppercase" type="submit">
-                            Get Estimate
-                        </Button>
+                        {!isTaxRequested ?
+                            <Button className="c--secondary u-width-full u-text-uppercase" type="submit">
+                                Get Estimate
+                            </Button>
+                        :
+                            <Button className="c--secondary u-width-full" type="submit">
```
mlworks: I would remove `type="submit"` since the form is already being submitted we probably don't want the user to click on it repeatedly causing multiple requests

web/app/containers/cart/_cart.scss
gsaray: inline loader theme is using `brand-color` as background color (you can see this when adding an item to wishlist). how can I add a different theme specific to the loader used in shipping modal?

gsaray: I added a modifier to inline loader instead of overriding it in template.

PR #548, opened 2017-04-21
[Integration manager] merlins confirmation page

jnurse:
This PR moves the merlins confirmation page actions to the merlins connector and implements some missing functionality for updating a users shipping and billing info that was previously broken on desktop. It also moves the container fetchers into the router, this is because the app container needs the current container's fetcher function to provide offline support (modal reload button) and the react-loader component doesn't let us access the template component. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-50
 **Linked PRs**: n/a

## Changes
- Moved fetch confirmation page command to the merlins connector
- Moved checkout registration command to the merlins connector
- Fixed bug with shipping page for logged in user that prevented submitting the form
- Moved updateShippingAndBilling command to the merlins connector
- Fixed bug where new users addresses weren't actually being saved
- Added back commented out code since desktop is no longer broken
- Update confirmation page action tests
- Moved page fetchers to the router so offline support reload function works

## How to test-drive this PR
- Run `npm run dev`
- add an item to your cart and complete an order
- the order confirmation page should work as expected
- *Note* this page should be tested as a logged in user and as a guest user


web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -165,6 +178,22 @@ export const checkoutSignIn = (formValues) => {
     }
 }
 
+export const checkoutRegister = (userCredentials) => {
+    const postCreateAccountURL = '/customer/account/createpost/'
```
DanielBrookRoberge: This can go outside the function since it doesn't change.

web/app/template.jsx
```diff
@@ -23,12 +23,10 @@ const template = (WrappedComponent) => {
 
             dispatch(onRouteChanged(url, route.routeName))
 
-            if (WrappedComponent.fetcher) {
-                WrappedComponent.fetcher(route.fetchUrl || url, route.routeName, dispatch)
+            if (route.fetcher) {
+                dispatch(route.fetcher(route.fetchUrl || url, route.routeName))
```
DanielBrookRoberge: This is a lot nicer than how things were before! What do you think about renaming it `route.fetchAction` to make clear that it needs to be an action creator?

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -165,6 +178,22 @@ export const checkoutSignIn = (formValues) => {
...
+    return submitForm(postCreateAccountURL, userCredentials, {method: 'POST'})
+        .then((response) => {
+            const responseUrlHas = (chunk) => response.url.search(chunk) >= 0
+            const redirectUrlIsNotToCreate = responseUrlHas('/account/') && !responseUrlHas('/create/')
+            const registrationIsSuccess = response.redirected && redirectUrlIsNotToCreate
```
DanielBrookRoberge: `response.redirected` appears to be a new feature in Chrome 57 and means that we'll get an error in earlier browsers (and probably with the polyfill, too).

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -227,3 +256,135 @@ export const submitPayment = (formValues) => (dispatch, getState) => {
             }
         })
 }
+
+const buildFormData = (formCredentials) => {
+    const formData = new FormData()
+
+    for (const key in formCredentials) {
+        if (Object.prototype.hasOwnProperty.call(formCredentials, key)) {
```
DanielBrookRoberge: Could this be `Object.keys(formCredentials).forEach()` instead?

web/app/template.jsx
jnurse: Good idea!

jnurse:
@DanielBrookRoberge comments have been addressed 😄

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: I'm fairly certain we won't need to use `hasOwnProperty` with the `Object.keys()` call.

web/app/containers/app/container.jsx
```diff
@@ -148,11 +148,11 @@ class App extends React.Component {
 
 App.propTypes = {
     children: PropTypes.element.isRequired,
-    fetchPage: PropTypes.func.isRequired,
     /**
```
jeremywiebe: Is this comment in the wrong spot now?

web/app/containers/checkout-confirmation/actions.js
```diff
@@ -132,17 +20,25 @@ export const submitRegisterForm = () => {
...
+        return checkoutRegister(userCredentials)
+            .then(() => {
+                dispatch(openModal(CHECKOUT_CONFIRMATION_MODAL))
+                dispatch(updatingShippingAndBilling())
+                dispatch(hideRegistrationForm())
+            })
+            .catch((error) => {
+                if (error.name !== 'SubmissionError') {
+                    dispatch(addNotification({
+                        content: `Sorry, registration Failed. Contact us for assistance. ${error.message}`,
```
jeremywiebe: Should 'Failed' be lowercase?

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -227,3 +255,132 @@ export const submitPayment = (formValues) => (dispatch, getState) => {
...
+        region,
+        postcode,
+        telephone
+    } = formValues
+
+    return {
+        firstname,
+        lastname,
+        company: company || '',
+        telephone: telephone ? telephone.replace(/\(\) /g, '') : '',
```
jeremywiebe: I'm not sure this does what you want. Did you mean to strip parens and spaces? 

screen shot 2017-04-28 at 2 27 08 pm

If so, I think you want to make that a character set in the regex: `replace(/[() ]/g, '')`

screen shot 2017-04-28 at 2 28 12 pm

web/app/router.jsx
```diff
@@ -13,6 +13,13 @@ import Home from './containers/home/container'
 import CheckoutHeader from './containers/checkout-header/container'
 import CheckoutFooter from './containers/checkout-footer/container'
 
+import {fetchHomeData} from './integration-manager/home/commands'
```
jeremywiebe: I'm wondering if we should be referencing IM commands here directly. In my mind we'd always wrap IM commands and surface those wrappers to anything in the app (UI components, page fetch actions, etc). 

Thoughts?  Maybe it's not a big thing to worry about now as it's pretty easy to create a wrapping thunk action and reference it here if we need to.

web/app/template.jsx
```diff
@@ -23,12 +23,10 @@ const template = (WrappedComponent) => {
 
             dispatch(onRouteChanged(url, route.routeName))
 
-            if (WrappedComponent.fetcher) {
-                WrappedComponent.fetcher(route.fetchUrl || url, route.routeName, dispatch)
+            if (route.fetchAction) {
```
jeremywiebe: Seeing this now, I really like it also!  🎉

web/app/router.jsx
jnurse: We discussed this offline and decided to re-visit it when we audit the integration manager command API.

PR #551, opened 2017-04-27
Web 1350  search result

gsaray:
Implement Search Results on Merlin's PWA

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1350

## Changes
- Implement Search Results on Merlin's PWA

## How to test-drive this PR
- Run `npm run dev`
- go to https://www.merlinspotions.com/catalogsearch/result/?q=potion


web/app/containers/search-result/partials/search-result-content.jsx
```diff
@@ -0,0 +1,116 @@
...
+const SearchResultContent = ({
+    hasProducts,
+    products,
+    sortChange
+}) => (
+    <div className="t-search-result__container u-padding-end u-padding-bottom-lg u-padding-top-lg u-padding-start">
+        {hasProducts ?
+            <div>
+                <div className="u-padding-bottom-md u-padding-start-sm u-padding-end-sm">
+                    <div className="u-flexbox">
```
jnurse: This section looks pretty similar to what we have on the PLP, could it be moved into a component that both templates could use?

web/app/containers/search-result/partials/search-result-content.jsx
```diff
@@ -0,0 +1,116 @@
...
+                        <div className="u-flex u-margin-end-md">
+                            <div className="u-text-semi-bold u-margin-bottom-sm">
+                                {products.length} Items
+                            </div>
+                            <Button disabled
```
jnurse: Will the filter always be disabled? or is it just disabled for now?

web/app/containers/search-result/partials/search-result-content.jsx
```diff
@@ -0,0 +1,116 @@
...
+    <div className="u-flexbox u-direction-column u-align-center">
+        <Image
+            className="u-flex-none"
+            alt="Crystal Ball"
+            width="122px"
+            height="110px"
+            src={getAssetUrl('static/img/global/no-results.png')} />
+
+        <div className="t-search-result__no-results-text u-text-align-center">
+            your search returned no results. Please check your spelling and try searching again.
```
jnurse: This should probably start with a capital y, instead of a lowercase one.

web/app/store/search-result/reducer.js
```diff
@@ -0,0 +1,12 @@
+import Immutable from 'immutable'
+import {handleActions} from 'redux-actions'
+import {mergePayload} from '../../utils/reducer-utils'
+import {receiveSearchResult, receiveSearchResultFilters, changeSearchResultSortOption} from './actions'
+
+const searchResultReducer = handleActions({
+    [receiveSearchResult]: mergePayload,
+    [changeSearchResultSortOption]: mergePayload,
+    [receiveSearchResultFilters]: mergePayload
```
jnurse: I wonder if it's necessary to have the search-results in it's own branch of the store, or if we can keep it under categories like other PLP data? That way you would be able to cache data for multiple searches within the store, instead of always overriding the previous search data when a new one is performed.

web/app/store/products/parser.js
```diff
@@ -42,3 +42,26 @@ export const productDetailsParser = ($, $html) => {
         description: getTextFrom($mainContent, '.product.info.detailed .product.attibute.description p')
     }
 }
+
+export const searchResultParser = ($, $html) => {
+    const $products = $html.find('.item.product-item')
+    const productMap = {}
+    $products.each((_, product) => {
```
jnurse: This parser looks very similar to `productListParser`, maybe you could use that parser instead of making a new one?

web/app/containers/search-result/partials/search-result-content.jsx
```diff
@@ -0,0 +1,116 @@
...
+                            </label>
+
+                            <div>
+                                <div className="u-position-relative u-width-full">
+                                    <select
+                                        className="t-search-result__sort-select"
+                                        onChange={(e) => { sortChange(e.target.value) }}
+                                        onBlur={(e) => { sortChange(e.target.value) }}
+                                    >
+                                        <option value="name">Name</option>
```
jnurse: Could the data for these options be pulled from the redux store instead of being hardcoded?

web/app/containers/search-result/partials/search-result-content.jsx
```diff
@@ -0,0 +1,116 @@
...
+import List from 'progressive-web-sdk/dist/components/list'
+import Button from 'progressive-web-sdk/dist/components/button'
+import Icon from 'progressive-web-sdk/dist/components/icon'
+import Image from 'progressive-web-sdk/dist/components/image'
+import ProductTile from '../../../components/product-tile'
+
+const ResultList = ({products}) => {
+    return (
+        <List className="c--borderless">
+            {products.map((product, idx) => (<ProductTile key={idx} {...product} />))}
```
jnurse: Using an index for the key prop can be problematic, [keys need to uniquely identify the item](https://facebook.github.io/react/docs/lists-and-keys.html#keys) so React can correctly determine what's changed. It would be better to use a string from the product object if possible, like the product name or product ID?

web/app/containers/search-result/partials/search-result-content.jsx
```diff
@@ -0,0 +1,116 @@
...
+            src={getAssetUrl('static/img/global/no-results.png')} />
+
+        <div className="t-search-result__no-results-text u-text-align-center">
+            your search returned no results. Please check your spelling and try searching again.
+        </div>
+    </div>
+)
+
+NoResultsList.propTypes = {
+    bodyText: PropTypes.string
```
jnurse: I don't think this prop is being used.

web/app/containers/search-result/selectors.js
```diff
@@ -0,0 +1,48 @@
...
+)
+export const getSearchResultProducts = createSelector(
+    getProducts,
+    getProductPaths,
+    (products, productUrls) => {
+        return productUrls.map((path) => {
+            return products.get(path)
+        })
+    }
+)
```
jnurse: These selectors look pretty similar to the ones for the PLP too. If these stay as separate templates then some of these selector functions should probably get moved into a util file so we can have just one copy of the functions.

web/app/router.jsx
```diff
@@ -47,6 +47,7 @@ const Router = ({store}) => (
                 <Route component={ProductList} path="charms.html" routeName="productListPage" />
                 <Route component={ProductDetails} path="checkout/cart/configure/id/*/product_id/*/" routeName="cartEditPage" />
                 <Route component={ProductDetails} path="*.html" routeName="productDetailsPage" />
+                <Route component={SearchResult} path="catalogsearch/result/*" routeName="searchResultPage" />
```
jnurse: This might be a good place to use React Routers [URL parameters](https://reacttraining.com/react-router/web/example/url-params).

web/app/store/search-result/actions.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const changeSearchResultSortOption = createAction('Change Search Result Sort Option', 'sort')
+
+export const changeSort = (sortValue) => (dispatch, getStore) => {
+    const currentState = getStore()
+    const getSortData = getSort(currentState).toJS()
+    getSortData.options.forEach((option) => {
+        option.selected = option.value === sortValue
+    })
+    dispatch(changeSearchResultSortOption(getSortData))
```
jnurse: This is another case of the logic being almost identical to the same action on a PLP. If these templates stay separate then we should find a way to refactor the two actions so all the logic is in just one place (like maybe in a util file).

web/app/containers/search-result/partials/search-result-content.jsx
gsaray: The filter is disabled for now becaus there is no filter on search results page

web/app/containers/product-list/partials/search-result-header.jsx
```diff
@@ -0,0 +1,42 @@
...
+]
+
+const SearchResultHeader = ({searchTerm}) => (
+    <div className="u-flexbox u-align-bottom">
+        <div className="u-flex u-padding-top-lg u-padding-bottom-lg u-padding-start-md">
+            <Breadcrumbs items={items} />
+            <div className="u-margin-top-md">
+                {searchTerm ?
+                    <h1 className="u-text-uppercase">
+                        <span>Result For </span>
```
jnurse: Should this say "Results"?

web/app/store/products/parser.js
```diff
@@ -42,3 +42,26 @@ export const productDetailsParser = ($, $html) => {
         description: getTextFrom($mainContent, '.product.info.detailed .product.attibute.description p')
     }
 }
+
+export const searchResultParser = ($, $html) => {
```
jnurse: Can we get rid of this parser now?

web/app/store/products/actions.js
```diff
@@ -1,11 +1,14 @@
...
 
 export const receiveProductListProductData = createAction('Receive ProductList product data')
 export const receiveProductDetailsProductData = createAction('Receive ProductDetails product data')
+export const receiveSearchResultProductData = createAction('Receive SearchResult product data')
 
 export const processProductList = ({payload: {$, $response}}) =>
     receiveProductListProductData(productListParser($, $response))
 export const processProductDetails = ({payload: {$, $response, url}}) =>
     receiveProductDetailsProductData({[urlToPathKey(url)]: productDetailsParser($, $response)})
+export const processSearchResult = ({payload: {$, $response}}) =>
```
jnurse: I don't think this or the `receiveSearchResultProductData` action are needed anymore.

web/app/store/products/reducer.js
```diff
@@ -1,14 +1,17 @@
...
+import {receiveProductListProductData, receiveProductDetailsProductData, receiveSearchResultProductData} from './actions'
 
 const initialState = Immutable.Map()
 
 const productReducer = handleActions({
     [receiveProductDetailsProductData]: mergePayload,
     [receiveProductListProductData]: (state, {payload}) => {
         return state.mergeDeepWith((prev, next) => prev || next, payload)
+    },
+    [receiveSearchResultProductData]: (state, {payload}) => {
```
jnurse: I think we can get rid of this too.

web/app/store/selectors.js
```diff
@@ -10,3 +10,4 @@ export const getProducts = ({products}) => products
 export const getCart = ({cart}) => cart
 export const getCheckout = ({checkout}) => checkout
 export const getForm = ({form}) => form
+export const getSearchResult = ({searchResult}) => searchResult
```
jnurse: 🔫

PR #556, opened 2017-04-28
Setup to use the new analytic connector

wizardlyhel:
**JIRA**: https://mobify.atlassian.net/browse/MTT-602
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/626

## Changes
- Setup required to use the new Analytic Connector

## How to test-drive this PR

**Requirement** We need the progressive-web-sdk's **develop** branch

For whatever reason, when you `npm link` the `progressive-web-sdk`, the react app will crash at PDP.

We need to get to the checkout in order test transaction analytics.

1. Change `package.json` so that progressive SDK is pointing to your local file `"progressive-web-sdk": "file:../../progressive-web-sdk"`
2. Run `rm -r node_modules/progressive-web-sdk/ && npm i && cd node_modules/progressive-web-sdk/ && npm i && npm run dev:build && rm -r node_modules/react node_modules/react-dom && cd -` This builds the SDK in your local node_modules
3. Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)

Observe the console log:
- Navigate normally and you see pageview analytics
- Buy a product and you will see transaction analytics


jansepar:
The linked PR seems to link to the wrong place.

wizardlyhel:
Updated

web/app/store/index.js
```diff
@@ -17,15 +17,21 @@ import {reducer as imReducer} from '../integration-manager/reducer'
...
+    mobifyGAID: 'UA-53825302-1',
+    ecommerceLibrary: 'ec',
+    debug: true
+})
 
 const configureStore = (initialState) => {
     const middlewares = [
-        thunk,
-        analytics(({type, payload}, state) => analyticManager.distribute(type, payload, state))
+        analytics(({type, payload}, state) => analyticsManager.distribute(type, payload, state)),
```
jeremywiebe: I haven't tested this, but does moving this above the `thunk` middleware cause the analytics middleware to see dispatched thunk actions?  And if so, does it safely ignore them?

web/app/integration-manager/checkout/results.js
```diff
@@ -2,10 +2,50 @@
...
+            [Product.PRICE]: detail.price,
+            [Product.QUANTITY]: product.quantity
+        })
+    })
+    return mappedProducts
+}
+
+const realReceiveCheckoutConfirmationData = createActionWithAnalytics(
+    'Receive Checkout Confirmation Data',
+    ['confirmationData', 'purchaseData'],
```
DanielBrookRoberge: The `purchaseData` variable will be merged into the Redux store if it is named here. We should probably avoid that.

web/app/integration-manager/checkout/results.js
```diff
@@ -2,10 +2,50 @@
...
+                [Transaction.REVENUE]: purchaseData[Transaction.REVENUE],
+                [Transaction.TAX]: purchaseData[Transaction.TAX]
+            },
+            purchaseData.products
+        )
+    }
+)
+
+// This is a proxy action to get state information before dispatching the real intended action
+export const receiveCheckoutConfirmationData = (confirmationData) => (dispatch, getState) => {
```
jeremywiebe: Talked with DBR about having this proxy thunk action. Sounds like the long term plan is that the various connectors would do this work of selecting data they need using selectors. Can you note that here with a comment in the code?

web/app/integration-manager/checkout/results.js
DanielBrookRoberge: Does the confirmation page still work?

wizardlyhel: Yea it does work ... but good point, I will remove that

web/app/store/index.js
wizardlyhel: ... now I think about it ... no, it doesn't make any difference if I change the order.

`thunk` evaluates the the payload and absorbs the dispatched action if it is a function. When the payload is at the function state, there is no way to know if the passed in arguments contains the analytic data the `redux-analytics` requires until you execute the function.

I'll revert the order

PR #561, opened 2017-05-02
Add Parallelization to Tests

jasecode:
Goal is to Decrease Average Build Times of the Platform scaffold to ~5 minutes.
Ideas:
1) effective Parallelization of tests across separate containers
2) backgrounding non-blocking tasks to decrease bottleneck
3) upgrade selenium to use 3.0

Average Build Time:
Goal:  ~5 minutes
Current Average Build Time ~6 minutes
Before: ~7.5 minutes

## Changes
- Circle.yml 
-- Include parallel towards our tasks
-- Made use of "background: true" for non-blocking dependencies
-- Strategically ordered the dependencies so they can work in parallel to set-up
- start-test-server.sh
-- starting "npm run dev" for the test server is now a separate script and backgrounded
- package.json
-- upgrade selenium and nightwatch
- creating run-all-tests.sh  which covers most of the logic of separating tests via circle nodes (containers)
-- Node 0 will run Lighthouse Tests
-- Node 1 will run every other test: Lint, Unit, End-to-End
-- Node 2 (if it exists) will split End-to-End tests and will run them in parallel
By default, we are expecting 2 Circle Containers per project.
-- If there is only 1 Node, it'll just run all our tests regularly
## How to test-drive this PR
- Rebuild on Circle


web/tests/system/start-test-server.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+# Start preview if branch is not master.
+if [ "$CURRENT_BRANCH" != "master" ]; then
+    echo "Running tests against local build."
+    # Kill background processes when this script exits.
+    trap 'kill $(jobs -p)' EXIT
+    export ACTIVE_PROFILE=local
+    echo "Running Test Server."
+    npm run dev &
```
ellenmobify: Can you use `npm run prod:build` and  `npm run test:server` here?

ellenmobify:
👍

circle.yml
```diff
@@ -6,24 +6,25 @@ dependencies:
...
     pre:
-        - ./web/scripts/install-google-chrome-stable.sh
+        - mkdir $CIRCLE_ARTIFACTS/logs
+        - ./web/scripts/install-google-chrome-stable.sh $> $CIRCLE_ARTIFACTS/logs/installChrome.log:
+            background: true
+            parallel: true
         - npm install:
             pwd: web
+            parallel: true
+        - ./tests/system/start-test-server.sh &> $CIRCLE_ARTIFACTS/logs/startTestServer.log:
```
jeremywiebe: Does this step depend on `npm install` having completed?  It looks like it's set to run in parallel with the previous two steps...

web/tests/system/run-all-tests.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -eu
+set -o pipefail
+set -o nounset
+
+
```
jeremywiebe: Nitpick, but we typically don't have more than one consecutive blank line in any source file. 

Shorthand is: 🔫   

:)

web/tests/system/run-all-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ $CIRCLE_NODE_TOTAL -eq 1 ]; then
+  echo 'Running lint'
+  npm run lint
+  echo 'Running Unit Tests'
+  npm test -- --runInBand
+  echo 'Running Lighthouse Tests'
+  npm run test:pwa-ci
+  echo 'Running End to End Tests'
+  ./tests/system/start-test-server.sh
```
jeremywiebe: Isn't this already running because of line 16 of `circle.yml`?

web/tests/system/start-test-server.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -eu
+set -o pipefail
+set -o nounset
+
+
+if git rev-parse ; then
```
jeremywiebe: What is this doing here? Checking if we're in a git repo?

jasecode:
Hey @jeremywiebe that appears to be circle2.0 which has more parallelization control. We'll address new ways to parallelize when we upgrade to 2.0.

PR #562, opened 2017-05-02
Integrate Messaging PWA support

benlast:
Integrate Messaging PWA support

 **JIRA**: https://mobify.atlassian.net/browse/MSG-474
 **Linked PRs**: https://github.com/mobify/duncan/pull/67

## Changes
- Remove old messaging support in `web/app/utils/push-messaging`
- Use Messaging worker parameters to load service worker in `web/app/loader.js`
- Load new messaging client in `web/app/loader.js` (after all other scripts)
- If not running in Astro, use `importScripts` to include Messaging service worker in `web/service-worker-loader.js`

## How to test-drive this PR
- Run the tests
- Run `npm run dev`


cdegit:
When I preview against this branch, the content doesn't seem to be loading:
screen shot 2017-05-03 at 4 19 20 pm

web/app/loader.js
```diff
@@ -35,9 +34,86 @@ const attemptToInitializeApp = () => {
     const ASTRO_CLIENT_CDN = `//assets.mobify.com/astro/astro-client-${ASTRO_VERSION}.min.js`
     const SW_LOADER_PATH = `/service-worker-loader.js?preview=${IS_PREVIEW}&b=${cacheHashManifest.buildDate}`
 
+    const MESSAGING_PWA_SW_VERSION_PATH = 'https://webpush-cdn.mobify.net/pwa-serviceworker-version.json'
```
cdegit: Could some of this be moved into loader-utils, or another file for messaging utils? If possible, I think limiting the amount code we include in `loader.js` will make it easier for partners to understand how the app actually starts up.

fractaltheory:
@cdegit resolved the issue you were seeing with the broken page loading: https://github.com/mobify/platform-scaffold/pull/562/commits/56dc3287f216c03c8db1774f6bbfc33ac8ccd6d3

web/app/loader.js
```diff
@@ -56,6 +132,67 @@ const attemptToInitializeApp = () => {
...
+                            siteId: MESSAGING_SITE_ID  // replaced at build time
+                        }
+                    )
+                }
+            )
+            .catch(
+                // Catch the error to silence logging
+                (error) => {
+                    console.log(`Error loading ${MESSAGING_PWA_CLIENT_PATH}: ${error}`)
+                    Promise.reject(error)
```
jeremywiebe: I've not seen this before.  We don't return the rejected Promise, so how does this propagate?

web/app/loader.js
```diff
@@ -101,49 +243,64 @@ const attemptToInitializeApp = () => {
...
+            }
 
-            loadScript({
-                id: 'progressive-web-vendor',
-                src: getAssetUrl('vendor.js')
-            })
+            Promise.all(loadingPromises)
+                .then(
+                    // Execute any deferred functions
+                    deferredUntilLoadComplete.forEach((def) => { def() })
```
jeremywiebe: Passing this function call as an expression to `.then()` confused me a bit.  We're essentially ignoring all the return values of the `Promise.all()`'d promises.  

What do you think about this slight modification?

```js
Promise.all(loadingPromises).then(() => deferredUntil... ) 
```

web/app/utils/loader-utils.js
```diff
@@ -1,4 +1,6 @@
-export const loadScript = ({id, src, onload, isAsync = true}) => {
+export const funcType = typeof function() {}
```
jeremywiebe: I think this is overkill.  The ECMA defines that `typeof someFunc` will return `'function'`.

web/app/utils/loader-utils.js
```diff
@@ -9,9 +11,40 @@ export const loadScript = ({id, src, onload, isAsync = true}) => {
...
     document.getElementsByTagName('body')[0].appendChild(script)
 }
+
+export const loadScriptAsPromise = ({id, src, onload, isAsync = true, rejectOnError = true}) => {
+    return new Promise(
+        (resolve, reject) => {
+            loadScript({
+                id,
+                src,
+                onload: typeof onload === funcType ? () => {
```
jeremywiebe: I wonder if there's a slightly more clear way to format this. It took me a second to realize this was a ternary.  

```js
onload: (typeof onload === 'function'
    ? () => { 
        onload()
        resolve()
    }
    : resolve),
```
??  :/ not sure that's any clearer, lol.

web/worker/main.js
```diff
@@ -3,7 +3,25 @@
 
 import worker from 'progressive-web-sdk/dist/worker/main'
 
-worker({
+const workerParams = worker({
```
jeremywiebe: This means we need to be on the latest version of the SDK which was released today. Do you mind adding updating the version number in `package.json`?

web/app/loader.js
benlast: I've refactored some of the consts and functions into the `loader-utils`.

web/webpack/base.loader.js
```diff
@@ -54,6 +56,7 @@ module.exports = {
             }
         }),
         new webpack.DefinePlugin({
+            MESSAGING_SITE_ID: JSON.stringify(webPackageJson.messagingSiteId),
```
benlast: Unless you stringify it one way or another, it gets inserted as an identifier. I changed the stringifying to match the approach taken for the ASTRO_VERSION.

web/app/loader.js
```diff
@@ -35,9 +34,86 @@ const attemptToInitializeApp = () => {
...
+        }
+
+        const hash = versionData.SERVICE_WORKER_CURRENT_HASH
+        if (hash) {
+            workerPathElements.push(`msg_sw_hash=${hash}`)
+        }
+
+        // Add a deferred function that will asynchronously update the Messaging
+        // version data.
+        deferredUntilLoadComplete.push(
```
benlast: I've slimmed it down and now we have one long line:

`deferredUntilLoadComplete.push(() => loadAndInitMessagingClient(IS_PREVIEW, MESSAGING_SITE_ID))`

Personally I don't like the super-long line, but whatever...

benlast: You don't want the function to be called at the same time as `getServiceWorkerURL` because then the fetch for the SW version data competes with more important network stuff. The function that kicks off the fetch has to be pushed before slow operations start, so that we can be sure the deferred list is ready when the scripts all load. I moved it into `setupMessagingClient`.

web/app/loader.js
```diff
@@ -56,6 +132,67 @@ const attemptToInitializeApp = () => {
...
     }
 
+    /**
+     * Start the asynchronous loading and intialization of the Messaging client,
+     * storing a Promise in window.Progressive.MessagingClientInitPromise that
+     * is resolved when the load and initialization is complete. If either load
+     * or init fails, the Promise is rejected.
+     */
+    const loadAndInitMessagingClient = () => {
+        window.Progressive.MessagingClientInitPromise = loadScriptAsPromise({
```
benlast: It's visible to other scripts such as the webpush-client-loader, that may need to know the state of the client and are not part of the PWA app.

web/app/loader.js
```diff
@@ -56,6 +132,67 @@ const attemptToInitializeApp = () => {
...
+        })
+            .then(
+                () => {
+                    // We assume window.Progressive will exist at this point.
+                    const messagingClient = window.Progressive.MessagingClient || {}
+                    // If init is not a function, this will
+                    // throw, and the catch below will
+                    // cause the promise to reject with
+                    // the error.
+                    return messagingClient.init(
```
benlast: Can't we get these in the lint checking? Not that I have a problem fixing up things to match a code style, but it makes for depressing PR feedback :( Anyway, fixed.

web/app/loader.js
benlast: I changed it to `throw`

web/app/loader.js
benlast: Do we care about the return values from `Promise.all`? The only reason the loads have been wrapped in `Promise`s is so that we can tell when they finish, which is what `Promise.all` does. But turning it into a function is fine, so I did that.

web/app/utils/loader-utils.js
benlast: Fair enough. Probably micro-optimization anyway, but I was more guarding against typing `'function'` wrong in one of the several places this is used.

web/app/utils/loader-utils.js
benlast: I pulled it out into an `if` instead.

web/worker/main.js
benlast: The version in `develop` of the SDK is 0.14.8, which is what's in the scaffold's `package.json` on my branch. What am I missing?

benlast:
@cdegit @jeremywiebe thanks for the feedback - I made a number of changes. Could you re-check? Thanks.

web/app/loader.js
jeremywiebe: I agree. I just looked and eslint doesn't appear to have a rule that could enforce this. But there is a tool called [prettier](https://prettier.github.io/prettier/) that would probably fix all these nitpicky things.

web/worker/main.js
jeremywiebe: I _think_ the Scaffold wasn't at 0.14.8 the other day. :)  We're good now.  Thanks

web/app/utils/loader-utils.js
```diff
@@ -13,9 +13,106 @@ export const loadScript = ({id, src, onload, isAsync = true}) => {
...
+    }
 
     document.getElementsByTagName('body')[0].appendChild(script)
 }
+
+export const loadScriptAsPromise = ({id, src, onload, isAsync = true, rejectOnError = true}) => {
+    return new Promise(
+        (resolve, reject) => {
+
+            const resolver = () => {
```
jeremywiebe: ✨

web/app/utils/loader-utils.js
```diff
@@ -13,9 +13,106 @@ export const loadScript = ({id, src, onload, isAsync = true}) => {
...
+            throw error
+        })
+}
+
+const MESSAGING_PWA_SW_VERSION_PATH = 'https://webpush-cdn.mobify.net/pwa-serviceworker-version.json'
+const messagingSWVersionKey = 'messagingServiceWorkerVersion'
+
+/**
+ * Kick off a fetch for the service worker version, returning a Promise
+ * thaht resolves when the fetch is done.
```
jeremywiebe: *that*

benlast: Ouch!

benlast: I was using the ethnic Lithuanian native dialect.

PR #564, opened 2017-05-02
Web 1278  implement saved addresses

jeffkamo:
Implement "Save Address" in Merlin's Shipping step of Checkout.

 **JIRA**: [WEB-1278](https://mobify.atlassian.net/browse/WEB-1278)

## Todos
- [] Users can check out while logged in
- [] Users are able to select from saved addresses on checkout shipping step
- [] Orders are successfully submitted with the selected saved address
- [] Payment step's Billing Address form updates according to the selected saved address

## How to test-drive this PR
- Run `npm run dev`
- *TBD*


web/app/store/checkout/shipping/parser.js
```diff
@@ -19,19 +19,30 @@ export const parseShippingInitialValues = (shippingFieldData) => {
         addressLine1: streetFields[0].value,
         company: fieldData.company.value,
         addressLine2: streetFields[1].value,
-        city: fieldData.city.value,
-        telephone: fieldData.telephone.value,
-        billing_same_as_shipping: true
+        billing_same_as_shipping: true,
+        saved_address: undefined
```
jnurse: Does this key need to be set to undefined or can you just leave it out?

web/app/utils/utils.js
```diff
@@ -157,8 +157,17 @@ export const parseLocationData = (formValues, registeredFieldNames) => {
         const getRegisteredFieldValue = (fieldName) => {
             return registeredFieldNames.includes(fieldName) ? formValues[fieldName] : undefined
         }
-        address.country_id = getRegisteredFieldValue('country_id')
-        address.postcode = getRegisteredFieldValue('postcode')
+
+        const countryId = getRegisteredFieldValue('country_id')
+        if (countryId) {
```
jnurse: If we have formValues shouldn't this data always be here?

jnurse:
Should one of the options be auto selected? Maybe the first saved address? Also have you chatted with Mitch about the design? There's a lot to scroll through if the user has a lot of addresses, although I'm sure the number of addresses on mobifyqa@gmail.com isn't exactly typical :stuck_out_tongue_closed_eyes: I wonder if we could make it scrollable if there are more than 3 addresses?

web/app/containers/checkout-payment/actions.js
```diff
@@ -66,15 +65,23 @@ export const submitPayment = () => {
...
-                countryId: country_id,
+                region_id,
+                country_id,
                 saveInAddressBook: false
             }
         }
 
         const paymentInformation = {
             billingAddress: {
-                ...address
```
jeffkamo: This turned out to be a fair amount of work cleaning this up. Please take a look at my recent commit that attempts to normalize alot of these keys, and formatting of these address objects.

web/app/containers/checkout-shipping/partials/shipping-address.jsx
```diff
@@ -2,226 +2,165 @@ import React from 'react'
...
+        const {city, country_id, firstname, id, lastname, postcode, region} = address
+        const street = address.street.join(', ')
+        const shippingAddress = (
+            <div className="u-color-neutral-40">
+                <p className="u-margin-bottom-sm">
+                    {/* eslint-disable camelcase */}
+                    {city}, {region.region_code}, {country_id}, {postcode}
+                    {/* eslint-enable camelcase */}
+                </p>
+                <p>{firstname} {lastname}</p>
```
jeffkamo: Nice to have, though I am hesitant at this point to put effort into that unless there is a strong reason to.

web/app/containers/checkout-shipping/partials/shipping-address.jsx
```diff
@@ -2,226 +2,165 @@ import React from 'react'
...
-                    <FieldRow>
+        )
+
+        return (
+            <FieldRow key={id}>
+                <ReduxForm.Field
+                    component={Field}
+                    name={SAVED_SHIPPING_ADDRESS_FIELD}
+                    label={
+                        <strong className="u-text-semi-bold">{street}</strong>
```
jeffkamo: Same as my above comment: nice to have, thought due to time, I may skip on adding such a selector unless there is a strong reason to do this now.

web/app/utils/utils.js
jeffkamo: I'm struggling at the moment to recall why this seemed necessary... What I suppose I should do is actually remove this utility and replace its usages with fetching from the state.

jeffkamo:
@jnurse I'll start working on the default address now. As for improving the design, I'd call that polish that we can come back to at a later date. As it is now, it's at least functional.

web/app/utils/utils.js
jnurse: We need this util in a few other places, so I would suggest leaving it here. But maybe we don't need this logic of checking for a value, then setting it?

jeffkamo: Removing these checks has some knock on effects the `fetchShippingMethodsEstimate` to fail due to `shippingMethods` somehow becoming undefined 😱 

_\*backing away slowly\*_

web/app/containers/checkout-shipping/partials/shipping-address.jsx
```diff
@@ -6,226 +6,181 @@ import React from 'react'
...
+            postcode,
+            regionCode
+        } = address
+        const street = address.street.join(', ')
+        const shippingAddress = (
+            <div className="u-color-neutral-40">
+                <p className="u-margin-bottom-sm">
+                    {/* eslint-disable camelcase */}
+                    {city}, {regionCode}, {countryId}, {postcode}
+                    {/* eslint-enable camelcase */}
```
jnurse: These comments aren't needed anymore.

web/app/containers/checkout-shipping/partials/shipping-address.jsx
```diff
@@ -178,9 +180,7 @@ const mapStateToProps = createPropsSelector({
 })
 
 const mapDispatchToProps = {
-    handleShowAddNewAddress: (e) => {
-        return setShowAddNewAddress(e.target.value === ADD_NEW_ADDRESS_FIELD)
-    }
+    handleShowAddNewAddress: (bool) => setShowAddNewAddress(bool)
```
jnurse: [Non-Blocking] This can now be: 

`handleShowAddNewAddress: setShowAddNewAddress`

PR #565, opened 2017-05-02
Web-1221: Implement promo codes

gsaray:
Add promo to cart and payment step

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1221

## Changes
- Add promo to cart and payment step

## How to test-drive this PR
- Run `npm run dev`
- Add an item to cart 
- apply promo 'magic'

## Post Merge To Dos
- [ ] detect promo code on page load
- [ ] Make sure tax is consistent across checkout
- [ ]  add loader while promo is getting added


web/app/containers/cart/actions.js
```diff
@@ -22,42 +22,49 @@ import {addNotification} from '../app/actions'
...
-    return makeJsonEncodedRequest(getTotalsURL, requestData, {method: 'POST'})
-        .then((response) => response.json())
+    const totalInfo = getCartTotalsInfo(getState)
+    totalInfo
         .then((responseJSON) => {
             const cartTotals = {
                 subtotal: `$${responseJSON.subtotal.toFixed(2)}`,
                 subtotal_incl_tax: `$${responseJSON.subtotal_incl_tax.toFixed(2)}`,
                 tax_amount: `$${responseJSON.tax_amount.toFixed(2)}`,
+                base_grand_total: `$${responseJSON.base_grand_total.toFixed(2)}`
```
jnurse: We'll need to revisit this after the integration-manager branch is merged since we've changed a few of these key names.

web/app/containers/cart/actions.js
```diff
@@ -134,3 +141,74 @@ export const openRemoveItemModal = (itemId) => {
...
+    const couponCode = getCouponValue(currentState)
+    const putPromoUrl = `/rest/default/V1/${isLoggedIn ? 'carts/mine' : `guest-carts/${entityID}`}/coupons/${couponCode}`
+    return makeJsonEncodedRequest(putPromoUrl, {}, {method: 'PUT'})
+        .then((response) => {
+            if (response.status === 404) {
+                throw Error
+            }
+            return response.json()
+        })
+        .catch(() => dispatch(addNotification(InvalidPromoNotification)))
```
jnurse: I'm not sure this will have the behaviour you're looking for. After a catch the promise chain will continue on to getCartTotalsInfo. Is that what you're expecting? 

Maybe another option would be to just have one catch at the end of your promise chain that receives an error, and add the specific error message to the error you throw, then show that in the notification. Something a bit like this:

```
.then((response) => {
    if (response.status === 404) {
        throw new Error('Unable to apply promo, code is invalid')
    }
    return response.json()
})
// Your other thens will go here
.catch((error) => dispatch(addNotification({
    content: error.message,
    id: 'promoError',
    showRemoveButton: true
}))
```

web/app/containers/cart/actions.js
```diff
@@ -134,3 +141,74 @@ export const openRemoveItemModal = (itemId) => {
...
+        .then(() => getCartTotalsInfo(getState))
+        .then((responseJSON) => {
+            const totalsInfo = {
+                subtotal_with_discount: `$${responseJSON.subtotal_with_discount.toFixed(2)}`,
+                coupon_code: responseJSON.coupon_code,
+                tax_amount: `$${responseJSON.tax_amount.toFixed(2)}`,
+                discount_amount: `-$${responseJSON.discount_amount.toFixed(2).replace('-', '')}`,
+                shipping_amount: `$${responseJSON.shipping_amount.toFixed(2)}`,
+                base_grand_total: `$${responseJSON.base_grand_total.toFixed(2)}`
```
jnurse: This looks pretty similar to the parsing you're doing in `fetchTaxEstimate` I wonder if this could be refactored into a parser that both actions call, and have some logic in it that will determine if keys like `discount_amount` should be included.

web/app/containers/cart/actions.js
```diff
@@ -134,3 +141,74 @@ export const openRemoveItemModal = (itemId) => {
...
+    const RemovePromoErrorNotification = {
+        content: 'Unable to remove promo',
+        id: 'promoError',
+        showRemoveButton: true
+    }
+
+    const currentState = getState()
+    const isLoggedIn = getIsLoggedIn(currentState)
+    const entityID = getCustomerEntityID(currentState)
+    const putPromoUrl = `/rest/default/V1/${isLoggedIn ? 'carts/mine' : `guest-carts/${entityID}`}/coupons/`
```
jnurse: Should this maybe be `deletePromoUrl`?

web/app/components/order-total/index.jsx
```diff
@@ -0,0 +1,83 @@
...
+                    isTotal={true}
+                    value={subtotalInclTax}
+                />
+            )
+        }
+    }
+
+    return (
+        <Ledger className={classes}>
+            {totals()}
```
jnurse: I wonder if the logic for this could be moved somewhere else, maybe a selector, and this component just gets the value it should display and not all the different possible total keys?

web/app/containers/cart/partials/cart-summary.jsx
```diff
@@ -31,6 +45,55 @@ const CartSummary = ({summaryCount, subtotalExclTax, subtotalInclTax, shippingRa
...
+        <Button innerClassName="u-color-brand u-padding-start-0 u-text-letter-spacing-normal" onClick={removePromoCode}>
+            Remove Discount
+        </Button>
+    )
+
+    const showTax = () => {
+        const myReturn = []
+
+        // Will only display if tax is calculated via Shipping Estimation
+        if (shippingRate) {
```
jnurse: Could all of these checks go in the render's return statement instead? You wouldn't need to build out an array, you could just make all these LedgerRows only render if their check is true, so this one would be:

```
{shippingRate &&
    <LedgerRow
        label={`Shipping (${shippingLabel})`}
        value={shippingRate}
    />
}

```
It can get confusing to debug components that have these render helper functions if you're not super familiar with all the logic.

web/app/containers/cart/partials/cart-summary.jsx
```diff
@@ -31,6 +45,55 @@ const CartSummary = ({summaryCount, subtotalExclTax, subtotalInclTax, shippingRa
...
+                    className="u-flex-none"
+                    label="Taxes"
+                    labelAction="Rates based on shipping location"
+                    valueAction={calculateButton}
+                    key="taxWithCalculate"
+                />
+            )
+        }
+
+        return myReturn
```
jnurse: If you end up leaving this function in, could you rename this variable to something more descriptive, maybe `totalRows`?

web/app/containers/checkout-payment/partials/order-summary.jsx
```diff
@@ -22,8 +25,8 @@ import PaymentProductItem from './payment-product-item'
 // SDK Components
 import {Accordion, AccordionItem} from 'progressive-web-sdk/dist/components/accordion'
 import Button from 'progressive-web-sdk/dist/components/button'
-import Field from 'progressive-web-sdk/dist/components/field'
-import FieldRow from 'progressive-web-sdk/dist/components/field-row'
+// import Field from 'progressive-web-sdk/dist/components/field'
+// import FieldRow from 'progressive-web-sdk/dist/components/field-row'
```
jnurse: 🔫

web/app/containers/checkout-payment/partials/order-summary.jsx
```diff
@@ -92,11 +101,17 @@ class OrderSummary extends React.Component {
...
+                            <LedgerRow
+                                label={`Shipping (${shippingLabel})`}
+                                value={cartshippingRate}
+                            />
+                        :
+                            <LedgerRow
+                                label={`Shipping (${shippingLabel})`}
+                                value={shippingRate}
+                            />
+                        }
```
jnurse: I wonder if there's an opportunity to re-use some of these components and logic between the cart and payment pages.

web/app/containers/checkout-shipping/partials/shipping-method.jsx
```diff
@@ -41,7 +42,7 @@ const ShippingMethod = ({shippingMethods}) => {
                 })}
 
                 <FieldRow className="u-margin-top-lg">
-                    <Button type="submit" className="c--primary u-width-full u-text-uppercase qa-checkout__continue-to-payment">
+                    <Button type="submit" onClick={submitShipping} className="c--primary u-width-full u-text-uppercase qa-checkout__continue-to-payment">
```
jnurse: Is this needed? Isn't this function getting called when the form is submitted?

web/app/containers/checkout-payment/partials/order-summary.jsx
gsaray: I don't think we can re-use the logic, because the shipping rate is set to default in cart but we need to update shipping rate when applying 'freeship' coupon in payment step. we also have 'calculate tax' button in cart but not in payment step.

web/app/containers/cart/actions.js
gsaray: the first catch is for when user applies an invalid promo code. the second one is for when user applies a valid code but code is not applied successfully.

jeffkamo:
Closing in favor of #596

PR #567, opened 2017-05-04
Fix - PDP Product Variant selection requires two clicks

Cdok:
# PDP Product Variant selection requires two clicks

The blur event triggered an update of the product images if the user selects a variation, and blurs off of the `<select>`

This PR changes the `onBlur` handlers to an `onChange` for when
the user selects a different variation of an item on the PDP

Prior to this, if a user changed their "item variations" and immediately
clicked "add to cart", the blur event would instead fire, preventing the add
to cart - thus requiring a second click

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-66

## Changes
- Switched `onBlur` event to update image to `onChange` on PDPs

## How to test-drive this PR
- Open `app/main.jsx`.
- Import the Demandware connector:
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1
)
- Navigate to a PDP with multiple variations
- Select a variation, and click add to cart immediately

### DIFF
--- a/web/app/containers/product-details/actions.js
+++ b/web/app/containers/product-details/actions.js
@@ -73,18 +73,18 @@ export const submitCartForm = (formValues) => (dispatch, getStore) => {
         .then(() => dispatch(addToCartComplete()))
 }
 
-const variationBlurSelector = createPropsSelector({
+const variationChangeSelector = createPropsSelector({
     variationSelections: getAddToCartFormValues,
     categoryIds: getProductVariationCategoryIds,
     variants: getProductVariants
 })
 
-export const onVariationBlur = () => (dispatch, getStore) => {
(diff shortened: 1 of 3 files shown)


Cdok:
@jnurse That's a good idea, it might be a bit jarring for the user that the page updates as soon as they click the option. I'm pushing up a change now

Let me know if this is what you were thinking, or if we could maybe move the timeout to a better spot

PR #569, opened 2017-05-04
[Integration Manager] Connector configuration

jeremywiebe:
This PR sets up configurability of the connectors. Connectors are now constructed (`new`) and the configuration is passed into the contructor. Each connector will have it's own set of config it receives. 

I've created a `defaultConfig` object which shows the supported config options and default values, if applicable. There might be a better way to model that. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-72
 **Linked PRs**: N/A

## Changes
- Convert connectors to have a constructor
- `main.jsx` now passes config to connector constructor
- Connectors look at `_connector/index.js`'s `config` export for current configuration values

## How to test-drive this PR
- Run `npm run dev`
- Test both the Merlin's and Demandware connectors


web/app/integration-manager/_demandware-connector/constants.js
```diff
@@ -1,12 +1,21 @@
...
-export const SIGN_IN_URL = `/on/demandware.store/${SITE_ID}/default/Account-Show`
-export const CHECKOUT_SHIPPING_URL = `/on/demandware.store/${SITE_ID}/default/COShipping-Start`
-export const CART_URL = `/on/demandware.store/${SITE_ID}/default/Cart-Show`
+
+const BASE_PAGE_URL = `/on/demandware.store/${SITE_ID}/default/`
+
+export const HOME_URL = `${BASE_PAGE_URL}/Home-Show`
+export const SIGN_IN_URL = `${BASE_PAGE_URL}/Account-Show`
+export const CHECKOUT_SHIPPING_URL = `${BASE_PAGE_URL}/COShipping-Start`
+export const CART_URL = `${BASE_PAGE_URL}/Cart-Show`
```
drewzboto: We will have future work to pull some of theses URLs out? The routes are going to be implementation specific, so I'm not sure if they really belong in the connector vs a scaffold

drewzboto:
Other piece to look at configuring is the servicer-worker script URL. In the SFCC integration cartridge, we serve it without the .js extension and we always have to change loader.js when swapping from Magento to SFCC

web/app/integration-manager/_demandware-connector/index.js
```diff
@@ -1,6 +1,22 @@
+import assign from 'lodash.assign'
```
DanielBrookRoberge: We should be able to use `Object.assign` here instead, since we have it polyfilled to support the object spread operator.

web/app/main.jsx
```diff
@@ -39,7 +39,11 @@ analyticManager.init({
 }, clientAnalytics, pushMessaging)
 initCacheManifest(cacheHashManifest)
 
-registerConnector(connector)
+registerConnector(new Connector({
+    /* SFCC Configuration */
+    siteID: '2017refresh',
+    clientID: '5640cc6b-f5e9-466e-9134-9853e9f9db93'
+}))
```
DanielBrookRoberge: I'm not sure of the utility of using `new` here; `Connector` is just a factory function anyways. I'm always nervous about the edge cases of the JS `new` operator.

web/app/integration-manager/_demandware-connector/index.js
jeremywiebe: @DanielBrookRoberge I've made the change. The only other place `lodash.assign` is used is in `webpack/production.js`  Do you see any need for it there either?  I switched it to `Object.assign` and run` npm run prod:build` and it seemed to work.

Cdok:
This branch has been updated with the new configuration setup, which introduces a `registerConfig()` function which is called when a connector is registered.

The config is now stored in the scope of the `constants` module, and each of the constants has been updated to return the appropriate value from the config.

web/app/integration-manager/_sfcc-connector/checkout/commands.js
```diff
@@ -214,7 +214,7 @@ export const submitPayment = (formValues) => (dispatch) => {
             dispatch(receiveOrderConfirmationContents({
                 orderNumber: order.order_no
             }))
-            return `/on/demandware.store/${SITE_ID}/default/COSummary-Submit`
+            return `/on/demandware.store/${getSiteID()}/default/COSummary-Submit`
```
DanielBrookRoberge: We might want to make this a `getConfirmationUrl` function in `constants.js`.

web/app/integration-manager/_sfcc-connector/constants.js
```diff
@@ -2,18 +2,32 @@
 /* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
 /* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
 
+let config = {}
+
+export const registerConfig = (cfg) => {
+    config = cfg
+}
```
DanielBrookRoberge: What do you think of renaming this from `constants.js` to `config.js`?

web/app/integration-manager/_sfcc-connector/parsers.js
```diff
@@ -124,7 +124,7 @@ export const parseCategories = (categories) => {
     return categories.map((category) => {
         return {
             title: category.name,
-            path: `/s/${SITE_ID}/${category.id}`,
+            path: `/s/${getSiteID()}/${category.id}`,
```
DanielBrookRoberge: We're duplicating this category path generation, it should probably also move to the constants/config file

web/app/integration-manager/_sfcc-connector/constants.js
Cdok: I think that's a good idea considering the file is no longer just a collection of `const` variables 👍

web/app/integration-manager/_sfcc-connector/config.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const getCategoryPath = (id) => `/s/${config.siteID}/${id}`
+
+export const getHomeURL = () => `/on/demandware.store/${config.siteID}/default/Home-Show`
+export const getSignInURL = () => `/on/demandware.store/${config.siteID}/default/Account-Show`
+export const getSignOutURL = () => `/on/demandware.store/${config.siteID}/default/Logout-Logout`
+export const getCheckoutShippingURL = () => `/on/demandware.store/${config.siteID}/default/COShipping-Start`
+export const getCartURL = () => `/on/demandware.store/${config.siteID}/default/Cart-Show`
+export const getPaymentURL = () => `/on/demandware.store/${config.siteID}/default/COBilling-Start`
+export const getConfirmationURL = () => `/on/demandware.store/${config.siteID}/default/COSummary-Submit`
```
DanielBrookRoberge: You could write a `getStoreURL` or similar function to DRY out the first part of these addresses (up to `/default/` probably), which will make it easier to modify this further if necessary

web/app/main.jsx
```diff
@@ -27,19 +27,22 @@ import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
...
-import connector from './integration-manager/_sfcc-connector'
+import {Connector} from './integration-manager/_merlins-connector'
+// import {Connector} from './integration-manager/_sfcc-connector'
 
 import {registerConnector} from './integration-manager'
 
+registerConnector(Connector({
+    siteID: '2017refresh',
+    clientID: '5640cc6b-f5e9-466e-9134-9853e9f9db93'
+}))
```
DanielBrookRoberge: Maybe add a comment about how it's ok to pass this to the Merlin's connector so we're just leaving this here all the time for now.

web/app/integration-manager/_sfcc-connector/config.js
```diff
@@ -0,0 +1,33 @@
...
+export const getApiEndPoint = () => `/s/${config.siteID}/dw/${API_TYPE}/${API_VERSION}`
+
+export const getRequestHeaders = () => ({
+    'Content-Type': 'application/json',
+    'x-dw-client-id': config.clientID
+})
+
+export const getCategoryPath = (id) => `/s/${config.siteID}/${id}`
+
+export const getBaseURL = () => `/on/demandware.store/${config.siteID}/default/`
```
jeremywiebe: This isn't used outside of this file. I think it'd be good to not export it in this case.

jeremywiebe: Also, I think this URL is not being built quite right.  It should be `/on/demandware.store/Sites-${config.siteID}-Site/default/`

web/app/integration-manager/_sfcc-connector/parsers.js
```diff
@@ -43,7 +43,7 @@ const setInitialVariantValues = (variants, id, variationCategories) => {
 
 /* eslint-enable camelcase */
 
-export const getProductHref = (productID) => `/s/2017refresh/${productID}.html`
+export const getProductHref = (productID) => `/s/${getSiteID()}/${productID}.html`
```
jeremywiebe: It's like whack-a-mole.  This one doesn't have the Site ID sandwiched between `Sites-${SiteID}-Site`.  I think `getSiteID()` should return the raw site and then you can build the site path with it separately.

Cdok: Pushed up the change reverting this - but is there a reason this one doesn't have the `Sites-______-Site` sandwich? I can see that previously it wasn't in `getProductHref`, but it seems like even for the sandbox instance on the SFCC documentation that it just says to use `/s/site_id/`.

Is our `site_id` the full string or just `2017refresh`?

After testing it, it doesn't seem like there are any adverse side effects to passing just `2017refresh` vs the `Sites-2017refresh-Site` in `getProductHref`

jeremywiebe: 🤔   The site ID is definitely just `2017refresh`.  If docs confirm what you say, let's drop the `Sites-______-Site` sandwich where we can!

Cdok: @jeremywiebe I've removed all of the `Sites-_______-Site` occurrences and after a quick pass there doesn't seem to be any regressions 🤔  

Do we know where the `Sites-_______-Site` originally came from? I wouldn't want to cause any regressions if there was a reason this was in our URLs

jeremywiebe: I suspect that came from watching what the Desktop site does... If it works and conforms to the docs, let's drop the `sites` sandwich.

Cdok: 🚫 🍔  site sandwich is dropped on this branch

jeremywiebe: I can't 👍  this PR as I'm the original author.. DBR already approved it so let's merge it!! :D

PR #573, opened 2017-05-08
Don't use merlins for package name

kerrmarin:
Changes some of the package descriptors for Android to enable the Generator to change the package.

 **JIRA**: https://mobify.atlassian.net/browse/PLATT-221

## Changes
- Change the package references from `merlins` to `scaffold`

## How to test-drive this PR
- `cd native`
- `npm run deps`
- Run the app in Android.
- The app should run and work 👍 

OR

- Deploy the build from buddybuild to a phone
- Run the app, should work 👍


native/ios/scaffoldTests/ProjectTests.swift
```diff
@@ -1,15 +1,15 @@
 //
-//  scaffoldTests.swift
-//  scaffoldTests
+//  ProjectTests.swift
+//  ProjectTests
 //
-//  Created by Mike Klemarewski on 2015-06-30.
+//  Created by Mobify on 2015-06-30.
```
jeremywiebe: I'm not sure this matters... but meh. :)

PR #577, opened 2017-05-09
Create a Test for Bundle size Analyzer

jasecode:
It'd be really cool if CircleCI would check for bundle size drift and fail the build if the bundle size changed more than a certain amount between builds.  
- Bonus points if we also track the size of `loader.js` for bloat over time as well. " 

## Changes
- webpack/base.main.js
-- Add webpackbundlesizeanalyzer plugin
-- Add json export out of the webpack bundle analyzer (different from above!)
- Examine each file in stats.json and verify it is below 2 MB

## How to test-drive this PR
- npm run test:stats
- npm run test:size

## To do:
- Find a way to compare json files from the current bundle to the one being developed


web/scripts/lighthouse-ci.sh
```diff
@@ -19,19 +19,18 @@ trap 'kill $(jobs -pr)' SIGINT SIGTERM EXIT
...
+	mkdir -p $HOME/.pki/nssdb
+	# Pass in a password
+	certutil -d $HOME/.pki/nssdb -N --empty-password
+	# Add self-signed SSL certificate
+	certutil -d sql:$HOME/.pki/nssdb -A -t "P,," -n dev-server/localhost.pem -i dev-server/localhost.pem
+fi
 
 # --ignore-certificate-errors thanks to https://github.com/GoogleChrome/lighthouse/issues/559
-sleep 5
+while ! echo exit | nc localhost 8443; do sleep 5; done
```
jasecode: A sleep 5 maybe too ambitious. Maybe lets try 10 seconds?

jasecode:
Closing as We moved our approach to this via: https://github.com/mobify/platform-scaffold/pull/746

PR #578, opened 2017-05-09
Basic AMP template

kschmidtdev:
Stubs out some of the folder structure for AMP, start of a `package.json`, include a basic express server (to be replaced with upcoming features in the mobify-amp-sdk)

 **JIRA**: https://mobify.atlassian.net/browse/AMP-31

## Changes
- Add basic React template that renders valid AMP
- Add start of `package.json` that links up to the web folder as well as `mobify-amp-sdk`
- Add a simple express server (to be replaced soon)

## How to test-drive this PR
- `cd amp`
- `npm i`
- `npm run dev`
- Check out http://localhost:3000/


amp/app.js
```diff
@@ -0,0 +1,33 @@
...
+router.get('*', (req, res) => {
+    const PUBLIC_FOLDER = path.join(__dirname, `app/containers/test-container`)
+
+    const appString = renderToString(<TestContainer/>)
+    const baseCSS = fs.readFileSync(`${PUBLIC_FOLDER}/container.css`).toString()
+
+    res.send(template({
+        body: appString,
+        title: 'Hello World from the server',
+        includeBaseCSS: baseCSS,
```
cole-sanderson: We have decided to change term of 'base' to 'global' to avoid confusion with 'base' styles in base folder in web directory. So it will make sense to change to 'global' instead of 'base'.

kschmidtdev:
@olibrook is there anything else that should go in this before a 👍 ?

olibrook:
@kschmidtdev I've just re-opened a PR over here which combines both of our work and adds a couple of extra features we need for AMP rendering (including AMP's JS per-component, for instance). That's what we briefly discussed on Thursday, if I remember right.

https://github.com/mobify/platform-scaffold/pull/594

I think it includes everything you had here. Can we merge that one instead?

kschmidtdev:
This functionality will arrive via #594

PR #579, opened 2017-05-09
[Integration Manager] Checkout Schema Part 1

DanielBrookRoberge:
This PR contains the first half of the checkout schema work. It has become large enough that it could use to be split up into smaller PRs.

## Changes
- Add types for the full checkout schema
- Convert the location logic
- Convert the shipping methods
- Switch to the new recommended method of connecting the Redux dev tools
- Clean up connector code in preparation for schematization.

## How to test-drive this PR
- Preview the checkout in Merlins and DW
- Ensure that it is still completable as far as is implemented in the base branch.


web/app/integration-manager/_merlins-connector/cart/commands.js
```diff
@@ -122,16 +122,11 @@ const ESTIMATE_FIELD_PATH = ['#block-summary', 'Magento_Ui/js/core/app', 'compon
...
-            const locationsData = parseLocations(magentoFieldData)
 
-            return dispatch(receiveCheckoutData({
-                customerEntityID,
-                ...locationsData
-            }))
+            dispatch(receiveCheckoutLocations(parseLocations(magentoFieldData)))
+            dispatch(receiveEntityID(parseCheckoutEntityID($response)))
         })
         .then(() => dispatch(fetchShippingMethodsEstimate(ESTIMATE_FORM_NAME)))
```
jeremywiebe: You've got other changes from my `fix-checkout` branch.  Do you want to move this `dispatch` into the previous block and get rid of the extra `.then()` here?

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -103,8 +106,35 @@ export const fetchCheckoutConfirmationData = (url) => (dispatch) => {
         })
 }
 
+const addressFormToAddressObject = ({
+    firstname,
```
jeremywiebe: I'm wondering if this is the most clear way to represent this. We're basically doing a simple mapping operation. Spreading the input and then building the result object doesn't provide a strong visual tie between inputs and outputs... Would this be clearer (not 100% sure, but it is definitely more compact)?

```js
const addressFormToAddressObject = (formValues) => ({
    firstName: formValues.firstName,
    lastName: formValues.lastName,
    addressLine1: formValues.addressLine1,
    addressLine2: formValues.addressLine2,
    countryId: formValues.countryId,
    city: formValues.city,
    regionId: formValues.regionId,
    region: formValues.regionId,
    postcode: formValues.postcode,
    telephone: formValues.telephone,
    saveInAddressBook: true
})
```

jeremywiebe: I kinda like the idea of providing a mapping layer from our Redux form values to our "typed" objects.  What do you think of adding a typecheck on the returned object to help catch regressions as our code changes over time?

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -113,23 +143,12 @@ export const submitShipping = (formValues) => (dispatch, getState) => {
...
-            city: newAddress.city,
-            street: newAddress.addressLine2
-                ? [newAddress.addressLine1, newAddress.addressLine2]
-                : [newAddress.addressLine1],
-            regionId: newAddress.regionId,
-            region: newAddress.region,
-            countryId: newAddress.countryId,
-            saveInAddressBook: true
-        }
+            ...formValues
```
jeremywiebe: I'm a bit paranoid about using spread operators for mapping operations. If the Address object changes and no longer matches what Merlin's expects, we break.  The key names will never change unless we change what version of Magento we're using.

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -178,70 +190,56 @@ export const submitShipping = (formValues) => (dispatch, getState) => {
...
+// Some of the endpoints don't work with fetch, getting a 400 error
+// from the backend. This function wraps the jQuery ajax() function
+// to make requests to these endpoints.
+//
+// It looks like the server may be looking for the header
+// X-Requested-With: XMLHttpRequest, which is not present with fetch.
+//
+// Alternatively, we could have an issue with header case:
+// http://stackoverflow.com/questions/34656412/fetch-sends-lower-case-header-keys
+const jqueryAjaxWrapper = (options) => {
```
jeremywiebe: 🎉

web/app/integration-manager/_sfcc-connector/checkout/commands.js
```diff
@@ -7,74 +7,68 @@ import {createBasket} from '../cart/utils'
...
+            /* eslint-enable camelcase */
+            dispatch(receiveShippingMethodInitialValues({initialValues}))
+            dispatch(receiveCheckoutLocations({
+                countries: [
+                    {id: 'us', label: 'United States', regionRequired: true, postcodeRequired: true}
+                ],
+                regions: STATES
+            }))
         })
+        .then(() => dispatch(fetchShippingMethodsEstimate()))
```
jeremywiebe: This can also move into the previous block, I think.

web/app/store/checkout/actions.js
```diff
@@ -6,3 +6,4 @@ import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
 
 export const receiveSavedShippingAddresses = createAction('Receive Saved Shipping Addresses', 'savedAddresses')
 
+export const receiveCheckoutData = createAction('Receive Checkout Data')
```
jeremywiebe: Is this still required?

web/app/store/index.js
```diff
@@ -17,9 +18,6 @@ import {reducer as formReducer} from 'redux-form'
 import analytics from 'redux-analytics'
 import {analyticManager} from 'progressive-web-sdk/dist/analytics/analytic-manager'
 
-const noop = (f) => f
```
jeremywiebe: \+ 10000

web/app/utils/utils.js
```diff
@@ -68,9 +68,11 @@ export const typecheck = (type, value) => {
     return value
 }
 
-export const createTypedAction = (description, type) => createReduxAction(
+export const createTypedAction = (description, type, key) => createReduxAction(
```
jeremywiebe: This is somewhat voodoo-ish.  Would you mind adding a JSDoc comment explaining what the new `key` does?

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: Yeah, worth a thought. I'll sit on this for a bit (since we're not going to get this merged today)

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: This is being passed to the `addressFormToAddressObject` function; I'm just tacking on some fields before passing it in.

web/app/store/checkout/actions.js
DanielBrookRoberge: It's still used by the checkout shipping container. I haven't had a chance to investigate yet.

web/app/store/index.js
DanielBrookRoberge: yeah, it's confusing because we have another `noop` that is `() => {}`.

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: I think that your suggestion is the better choice, after reflecting on it. I'll make the change.

jnurse:
@DanielBrookRoberge With both connectors I'm not seeing the region drop down get populated correctly after I select my country (United States) on the shipping page I also can't continue past the shipping page.

DanielBrookRoberge:
Switching back to WIP because I haven't had a chance to test at all since the last round of merges.

jeremywiebe:
Checkout with the Merlin's connector still doesn't work. :(

image

That's the error we get back. Here's the full request:

```json
{
    "billingAddress": {
        "firstname": "Jeremy",
        "lastname": "Wiebe",
        "company": "",
        "postcode": "90210",
        "city": "Vancouver",
        "street": [
            null
        ],
        "regionId": "12",
        "countryId": "US",
        "saveInAddressBook": false
    },
    "cartId": "2571",
    "paymentMethod": {
        "additional_data": null,
        "method": "checkmo",
        "po_number": null
    }
}
```

DanielBrookRoberge:
Closing this PR since everything in it has been redone.

PR #580, opened 2017-05-10
Fix PDP variations dropdown on SFCC Connector

Cdok:
# Fix PDP variation selection

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-67

## Changes
- When switching between variations on a PDP, the user will no longer be scrolled to the top of the PDP
- Stores the selected variation URL in the app state, and instead of using `browserHistory.push()` to navigate to a new variant, `fetchPdpData` is called to refresh the data on the PDP
- Adds a default option in the `<select/>` element of variant dropdowns
- This PR fixes the correct variants not being selected when clicking `Edit Cart`

## How to test-drive this PR
- Use `demandware-connector`
- Run `npm run dev`
- Navigate to a PDP with multiple variants (Womens -> Lace trimmed cardigan)
- Select options, and ensure the page does not scroll to the top
- Add to cart, and proceed to edit cart, ensuring that the proper variant is prepopulated


Cdok:
A note about the most recent commits, `setInitialVariantValues` and `fetchPdpData` have been changed to fix master products values not pre-populating.

This feels a bit hacky, since the master product will load, and we check if the master product is selected - if it is, `fetchPdp` will be called again, and we instead set the current selected product to the first variation.

The alternative to this is having the default option be `Please select a ______` on the master product page.

I'm open to ideas for how to tackle this one 🤔

web/app/integration-manager/_demandware-connector/products/commands.js
```diff
@@ -26,22 +26,31 @@ export const fetchPdpData = () => (dispatch) => {
...
+
+            // since the pathname will always be master, the productHref will
+            // only === pathname when landing on master page
+            if (getProductHref(productDetailsData.id) === window.location.pathname) {
+                const {variants, initialValues} = productDetailsData
+                const defaultVariant = getInitialSelectedVariant(variants, initialValues)
+                const currentProductHref = defaultVariant.id
+
+                dispatch(setCurrentURL(getProductHref(currentProductHref)))
+                dispatch(fetchPdpData(getProductHref(currentProductHref)))
```
Cdok: This is the fallback for master products that recursively calls `fetchPdpData` with the first variation's.

@jeremywiebe @DanielBrookRoberge - thoughts on this? Jen and I spoke with Mitch and he leaned more towards the `Please select a <variant>` default option, especially because this might have the side effect of the master product and first variant's thumbnails being different (shopper might see the images briefly flash)

Cdok:
The current implementation on this branch does not prepopulate the dropdowns on the _Master Product_ page.

To populate these fields on the Master product with the current implementation, we would have to find the first variant after landing on the master page, and dispatch `fetchPdp` again with the default first variant as the parameter.

This was tested at first, and it caused a flash of the thumbnail. Potentially, if the master product had a _different_ image than the first variant, the changing thumbnail would be even more apparent.

I'm open to feedback for a different approach to the above problem. Jen and I spoke with Mitch to get a design perspective, and he recommended having the fields populated with `Please select a color/size/etc`

drewzboto:
Can we discuss the constraints here? As per the story, I'm most interested in reducing the number of unnecessary clicks, which adding a default option doesn't seem to do. 

Ideal scenario is that this would select the default variant (which is configurable in SFCC). However, I believe OCAPI doesn't support this OOTB and we would have to extend OCAPI. See https://xchange.demandware.com/message/69349#69349

I see two paths here:
1. Correctly select the default variant in Merlin's on Magento (once we complete the story to add variants)
2. Get Astound to add default to OCAPI and support this in the future

Cdok:
Just an update on the current state of this PR:

- The initial values are prepopulating, but this solution feels less than ideal / a bit hacky.
- the first variation is fetched after landing on a master product page

If there is a better way to handle the initial value I'm definitely open to some suggestions. Jen and I went over a couple of approaches, and it seemed like none of them would be as clean as retrieving the default from SFCC.

PR #582, opened 2017-05-10
[Integration manager] Remove duplicate login and register commands

jnurse:
This PR removes the checkout login & registration commands so we have just one command each for logging in and registering a user. 

 **JIRA**: n/a
 **Linked PRs**: n/a

## Changes
- Updated login and registration commands to accept arguments instead of a form values object
- Updated actions in the app that call the login & registration commands to pass correct data to the commands

## How to test-drive this PR
- Run `npm run dev`
- open the menu and click "sign in"
- try signing in with valid credentials
- log out
- navigate back to the sign in page
- try creating a user with valid credentials
- log out
- add an item to your cart and proceed to the checkout shipping page
- log in with valid credentials from that page
- log out
- add an item to your cart and proceed through the checkout to the order confirmation page 
** make sure you use an email address that doesn't have an account associated with it **
- enter a password & confirmation password on the page and create an account


web/app/integration-manager/_merlins-connector/login/commands.js
```diff
@@ -2,28 +2,26 @@ import {makeRequest, makeFormEncodedRequest} from 'progressive-web-sdk/dist/util
...
 import {setSigninLoaded, setRegisterLoaded} from '../../login/results'
 import {receiveNavigationData} from '../../results'
 import {parseNavigation} from '../navigation/parser'
+import {LOGIN_POST_URL, CREATE_ACCOUNT_POST_URL} from '../constants'
 
-import {isFormResponseInvalid, parseSigninHref, parseRegisterHref} from './parsers/parsers'
+import {isFormResponseInvalid} from './parsers/parsers'
 
 export const fetchSigninData = (url) => (dispatch) => {
     return dispatch(fetchPageData(url))
```
DanielBrookRoberge: Now that we're not parsing out the URLs, do we even need to do these fetches? We're not using the result for anything.

web/app/containers/login/actions.js
```diff
@@ -94,7 +94,15 @@ export const submitSignInForm = (formValues) => (dispatch) => {
         return Promise.reject(new SubmissionError(errors))
     }
 
-    return dispatch(login(formValues))
+    const {
+        login: {
+            username,
+            password
+        },
```
DanielBrookRoberge: I don't know that this is in scope here, but with the new structure no longer needing to reflect the shape of the form on Merlin's desktop we could flatten this.

web/app/integration-manager/_merlins-connector/login/commands.js
jnurse: Good catch 😄

jnurse: Actually after taking another look at this, we need to leave this fetch in. It's the global fetch that updates a bunch of app data.

PR #584, opened 2017-05-11
[Integration Manager] Command audit

jeremywiebe:
This PR makes all IM command parameters explicit as well as fixes some outstanding bugs related to that. 

 **JIRA**: [**WEBDATA-17**: Finalize Command API](https://mobify.atlassian.net/browse/WEBDATA-17)
 **Linked PRs**: N/A

## Changes
- Removes the `...args` parameter style for all IM commands and replaces them with the actual list of arguments each command expects
- Adds JSDocs to every IM command
- Centralizes the `XYZ_FORM_NAME` constants as they are used by both selectors and `/app/containers/...`
- Renamed all `getXyxPage` commands to `initXyzPage` as not all commands need to fetch data

## How to test-drive this PR
- Run `npm run dev`
- Make sure `main.jsx` has the Merlin's connector configured
    ```js
    import connector from './integration-manager/_merlins-connector'
    import {registerConnector} from './integration-manager'
    ````
- Preview [Magento Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- **Verify**: Make sure all features work (basic smoke test through checking and `npm run test:e2e` should be enough
- Switch `main.jsx` to use the SFCC connector
    ```js
    import connector from './integration-manager/_demandware-connector'
    import {registerConnector} from './integration-manager'
    ````
- Preview [SFCC Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- **Verify**: Make sure all features work (basic smoke test through checking and `npm run test:e2e` should be enough


web/app/containers/navigation/reducer.js
```diff
@@ -17,9 +17,9 @@ export const reducer = handleActions({
     [setNavigationPath]: mergePayload,
     [setLoggedIn]: (state, {payload: {isLoggedIn}}) => {
         return state.setIn(
-            ['root', 'children', 0, 'title'],
-            isLoggedIn ? ACCOUNT_LINK_TEXT : SIGN_IN_LINK_TEXT
-        )
+                ['root', 'children', 0, 'title'],
+                isLoggedIn ? SIGN_OUT_LINK_TEXT : SIGN_IN_LINK_TEXT
+            )
```
DanielBrookRoberge: Shouldn't this also change the `type`? In the parser sign in and sign out have different types.

web/app/integration-manager/_demandware-connector/app/commands.js
```diff
@@ -34,7 +41,7 @@ export const initApp = () => (dispatch) => {
         .then(() => {
             dispatch(setCheckoutShippingURL(CHECKOUT_SHIPPING_URL))
             dispatch(setCartURL(CART_URL))
-            return dispatch(setLoggedIn(utils.isUserLoggedIn(utils.getAuthToken())))
+            dispatch(setLoggedIn(utils.isUserLoggedIn(utils.getAuthToken())))
         })
         .then(() => dispatch(getCart()))
```
DanielBrookRoberge: If we're not serializing with `return` here (which I think is the right choice, then this can just be consolidated into the previous callback.

web/app/integration-manager/_demandware-connector/cart/commands.js
```diff
@@ -96,15 +96,13 @@ export const addToWishlist = (productId) => (dispatch) => {
...
-
-
         })
         .then(({id}) => {
             const requestOptions = {
                 method: 'POST',
                 body: JSON.stringify({
                     type: 'product',
-                    product_id: productId,
+                    product_id: productID,
```
DanielBrookRoberge: We should probably talk about this; I've been trying to consistently name things with acronyms capitalized as words (`productId`, `cartUrl`, etc.), but if we have a standard for doing things the other way then I should change.

web/app/integration-manager/_demandware-connector/checkout/commands.js
```diff
@@ -133,9 +133,11 @@ export const submitShipping = (formValues) => (dispatch) => {
         })
 }
 
+export const updatingShippingAndBilling = () => () => Promise.resolve()
```
DanielBrookRoberge: Like above, should this be `updateShippingAndBilling`?

web/app/integration-manager/_demandware-connector/login/commands.js
```diff
@@ -88,6 +88,29 @@ export const login = ({login}) => (dispatch) => {
...
+                console.error('Error logging out. Clearing auth tokens anyways.', response.json())
+            }
+
+            deleteBasketID()
+            deleteAuthToken()
+            dispatch(setLoggedIn(false))
+
+            // This isn't great, but it's the only _simple_ way to refresh the
+            // navigation menu to show the Account status correctly.
+            return dispatch(fetchNavigationData())
```
DanielBrookRoberge: If the reducer on `setLoggedIn` for the navigation branch is operating correctly, we shouldn't need this. Otherwise we don't need that reducer.

web/app/store/checkout/locations/selectors.js
```diff
@@ -16,12 +16,24 @@ export const getDefaultRegionEntry = createGetSelector(getRegions, 0)
...
+    getRegionsForCountry(formKey),
+    getDefaultRegionEntry,
+    (regions, defaultRegion) => {
+        if (regions.size !== 0 && !defaultRegion.equals(regions.get(0))) {
+            regions.unshift(defaultRegion)
+        }
+
+        return regions
+    }
+)
```
DanielBrookRoberge: The use of `formKey` as the form identifier here keeps confusing me with the `formKey` that's in the Merlin's connector (that's a CSRF token). It's not really in the scope of this PR but I think this should be `formName` instead.

web/app/containers/login/container.jsx
```diff
@@ -25,11 +25,11 @@ class Login extends React.Component {
...
         super(props)
 
-        this.navigateToSection = this.navigateToSection.bind(this)
+        this.onChangeTab = this.onChangeTab.bind(this)
     }
 
-    navigateToSection(index) {
-        this.props.navigateToSection(
+    onChangeTab(index) {
+        this.props.navigatedToSection(
```
jeremywiebe: I was initially thinking the same thing. But in this case, the navigation doesn't seem to actually happen in the connector method. We do a `router.push()` in the connector, but when I commented it out, it still swapped views, it just didn't update the URL.

web/app/integration-manager/checkout/commands.js
```diff
@@ -4,11 +4,63 @@ export const register = (commands) => {
...
+ * @param {string} formName The stage's form name to estimate shipping on
+ */
+export const fetchShippingMethodsEstimate = (formName) => connector.fetchShippingMethodsEstimate(formName)
+
+/**
+ * Updates the registered customer's billing and shipping addresses using
+ * the address that was given during checkout. This should be called during
+ * checkout once shipping information has been provided.
+ */
+export const updatingShippingAndBilling = () => connector.updatingShippingAndBilling()
```
jeremywiebe: This should probably be `updateShippingAndBillingAddresses`

web/app/integration-manager/_sfcc-connector/checkout/commands.js
```diff
@@ -133,6 +133,8 @@ export const submitShipping = (formValues) => (dispatch) => {
         })
 }
 
-// We're not currently checking the customer's email on the sfcc site
+export const updatingShippingAndBilling = () => () => Promise.resolve()
```
jnurse: I'm looking at implementing this command on SFCC, and it seems like their system doesn't distinguish between a saved shipping or billing address. What do you think about separating these into `updateUsersShippingAddress` and `updateUsersBillingAddress` or something similar? 

It's probably easiest if I do that work in the confirmation page branch.

jeremywiebe: Definitely think that's useful, yes. I think the `Users` part of those commands is implied in almost all our commands... not sure it's needed or not.

jnurse: I thought calling out that you're setting a users address might prevent some confusion about if the address is being set for your checkout session, or your account, but maybe it's implied already if we put it in the login/account section of the connector.

web/app/integration-manager/_demandware-connector/cart/commands.js
jeremywiebe: I agree. I thought I was following the pattern others had used in the code, but we obviously have a mixture. I've made a task to add that to our `mobify-code-style` so we can discuss it in a PR, settle on one, and then move on. :)

web/app/integration-manager/_sfcc-connector/login/commands.js
```diff
@@ -111,8 +129,9 @@ export const registerUser = ({firstname, lastname, email, password}) => (dispatc
             if (responseJSON.fault) {
                 throw new SubmissionError({_error: 'Unable to create account.'})
             }
+
             // Creating a user doesn't sign them in automatically, so dispatch the login command
-            return dispatch(login({login: {username: email, password}}))
+            return dispatch(login(email, password, true))
```
jnurse: Should the login function above (line 25) be getting three arguments? Looks like it only expects a username and password.

jeremywiebe: Hm. That opens up an interesting question. We're dispatching the connector implementation of the `login` integration manager command. The IM command takes three parameters: `username`, `password`, and `rememberMe`. 

The big question we should discuss is if we should go back through the IM command when we leverage existing IM commands from within a connector. I am leaning towards "yes", otherwise command overriding/replacement won't work...

DanielBrookRoberge: Yeah, we'll need to do that I think. At this point I think that should be a separate PR. 

If things break because we're bouncing out to the IM, that's probably uncovering a bug.

jnurse: Yeah I agree we might want to look at doing that, but probably in a separate PR.

web/app/containers/navigation/reducer.js
```diff
@@ -19,16 +19,23 @@ export const reducer = handleActions({
...
+
+        return state
+            .setIn(
+                [...accountNodePath, 'title'],
+                isLoggedIn ? SIGN_OUT_LINK_TEXT : SIGN_IN_LINK_TEXT
+            )
+            .setIn(
+                [...accountNodePath, 'type'],
+                isLoggedIn ? SIGNED_IN_NAV_ITEM_TYPE : GUEST_NAV_ITEM_TYPE
+            )
```
DanielBrookRoberge: If we define objects with these pairings, i.e.:
```javascript
const LOGGED_IN_NAV = {
    title: SIGN_OUT_LINK_TEXT,
    type: SIGNED_IN_NAV_ITEM_TYPE
}
```

we can simplify all of this to 
```javascript
return state.mergeDeepIn(accountNodePath, isLoggedIn ? LOGGED_IN_NAV : GUEST_NAV)
```

PR #585, opened 2017-05-11
Push Integration Redux: Push Harder

fractaltheory:
Adds a simple 'connector' component and Redux actions that communicate with MessagingClient, as well as a prototype soft ask UI component.

## Changes
- Track page counts in app state
- Track external state from MessagingClient in Redux store
- Add prototype localStorage persistence and rehydration to Redux store
- Add SoftAsk component for displaying a soft ask
- Add push messaging selectors, actions, and reducer

## TODO
- Move _a ton_ of this to the SDK when I'm happy with it
- A bunch of TODOs in the source

## How to test-drive this PR
- Run `npm run dev`
- After visiting 3 pages (this can be changed in app/containers/app/container.jsx - the SoftAsk component accepts a `showOnPageVisit` prop)


benlast:
Even though I haven't reviewed this, I'm very inclined to just give it a plus-1 based on the PR title :)

web/app/components/push-messaging/web-push-connector.jsx
```diff
@@ -0,0 +1,72 @@
+import React, {PropTypes} from 'react'
+import {connect} from 'react-redux'
+import {createPropsSelector} from 'reselect-immutable-helpers'
+
+import * as messagingActions from '../../store/push-messaging/actions'
+import * as appSelectors from '../../containers/app/selectors'
+
+class WebPushConnector extends React.Component {
```
jansepar: I wonder if we want to use a different name for this - "Connector" will have a very special meaning once the Integration Manager branch lands. When it lands, this will be confusing.

jeremywiebe: @jansepar We based it off of the `native-connector` which performs the same job.  Maybe we could call these things "bridges"? They bridge from the PWA/React world to some other world...

https://github.com/mobify/platform-scaffold/blob/develop/web/app/containers/native-connector/container.jsx#L14

web/app/components/push-messaging/web-push-connector.jsx
```diff
@@ -0,0 +1,70 @@
...
+        // We don't render any UI - so don't update
+        return false
+    }
+
+    componentWillReceiveProps() {
+        // TODO - Page visit persistence?
+    }
+
+    render() {
+        return <noscript />
```
jeremywiebe: You can just return `false` or `null` to render nothing: https://facebook.github.io/react/docs/react-component.html#render

web/app/components/push-messaging/soft-ask.jsx
```diff
@@ -0,0 +1,125 @@
+import React, {PropTypes} from 'react'
```
benlast: It occurs to me that @johnboxall would prefer we avoid the term _soft-ask_ (in things visible outside Mobify) and here it's in the names of the files. I don't think we've agreed on a different term, though.

web/app/components/push-messaging/soft-ask.jsx
```diff
@@ -0,0 +1,125 @@
...
+class SoftAsk extends React.Component {
+    constructor(props) {
+        super(props)
+
+        this.state = {
+            isShown: false,
+            isDismissed: false
+        }
+
+        this.onDismiss = this.onDismiss.bind(this)
```
benlast: I'm guessing then that a dismiss is equivalent to a No? The old client sends different events for ALLOWED, DENIED and DISMISSED, and I *think* the old code may support backing-off in different ways depending on whether the shopper denies or dismisses.

web/app/components/push-messaging/soft-ask.jsx
```diff
@@ -0,0 +1,125 @@
...
+            className
+        } = this.props
+
+        const classes = classNames('c-messaging-soft-ask', className)
+
+        return (
+            <Sheet
+                disableScrollLock
+                open={this.state.isShown}
+                effect="slide-bottom"
```
benlast: Not a comment that needs any change, but it's glorious to see how much easier this stuff is to customize than messing around with the old soft-ask!

web/app/components/push-messaging/web-push-connector.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+        this.handleStateUpdate = this.handleStateUpdate.bind(this)
+    }
+
+    componentDidMount() {
+        console.info('[Messaging] Init start')
+        window.Progressive.MessagingClientInitPromise
+            .then(() => window.Progressive.MessagingClient.register(this.handleStateUpdate))
+            .then((state) => this.props.stateUpdate(state))
+            .then(() => console.info('[Messaging] Init finish'))
```
benlast: Does this need a `catch` in case the Messaging init fails?

web/app/loader.js
```diff
@@ -133,10 +133,19 @@ const attemptToInitializeApp = () => {
         // the Messaging worker version data.
         deferredUntilLoadComplete.push(updateMessagingSWVersion)
 
+        // Creating an early promise that users of the Messaging Client can
+        // chain means they don't need to poll for it's existence
+        let clientInitResolver
```
benlast: I suspect that one or more of the other reviewer will want this moved into `loader-utils.js` to keep the actual loader small :)

web/app/components/push-messaging/web-push-connector.jsx
fractaltheory: Hm, only if we want to be able to do something like `console.log` about it if it failed. But if it did fail, `loadAndInitMessagingClient` would catch the error, reject the promise, and we would never get to any of the `then`'s in this chain. 

As long as interaction with Messaging Client in the app is via the `MessagingClientInitPromise` - we don't really need any further guards, I don't think.

fractaltheory:
Closing in lieu of a smaller PR that leverages changes in the SDK.

PR #587, opened 2017-05-11
[Integration Manager] Switch to using the notifications Redux branch from the SDK

DanielBrookRoberge:
This PR removes the notifications storage from the `app/ui` store branch, instead installing the SDK port of the notification storage at the root of the store. 

 **JIRA**: N/A
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/629

## Changes
- Remove the old notification system
- Use the new one

## How to test-drive this PR
- Preview the site after an `npm install`
- Cause a notification to occur (the easiest way I have found is by setting the page to offline and then attempting to add to cart)
- See that the notification appears, and dismisses when the close button is clicked
- Cause another notification to occur
- See that it goes away on navigating to a new page.


jnurse:
Maybe it's unrelated or my npm link isn't working properly, but I'm getting errors when I try to do things like open the menu.

jeremywiebe:
I had the same issue, I think. Tapping the menu causes this error in the console:

image

DanielBrookRoberge:
That's the error that happens when there are two copies of React, one from the SDK and one from the scaffold. You'll need to do `rm node_modules/react node_modules/react-dom` in the SDK directory or install into the scaffold using a `file:` URL in package.json

PR #589, opened 2017-05-12
Fix checkout after merge

jeremywiebe:
Related to: https://github.com/mobify/platform-scaffold/pull/579

I broke checkout when I merged `develop` into `integration-manager`. This PR fixes it so you can flow all the way through checkout again!  *whew*

 **JIRA**: N/A
 **Linked PRs**: N/A

## Changes
- Updated some imports to use integration manager objects now
- Fixed structuring of the payment info we submit to Merlin's

## How to test-drive this PR
- Run `npm run dev`
- Add one, or more items to cart, and checkout


web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -202,35 +202,25 @@ export const submitPayment = (formValues) => (dispatch, getState) => {
...
-        username
-    } = formValues
+
     const address = {
-        firstname,
-        lastname,
-        company: company || '',
-        postcode,
-        city,
-        street: addressLine2 ? [addressLine1, addressLine2] : [addressLine1],
```
DanielBrookRoberge: So the form values no longer contain `addressLine1` and `addressLine2`?

jeremywiebe: That's correct. We "normalize" that before dispatching the IM command [here](https://github.com/mobify/platform-scaffold/blob/integration-manager/web/app/containers/checkout-payment/actions.js#L49,L77)

PR #594, opened 2017-05-15
Basic AMP scaffold combining Oli and Karl's ideas.

olibrook:
- Compiles CSS and JS to a single file using webpack (lambda has file
  size limit)
- Supports inclusion of AMP's JS code on a per-component-rendered
  basis.
- Includes a real usage of AMP's lightbox component as a POC with
  React.


amp/app/components/amp-lightbox/index.jsx
```diff
@@ -0,0 +1,28 @@
+import React, {PropTypes} from 'react'
+import classNames from 'classnames'
+import * as ampSDK from '../../amp-sdk'
+
+/**
+ * An AMP lightbox
+ */
+const AmpLightbox = ({className, children, id}) => {
```
kschmidtdev: I think we may want to capitalize `AMP` even for our code naming, since it is an acronym - thoughts?

amp/app/components/amp-lightbox/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+AmpLightbox.propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    id: PropTypes.string
+}
+
+export default ampSDK.ampComponent(
+    AmpLightbox,
+    '<script async custom-element="amp-lightbox" src="https://cdn.ampproject.org/v0/amp-lightbox-0.1.js"></script>'
```
kschmidtdev: Nice!

amp/app/containers/pdp/index.jsx
```diff
@@ -0,0 +1,22 @@
+import Page from '../../components/page/'
```
kschmidtdev: The PWA currently names the container `jsx` as `container.jsx` instead of `index.jsx` - thoughts on keeping that naming convention for AMP?

amp/app/main.js
```diff
@@ -0,0 +1,105 @@
...
+        .catch(next)
+}
+
+const homePage = (req, res, next) => {
+    initializeStore(req)
+        .then((store) => render(req, res, store, Home))
+        .catch(next)
+}
+
+app.get('/', homePage)
```
kschmidtdev: This can be done in a second pass, but I'd prefer to see the routing configuration in a separate `js` file (to address https://mobify.atlassian.net/browse/AMP-44)

amp/app/components/amp-lightbox/index.jsx
olibrook: The generator we are using which is forked from the `pwa` project purposefully converts `AMP->Amp` - there are even tests for this! I don't mind either way... You can choose.

amp/app/components/amp-lightbox/index.jsx
olibrook: Thanks!

amp/app/containers/pdp/index.jsx
olibrook: Yeaaah, I changed this so that the imports for `containers` and `components` would match -
 one is `import components/thing` and the other is `import containers/thing/container` in the PWA. I am happy to change it back if you prefer consistency with the PWA.

amp/app/components/amp-lightbox/index.jsx
kschmidtdev: Haha oh dear... it's just naming so not really a big deal, maybe I'll try to check our own guides for naming conventions just to make sure we're 'doing it right' 😂

amp/app/main.js
olibrook: Not sure I understand what you mean. I was trying to capture, in this one file, what we would expect a partner has to change from project-to-project. That includes routers/rendering/parsing etc. 

Do you just mean you want to see `routes.js` for aesthetic reasons? That is fine if so!

amp/app/components/amp-lightbox/index.jsx
kschmidtdev: Ok this appears to be the accepted way of dealing with Camel case, so no changes needed.

amp/app/containers/pdp/index.jsx
kschmidtdev: I think this is a good case for consistency, thanks!

amp/app/main.js
kschmidtdev: That's a good idea, keeping everything project-specific in a single file. I think `render` might be able to be factored out maybe? 👍 to `routes.js` mostly for aesthetic reasons yes. Also so when a developer searching for `a file with route in the name` that comes up 😄

amp/app/containers/pdp/index.jsx
olibrook: Fixed here and in the amp-sdk.

amp/app/main.js
olibrook: If it's okay by you can we do this later? Just moving the routes around on their own causes a circular dependency that I would like to avoid. I think a clearer picture of what-belongs where might emerge later on anyway.

I've addressed all the other points, so a thumbs-up would be just grand!

kschmidtdev: Yes definitely do this later!

kschmidtdev:
How do I go about testing this locally? I tried a fresh `npm i` and then `npm run dev` - and when I try to hit `localhost:3000/books.html` I get:
```/Users/karl/dev/progressive-web-scaffold/amp/node_modules/babel-core/lib/transformation/file/index.js:590
      throw err;
      ^

SyntaxError: /Users/karl/dev/progressive-web-scaffold/amp/app/stylesheet.scss: Unexpected token (1:0)
> 1 | @charset 'UTF-8';
```

Same thing with `localhost:3000/eye-of-newt.html`

olibrook:
Ah, I fixed the dev server in the next ticket. Can you do 

`./node_modules/.bin/webpack` to build and then 
`node build/main.js` to run

The next jira ticket (PR open!) bundles that up into a single task.

kschmidtdev:
Ok awesome! I think we should just merge this one now 👍 

One non-blocking thing I noticed was the AMP Validator reported this error: ```The attribute 'is' may not appear in tag 'button'.```

olibrook:
Urgh! I'm merging this, but just so it's written down - that `is` attribute is how you tell React that you are going to use non-standard tags on an element. Very frustrating and surprising that it actually outputs those. I thought it was for internal use only.

PR #595, opened 2017-05-15
WEB-1374: Prevent PWA from loading on Firefox

gsaray:
Prevent PWA from loading on Firefox

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1374

## Changes
- Prevent PWA from loading on Firefox

## How to test-drive this PR
- Run `npm run dev`
- Change user agent to firefox


web/app/loader.js
```diff
@@ -141,7 +141,7 @@ const attemptToInitializeApp = () => {
         deferredUntilLoadComplete.push(() => loadAndInitMessagingClient(IS_PREVIEW, MESSAGING_SITE_ID))
     }
 
-    if (isReactRoute() && !isSamsungBrowser(window.navigator.userAgent)) {
+    if (isReactRoute() && !isSamsungBrowser(window.navigator.userAgent) && !isFirefoxBrowser(window.navigator.userAgent)) {
```
jeremywiebe: What do you think about introducing a small `isSupportedBrowser()` function here that we could add to.  It'll make this `if` check easier to read and gives us expandability for future unsupported browser checks.

jansepar: What we had talked about was that which browsers we support is actuallyvery project specific. A generic function could be dangerous because if we change it, and an older build upgrades, it could suddenly run on a browser it previously had no intention of running on.

jansepar: Unless you mean adding that function in the loader itself, if so, +1 :)

gsaray: first I thought about it but I realized that we are going to test and support samsung browser next quarter that's why I added separate checks so it's easy to remove when it's supported.

PR #600, opened 2017-05-16
Make Messaging configurable. And great again. Yuge.

benlast:
Make Messaging configurable.

To support integration of Messaging configuration into the `platform-generator`, add a single `messagingEnabled` value in `package.json` (for both `web` and `native`) that can be set by
the generator code.

 **JIRA**: https://mobify.atlassian.net/browse/MSG-482
 **Linked PRs**: https://github.com/mobify/platform-generator/pull/35

## Changes
- Extend the `webpack` configs `web/webpack/base.loader.js`, `web/webpack/base.main.js` and `native/webpack.app.config.js` to use a `DefinePlugin` to set `MESSAGING_ENABLED` from the `messagingEnabled` value of the appropriate `package.json`. In the case of the app config, also use `MESSAGING_SITE_ID` in the same way that it's done in the web config.
- In `web/app/loader.js`, if `MESSAGING_ENABLED` is false, don't load the PWA Messaging client. Set `window.Progressive.Messaging.enabled` to match `MESSAGING_ENABLED`.
- In `web/app/native/onboarding/onboarding.jsx`, if MESSAGING_ENABLED is false, don't include the `notification` part of the onboarding carousel.
- Set `AstroNative.Configuration.messagingEnabled` from `MESSAGING_ENABLED` in `native/app/config/baseConfig.js`
- In `native/app/onboarding/onboardingModalController.js`, if `AstroNative.Configuration.messagingEnabled` is false, don't initialize the push controller and make the `push:enable` RPC call a no-op

## How to test-drive this PR
- Set `messagingEnabled` to `true` and `false` in `web/package.json` and `native/package.json`> For each setting:
  - in `web`, run `npm run dev`
  - in `native`, run `npm run deps`
  - In Chrome (web) preview and verify that messaging is or is not loaded as expected by checking the console output
  - In XCode, build and run the app and check that the Notifications pane in the carousel is 
included or not as expected
  - In Android Studio, build and run the app and check that the Notifications pane in the carousel is included or not as expected

Note that there are **two** places where `messagingEnabled` is defined, one in each `package.json`. Because of the way code is shared between the `native` and `web` parts of the scaffold, the setting in `web/package.json` affects the PWA **and** the Onboarding part of the app, and the setting in `native/package.json` affects whether the app scaffold initializes messaging.


native/package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "messagingSiteId": "merlinspotions",
```
jeremywiebe: I wonder if we should have a `mobify` key in the `package.json`  I wonder if things will get out of hand with custom-defined keys in our `package.json` files that aren't obviously custom-Mobify settings. Reference how Jest [does it](https://facebook.github.io/jest/docs/configuration.html).

web/app/native/onboarding/onboarding.jsx
```diff
@@ -69,10 +71,14 @@ const login = {
 
 const carouselData = {
     location,
-    notifications,
     login
 }
 
+// MESSAGING_ENABLED is replaced at build time by webpack.
```
jeremywiebe: I was scrolling through and resisting commenting...  but I will.  :)

What if we had a convention that webpack-provided global constants were all prefixed with `WEBPACK_`?  That would reduce the need for these types of comments and it would become easier to see, at a glance, that the constant came from webpack.

web/webpack/base.main.js
```diff
@@ -12,6 +12,8 @@ const ExtractTextPlugin = require('extract-text-webpack-plugin')
 const CopyPlugin = require('copy-webpack-plugin')
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
 
+const webPackageJson = require('../package.json')   // eslint-disable-line import/no-extraneous-dependencies
```
jeremywiebe: So if you are ok with the idea of a `mobify` key in the `package.json`, we could move all these calls that read the web's `package.json` into `base.common` and export the entire contents of that key as a part of what `base.common.js` exports.  I _think_ that is cleaner... not sure about clearer. :)

native/package.json
benlast: I don't mind either way, but I'm wary of my PR expanding to try and cover that.

benlast:
Chatted to @jeremywiebe about his comments - I agree they're good suggestions, but I don't really want to get into refactoring the whole package-variable-definition setup as part of this PR.

native/app/onboarding/onboardingModalController.js
```diff
@@ -24,14 +24,17 @@ const OnboardingModalController = function(modalView, onboardingController) {
 }
 
 OnboardingModalController.init = async function() {
+    const messagingEnabled = AstroNative.Configuration.messagingEnabled
```
kschmidtdev: I think `messagingEnabled` lives in `baseConfig` and not within `AstroNative.Configuration.`

native/app/onboarding/onboardingModalController.js
```diff
@@ -51,10 +54,13 @@ OnboardingModalController.init = async function() {
     })
 
     Astro.registerRpcMethod(AppRpc.names.pushEnable, [], () => {
-        if (AstroNative.Configuration.DEBUG) {
-            pushController.subscribeTest()
-        } else {
-            pushController.subscribe()
+        // If pushController wasn't initialized, this becomes a no-op
```
kschmidtdev: I'm torn about whether this is the right spot to branch on push enabled-ness. Doing it here means a new project will have built-in onboarding with an 'enable push' button that will just not work at all, if you don't have messaging enabled. Do we do anything else in onboarding? Skipping that step in onboarding isn't a great solution either, since there is push configuration/init that would happen when it shouldn't. Maybe both?

benlast: @kschmidtdev the notification pane of the onboarding carousel is removed when messaging is disabled (though that's controlled by the flag in `web/package.json`)

native/app/onboarding/onboardingModalController.js
benlast: @kschmidtdev I'm a bit confused then... and can't see how I access the config. But I can also do this using the webpack-defined `MESSAGING_ENABLED`, which I guess avoids adding the flag.

web/app/loader.js
```diff
@@ -31,18 +31,22 @@ const attemptToInitializeApp = () => {
...
+    const messagingEnabled = MESSAGING_ENABLED  // replaced at build time
 
     const CAPTURING_CDN = '//cdn.mobify.com/capturejs/capture-latest.min.js'
     const ASTRO_CLIENT_CDN = `//assets.mobify.com/astro/astro-client-${ASTRO_VERSION}.min.js`
     const SW_LOADER_PATH = `/service-worker-loader.js?preview=${IS_PREVIEW}&b=${cacheHashManifest.buildDate}`
 
+    window.Progressive = {
+        AstroPromise: Promise.resolve({}),
+        Messaging: {
+            enabled: messagingEnabled
```
kschmidtdev: Does the messaging client use this? I can't find any references to it in this PR/repo

native/app/onboarding/onboardingModalController.js
kschmidtdev: @benlast OH, I see where the notifications pane is conditional set now.

web/app/loader.js
benlast: It's available for use by code that isn't part of this repo

web/app/native/onboarding/onboarding.jsx
```diff
@@ -69,10 +71,14 @@ const login = {
...
+if (MESSAGING_ENABLED) {
+    carouselData.notifications = notifications
```
kschmidtdev: I'm not sure if it's just my machine, but I'm noticing that the order of pages in onboarding is different with these changes (if notifications are enabled) - it ends up being the last entry, when login should be the last. Unfortunately the original version isn't strictly ordered so some refactoring is needed to fix..

kschmidtdev: Reproduced in iOS and Android

web/messaging/progressive-web-scaffold.yaml
```diff
@@ -0,0 +1,41 @@
...
+    friendly_site_name: "Mobify Progressive Web Scaffold"
+
+    # Must be a 192x192 icon file.
+    default_icon_url: "https://cdn.mobify.com/sites/progressive-web-scaffold/production/static/img/global/app-icon.png"
+
+    # Default is to use the URL for the target domain, via https
+    default_url:
+
+    # The domain (netloc) of the project's website, which must be
+    # https-everywhere.
```
fractaltheory: Similar to what we recently added to the old generator, how about a comment noting that the value for this setting _shouldn't_ include a protocol?

web/messaging/progressive-web-scaffold.yaml
```diff
@@ -0,0 +1,41 @@
...
+    cloud_slug: "progressive-web-scaffold"
+
+    # Human-readable site name
+    friendly_site_name: "Mobify Progressive Web Scaffold"
+
+    # Must be a 192x192 icon file.
+    default_icon_url: "https://cdn.mobify.com/sites/progressive-web-scaffold/production/static/img/global/app-icon.png"
+
+    # Default is to use the URL for the target domain, via https
+    default_url:
```
fractaltheory: Is this meant to be blank here?

benlast: Yes :)

web/messaging/progressive-web-scaffold.yaml
benlast: Good idea...

web/app/native/onboarding/onboarding.jsx
benlast: I'm not sure how to fix this... is it a blocker? @kschmidtdev ?

jeremywiebe: It's because the Onboarding screen does [`Object.keys(carouselData)`](https://github.com/mobify/platform-scaffold/blob/59ed3326bf527731b7642cc17c4d0d9e3b5b6e38/web/app/containers/onboarding/container.jsx#L56) so to me that means the order is undefined.

@benlast Could you log a bug in the Astro JIRA to fix this and then lets move on.

web/app/native/onboarding/onboarding.jsx
benlast: 💯 points of agreement, but not in this PR :)

web/webpack/base.main.js
benlast: That would (now) require a matching change in the generator, which processes just the top-level keys to do replacements (in https://github.com/mobify/platform-generator/pull/35 )

web/app/native/onboarding/onboarding.jsx
benlast: Will do... but I'm curious - did that work before by luck/coincidence, since JS object key order is (unless I missed something) always undefined?

jeremywiebe: I suspect it was by luck and probably because we never changed the order that we added things.

kschmidtdev: Fix in #634

PR #601, opened 2017-05-16
[Polish] Reorder theme imports

jeffkamo:
Previously, some custom SDK component styles would be overridden (beaten) by base styles due to the CSS cascade. For example...

```scss
@import 'node_modules/progressive-web-sdk/dist/components/button/base';
@import 'pw-components/button';
@import 'node_modules/progressive-web-sdk/dist/components/icon/base';
```

In the above example, if our custom `button` style had something to do with specific `icon` variation, the icon base could override our custom button's icon style.

Thus, to get around this, all custom styles are imported AFTER all the base styles to ensure developers have complete control over the SDK component styles.

## Changes
- Re-order SDK component imports so local customization imports come after all base style imports

## How to test-drive this PR
- Run `npm run dev`
- **Verify** that component styles still look correct throughout the app (nothing was lost or broken)

### DIFF
--- a/web/app/styles/themes/_pw-components.scss
+++ b/web/app/styles/themes/_pw-components.scss
@@ -52,7 +52,6 @@
 // @import 'node_modules/progressive-web-sdk/dist/components/rating/base'; // has base styles, but isn't used
 @import 'node_modules/progressive-web-sdk/dist/components/ratio/base';
 @import 'node_modules/progressive-web-sdk/dist/components/search/base';
-// @import 'node_modules/progressive-web-sdk/dist/components/select/base'; // Select is deprecated @TODO: deprecate this component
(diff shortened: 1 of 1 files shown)


web/app/styles/themes/_pw-components.scss
```diff
@@ -7,289 +7,95 @@
...
-// ---
-
-// @import 'node_modules/progressive-web-sdk/dist/components/select/base';
-// @import 'pw-components/select';
-
-
-// Sheet
-// ---
-
+// @import 'node_modules/progressive-web-sdk/dist/components/select/base'; // Select is deprecated @TODO: deprecate this component
```
gsaray: As we discussed, let's remove select component import since it's deprecated.

PR #603, opened 2017-05-17
Surface TTI and performance values from Lighthouse report

ellenmobify:
Print out Time To Interactive and Total Byte Weight (total payload size).

## Changes
- Get the Lighthouse report in JSON and HTML formats
- Surface TTI and bundle sizes

## How to test-drive this PR
- Rebuild on CI: https://circleci.com/gh/mobify/platform-scaffold/tree/lighthouse-performance

## Discussion
screen shot 2017-05-15 at 3 03 29 pm
With regards to bundle sizes, the JSON report gives us all the data shown in the screenshot. The “Total Size” corresponds to “Parsed Size” in [webpack bundle analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) report, defined as the **minified** size.. In this or a future PR, we can add a check for bundle sizes above a certain threshold. What should that threshold be?


web/scripts/lighthouse-ci.sh
```diff
@@ -2,7 +2,7 @@
 # Run the Lighthouse test against the dev build with continuous integration.
 
 # Location to save the generated HTML report.
```
jeremywiebe: This becomes the output filename prefix, right?  Could you update the comment to reflect that?

web/lighthouse/check-score.js
```diff
@@ -3,17 +3,26 @@
...
  */
 
-const report = fs.readFileSync('./lighthouse/audit-local.html', 'utf8')
+const htmlReport = fs.readFileSync('lighthouse/audit-local.report.html', 'utf8')
 
+const jsonResults = JSON.parse(fs.readFileSync('lighthouse/audit-local.report.json', 'utf8'))
+
+// Tell me about some performance metrics that are important to me.
+console.log(`Time to interactive: ${jsonResults.audits['time-to-interactive'].displayValue}`)
+console.log(`${jsonResults.audits['total-byte-weight'].displayValue}`)
```
johnboxall: @ellenmobify just one note here: 

When we start looking at the threshold for weight, we only want to be looking at the files we control ...

image

So of these files, we're interested in the size of everything coming off cdn.mobify.com – we might actually need to look at Lighthouse to see if we can get the full URL stored here ... because right now it appears only the partial URL shows up in the JSON.

johnboxall:
One more thought on this from our discussion the other day:

Our JavaScript pipeline looks something like this:

Raw Files => Built File => Minified File => Compressed File

The thing we want to test is the size of the minified file – if it is above 2 MB it has serious consequences for our distribution (our CDN will not compress it)

The other way to approach file weight is like https://github.com/mobify/platform-scaffold/pull/577 – this may be a better check than looking at these in Lighthouse...

PR #604, opened 2017-05-17
Remove !important to utilities

cole-sanderson:
Remove `!important` styles to utilities. 

 **JIRA**: https://mobify.atlassian.net/browse/AMP-38
 **Linked PRs**: N/A

## Changes
- change `fadeIn` to `fade-in`
- change `inlineLoaderPulseAnimation` to `inline-loader-pulse`
- move animation styles in its own files
- new mixin: `wrap-at-root`
- apply `wrap-at-root` to utilities
- remove `!important` styles to utilities 

## How to test-drive this PR
- Run `npm run dev`
- Check all pages to make sure that there are no broken styles


web/app/styles/lib/_style-override.scss
```diff
@@ -0,0 +1,40 @@
...
+// must be dealt with separately.
+
+
+// Configurable variables
+// ---
+
+// The default selector used to build the override. Since it’s a special
+// character in CSS, the `!` must be escaped. Eg. <div id="!">.
+
+$style-override-selector: '#\\!' !default;
```
jeffkamo: Since this mixin exists in the scaffold now, is there a reason to have a default selector that is different from what we're using elsewhere? If `#!` isn't present in the markup anywhere, then this isn't as useful anymore.

web/app/styles/utilities/_heading.scss
```diff
@@ -3,36 +3,36 @@
 
 %h1,
```
jeffkamo: This is something to perhaps be wary of. Do the HTML heading base styles (h1, h2, etc.) inherit the ID selectors since they're extending the utility classes?

web/app/styles/utilities/_text.scss
```diff
@@ -6,163 +6,163 @@
...
 }
 
 .u-text-family-header {
-    font-family: $header-font-family !important;
+    font-family: $header-font-family;
 
     // Added with webfontloader. See loadFonts() in app/container.js
     .wf-active & {
-        font-family: $loaded-header-font-family !important; // TODO: Remove !important
+        font-family: $loaded-header-font-family; // TODO: CHECK HERE
```
jeffkamo: What's this Todo for now?

web/app/styles/utilities/_layout.scss
```diff
@@ -115,7 +115,7 @@
 // will contain floated children and their margins.
 
 .u-block-context {
-    overflow: hidden !important;
+    overflow: hidden;
 }
```
jeffkamo: Below this line is `.u-center-piece` which didn't have `!important` before. It outputs the following:

```css
.u-center-piece {
  max-width: 1200px;
  margin-left: auto;
  margin-right: auto; }
  .u-center-piece:after {
    content: " ";
    display: block;
    clear: both; }
```

Should this have higher specificity like the rest? Or stay at low specificity?

web/app/styles/themes/pw-components/_search.scss
```diff
@@ -80,7 +80,7 @@
 // ---
 
 .pw-search__suggestion-heading {
-    @include visually-hidden(true);
+    @include visually-hidden;
```
jeffkamo: Is this style going to be shared with the AMP app? If not, is its lower specificity still adequate to work in this situation?

web/app/styles/_utilities.scss
```diff
@@ -1,18 +1,20 @@
...
-@import 'utilities/flexbox';
-@import 'utilities/heading';
-@import 'utilities/layout';
-@import 'utilities/margin';
-@import 'utilities/padding';
-@import 'utilities/text-content';
-@import 'utilities/text';
-@import 'utilities/visibility';
-@import 'utilities/z-index';
+@include style-override(3, $selector: "#app") {
```
jeffkamo: Also, I don't think you need to specify the second argument's variable. You can just pass in the property:

```scss
@include style-override(3, "#app") { ... }
```

jeffkamo: Does this override need to start at a degree of 3? Is that the lowest we can go?

web/app/styles/utilities/_heading.scss
jeffkamo: Yup, the tags inherit the IDs too. This will need to be tweaked.

screen shot 2017-05-17 at 3 26 41 pm

web/app/styles/lib/_style-override.scss
```diff
@@ -0,0 +1,40 @@
...
+// @param $selector [String]: An id selector that matches the
+// $style-override-selector tag.
+
+@mixin style-override($degree: 1, $selector: $style-override-selector) {
+    $selector-chain: '';
+
+    // Build an id selector by chaining the same id onto itself once more than
+    // the specified degree. So if degree: 4, we get #id#id#id#id#id.
+
+    @for $i from 1 through $degree + 1 {
```
jeffkamo: This actually makes the `$degree` count output one more ID than indicated when using the mixin. I wonder if this should still be the logic... I'm beginning to think this should behave explicitly to what the user declares. So specifying a degree of `3` should output `#foo#foo#foo` not `#foo#foo#foo#foo`.

web/app/styles/_utilities.scss
cole-sanderson: I think 3 is reasonable?

jeffkamo: I feel hesitant to wave high specificity, that also has excess bloat, without very good reason. Is there an explicit need for three IDs in order for our utility classes to function as intended in the current state of our scaffold?

web/app/styles/utilities/_text.scss
cole-sanderson: confirmed that there is no issue, removed comments.

web/app/styles/utilities/_layout.scss
cole-sanderson: since its utility class and lets make it higher specificity.

web/app/styles/themes/pw-components/_search.scss
cole-sanderson: No, it wont be shared with AMP. There is no !important option in mixin.

web/app/styles/lib/_style-override.scss
cole-sanderson: Fixed the logic.

web/app/styles/utilities/_heading.scss
cole-sanderson: good catch! Will move heading tags outside of heading utility class.

cole-sanderson: fixed the issue

web/app/styles/_utilities.scss
cole-sanderson: I think 1 is too low and if you think 3 is too many, how about to make it 2?

web/app/styles/lib/_wrap-at-root.scss
```diff
@@ -0,0 +1,30 @@
...
+
+// Parameters
+// ---
+//
+// @param $degree [Number]: The “force” of the override. Effectively the number
+// of id-level selectors you need to override.
+// @param $selector [String]: An id selector that matches the
+// $style-override-selector tag.
+
+@mixin wrap-at-root($degree: 1, $selector: '') {
```
jeffkamo: We might want to tweak `$selector` to not have a default value at all (blank `''` is arguably still a default value of sorts).

This might require reordering the arguments to `@mixin wrap-at-root($selector, $degree: 1)`

web/app/styles/lib/_wrap-at-root.scss
```diff
@@ -0,0 +1,30 @@
...
+// styles; those must be dealt with separately.
+
+
+// Parameters
+// ---
+//
+// @param $degree [Number]: The “force” of the override. Effectively the number
+// of id-level selectors you need to override.
+// @param $selector [String]: An id selector that matches the
+// $style-override-selector tag.
```
jeffkamo: Is `$style-override-selector` a thing? This could perhaps be rephrased for clarity

web/app/styles/lib/_wrap-at-root.scss
```diff
@@ -0,0 +1,30 @@
+// Wrap At Root
+// ===
+//
+// Allows overriding exisiting styles with extreme specificty while keeping
+// source code clean and readable. Does not override !important or inline
+// styles; those must be dealt with separately.
```
jeffkamo: With the renaming of this mixin, it perhaps should not be described as a method of overriding existing styles anymore.

web/app/styles/lib/_wrap-at-root.scss
```diff
@@ -0,0 +1,30 @@
...
+
+
+// Parameters
+// ---
+//
+// @param $degree [Number]: The “force” of the override. Effectively the number
+// of id-level selectors you need to override.
```
jeffkamo: This too, can this be described without mentioning `override`?

web/app/styles/lib/_wrap-at-root.scss
```diff
@@ -0,0 +1,30 @@
...
+// @param $selector [String]: An id selector that matches the
+// $style-override-selector tag.
+
+@mixin wrap-at-root($degree: 1, $selector: '') {
+    $selector-chain: '';
+
+    // Build an id selector by chaining the same id onto itself once more than
+    // the specified degree. So if degree: 4, we get #id#id#id#id#id.
```
jeffkamo: This comment needs to be updated so that a degree of 4 gets you `#id#id#id#id`

cole-sanderson:
@jeffkamo just fixed `wrap-at-root` comments.

web/app/styles/_utilities.scss
```diff
@@ -5,7 +5,7 @@
 // specificity while keeping source code clean and readable. Does not override
 // !important or inline styles; those must be dealt with separately.
 
-@include wrap-at-root(1, "#app") {
+@include wrap-at-root("#app") {
```
jeffkamo: Minor: Maybe still include the the `1` so it's easy for people to increase if needed.

PR #605, opened 2017-05-17
Updated README and added new cert for Windows users

rsexton404:
## Progressive Web Scaffold - Hotfix release!
- [x] Branch off of `master` and create a `hotfix-documentation-windows-install` branch.
- [x] Create a new pull request with the following settings:
      * Base: `master`
      * Compare: `hotfix-documentation-windows-install`
      Paste the contents of this checklist into this pull request.
- [x] Update the version of the `progressive-web-sdk` npm module in `package.json`.
- [x] Make necessary changes to make it compatible with the latest release of the `progressive-web-sdk`.
- [x] In the CHANGELOG.md file of this release branch, change the 'To be released' header to 'vX.Y.Z'.
- [ ] Merge `release-vX.Y.Z` into `master` using the pull request and delete the release branch.
- [ ] Draft a new Github release with the following settings:
      * Tag version: `X.Y.Z` @ `master`
      * Release title: `X.Y.Z - <release_name>`
      * Description: Use highlights from the CHANGELOG.md (only pick out the most significant changes)
- [ ] Merge `master` into `develop` (no need for review on PR, just merge).

## Bundle Release
- [ ] Run `npm run push -- -m '<X.Y.Z>:<COMMIT_SHA>,<BRANCH_RELEASED_FROM>'`. You can find your `<COMMIT_SHA>` from `git rev-parse --short HEAD`

# Changes

On Windows 10 (not sure about earlier versions), it is not possible to add a `.pem` file to the Windows Trusted Root Certificate Store to suppress SSL errors during development. They will need a `.p7b` file instead. To support our Windows users on upcoming training events, we wanted to create the `.p7b` file for them and include it in the scaffold. 

**Security note:** I converted our `.pem` file `.p7b` format using a third-party site called [SSLShopper](https://www.sslshopper.com/). If this is not kosher from a security standpoint (do we fully trust the converted file?), please say so!

I've also added clarification about having to run npm scripts from the web directory inside the project directory since our users frequently get tripped up by this during training.

**Question**: Should we be describing these as "HTTPS errors" (as we do in the section title) or "security warnings" (as we do in the body text)? If we change the section title to match, we may have to update deep links to this section in a few places.

## Changes
- Added `.p7b` file
- Added new section at the top of README: Running npm scripts
- Edited instructions for HTTPS errors for macOS and Windows SSL for accuracy and clarity

## How to test-drive this PR
- Open README.md
- Follow instructions for "Avoiding HTTPS errors in local development" 
- Follow both macOS and Windows instructions on fresh (virtual) machines, if possible
- Verify that both sets of steps are numbered properly
- Verify that step #4 in the macOS instructions makes sense (I don't have a fresh config to test)
- Make sure to add the `.p7b` file when following the Windows instructions
- Verify that all you have to is close the browser and reopen it for the new certificate to take effect. Do you also have to clear Mobify cookies? Is this behaviour the same for Firefox (and others) as it is with Chrome?
- Run `npm run dev`
- There should be no security warnings or errors, as we promise


web/README.md
```diff
@@ -126,9 +139,9 @@ npm run lint
 If this code style is a poor match for your pre-existing practices,
 there are two ways you can modify the linter configuration to suit
 your use case better. For small changes, you can add rules to the
-`.eslintrc.yml` file in the root web directory. Rules specified in
-this file override rules in the Mobify standard; the following `rules`
-section adds an additional rule and disables an existing rule:
+`.eslintrc.yml` file in the `web` directory inside your project directory.
```
jeremywiebe: Given that this `README.md` is already inside the `/web` directory of the developer's project, I feel like both of these statements are redundant. I think what we should be doing is using links to point the user definitively to the right file.

"...an existing rule: [`.eslintrc.yml`](./.eslintrc.yml) file."  

*Note that the link won't work in the PR on Github, but it will work on your local system and on Github once it's merged*

Reference how we do links in a README [here](https://raw.githubusercontent.com/mobify/platform-scaffold/develop/README.md)

web/README.md
```diff
@@ -22,22 +33,24 @@ npm run push -- -m "Test push by <name>"
 
 ## 🔒 Avoiding HTTPS errors in local development
 
-The development server uses a self-signed SSL certificate which is valid, but
-must be added to your operating system to work correctly.
+The development server uses a self-signed SSL certificate. This is totally valid, but
+the self-signed certificate must be added to your operating system to work correctly.
```
jeremywiebe: " to work correctly" glosses over alot of details. :)   We have to add them to the system keychain so that the OS **trusts** the certificate. The system trusts a certificate issued by a certificate vendor because OS and Browser vendors (Microsoft, Apple, Google) ship with built-in "Trusted Root Certificates". The result is that any certificates signed by those trusted root certificates, or certificates issue by them (chain of trust), are trusted by the browser and operating system. 

We step around the entire chain of trust with a self-signed certificate by simply adding it as a trusted root certificate. 

This is probably way too much info, but I think we could expand on the "why" we need to do this a bit more than "to work correctly".  Maybe not in this PR as I know that wasn't what your changes are addressing. :)

`</rant>`

rsexton404: For now, replace "to work correctly" with "to be trusted?"

web/README.md
rsexton404: Keep in mind that sometimes we have to point people to the README on Github just to get everyone on the same page in training, so I'd like to leave the redundancy in there just in case.

Can we use a relative link for the `.eslintrc.yml` file?

jeremywiebe: Yep, you can use the link exactly as I posted in my comment below. 
```md
 [`.eslintrc.yml`](./.eslintrc.yml)
```

jvoll: One thing to keep in mind here is that (I'd estimate) 90% of the people reading the README are actually on github and not on their local machine. I agree it's totally redundant if you're on your machine, but most people just click into this README from the root README and it doesn't really feel like you've "changed directories" so to speak.

web/README.md
```diff
@@ -1,5 +1,16 @@
 # Mobify Progressive Web Scaffold
 
+## Running npm scripts
+
+All the `npm` scripts listed below must be run from the `web` directory inside your project directory. 
+
+Here's how to get a list of all the available npm scripts in the Progressive Web Scaffold:
+
+```
+cd ~/MYPROJECT/web
```
rsexton404: Removed cd command

web/README.md
```diff
@@ -1,5 +1,16 @@
...
+npm run
```
rsexton404: No idea what TIL mean, sorry!

web/README.md
rsexton404: Thanks for providing the exact syntax for the relative link. It wasn't obvious (to me) from the example.

rsexton404:
@jeremywiebe With Jason's help, I updated the `web/scripts/create-localhost-certificate.sh` and it seems to correctly generate the `.p7b` file in the same way as the stupid website I was using did. **HOWEVER**, I did not include this extra bit at the end of the `openssl` command you provided: `-certfile CAcert.cer`

Isn't the `certfile` option already passed in earlier in the command? Is this just an error on the [OpenSSL Commands page](https://www.sslsupportdesk.com/openssl-commands/)? Or am I missing something?

ellenmobify:
👍  We can fix the failing Lighthouse test in a future PR.

jeremywiebe:
Failing lighthouse test recorded as bug: https://mobify.atlassian.net/browse/WEBDATA-83   

Moving forward with release.

PR #608, opened 2017-05-18
Extract steps of submitPayment() so it's easier to follow

jeremywiebe:
This PR just refactors the `submitPayment` command in the SFCC connector so the steps are a bit easier to follow. 

 **JIRA**: (link to JIRA ticket)
 **Linked PRs**: (links to corresponding PRs, optional)

## Changes
- Extract the steps in `submitPayment` into individual functions so it's easier to see the overall flow

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions SFCC](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Run all the way through checkout

PR #612, opened 2017-05-18
JSON request utility for the SFCC connector.

DanielBrookRoberge:
This PR adds a utility function for making JSON-body requests to SFCC. It also refactors the `submitShipping` command in a similar way to #608 for clarity.

 **JIRA**: N/A

## Changes
- Add a utility function to the SFCC connector that automates some of the error checking for JSON requests.
- Use it in the most obviously applicable places
- Split up the `submitShipping` command for clarity

## How to test-drive this PR
- Preview the SFCC site
- Go through the checkout
- See that everything works.


web/app/integration-manager/_sfcc-connector/cart/commands.js
```diff
@@ -96,22 +96,18 @@ export const addToWishlist = (productId) => (dispatch) => {
...
-                .then((response) => response.json())
-                .then((responseJSON) => {
-                    if (responseJSON.fault) {
-                        throw new Error('Unable to add item to wishlist.')
-                    }
-                })
+            return makeSfccJsonRequest(
+                `${API_END_POINT_URL}/customers/${customerID}/product_lists/${id}/items`,
+                requestBody,
+                {method: 'POST'}
```
jeremywiebe: Would it make sense to assume `POST` is the action and let callers override or is that too much magic? Or maybe we have a small suite of convenience methods like `makeSfccJsonPostRequest`, `makeSfccJsonPutRequest`, `makeSfccJsonPatchRequest` ... just for the most common ones. 

Going a bit further... I wonder if in another PR we could just remove `Sfcc` from the method name(s)... it's already inside the connector so it's obviously for SFCC... (i think :) )

web/app/integration-manager/_sfcc-connector/utils.js
```diff
@@ -138,6 +138,20 @@ export const makeSfccRequest = (url, options) => {
         })
 }
 
+export const makeSfccJsonRequest = (url, body, options) => {
+    return makeSfccRequest(url, {
+        ...options,
+        body: JSON.stringify(body)
+    })
+        .then((response) => response.json())
```
jeremywiebe: This could fail (which might be fine) in cases where we get back non-JSON data.  I wonder if we should look at the response `Content-Type` or `status` before we try to unwrap it as JSON?

web/app/integration-manager/_sfcc-connector/cart/commands.js
DanielBrookRoberge: Yeah, what these functions are really for is SFCC API requests, so maybe it could be `makeApiRequest` and `makeApiJsonRequest`. I'm not against the idea of having convenience methods for the different HTTP methods, but I'd rather not default to `POST`

web/app/integration-manager/_sfcc-connector/utils.js
DanielBrookRoberge: It'll definitely throw if we don't have JSON data, though I think that's probably what we want. I'm contemplating status code checks for the follow-up.

PR #613, opened 2017-05-18
AMP build script - creating zip packages runnable on Lambda

olibrook:
Adds script to build an AMP bundle and ensure it can be run on AWS Lambda.

    - Change main.js so that it is runnable locally and exposes a proper
      handler if it looks like it is running under Lambda.
    - Update build-script to use short git commit hashes in bundle names
      (matches PWA).
    - Update build-script to zip files correctly. Tested to ensure
      server.zip can be uploaded and run unmodified on Lambda.

 **JIRA**: https://mobify.atlassian.net/browse/AMP-41

## How to test-drive this PR
- Checkout, and run `npm run build`
- If you like, upload `/build/[hash]/server.zip` to Lambda, and click 'save and test' to convince yourself it works. You can do that here https://console.aws.amazon.com/lambda/home?region=us-east-1#/functions/index?tab=code


amp/webpack.config.js
```diff
@@ -8,7 +8,9 @@ module.exports = {
     target: 'node',
     output: {
         path: path.join(__dirname, 'build'),
-        filename: 'main.js'
+        filename: 'main.js',
+        library: 'main',
+        libraryTarget: 'commonjs2'
```
olibrook: This ensures that the build file exposes the `handler` variable which Lambda requires.

kschmidtdev:
What version of node are you using? `nodegit` is failing to install for me (error), I'm using node `v6.9.1`

olibrook:
I'm on 6.9.4. Might be an idea for us to use 6.10 though, since that's what's running on AWS.

kschmidtdev:
@olibrook `nodegit` fails on 6.10.x too, I had to `brew install libgcrypt`. It seems to be a [known issue](https://github.com/nodegit/nodegit/issues/1166) - so I think we should include that in the README or switch to controlling git via command-line

olibrook:
Ah, good point. I've swapped to `node-rev-sync` instead. Should fix it.

PR #617, opened 2017-05-19
Add a utility to encapsulate $.ajax for Merlin's

DanielBrookRoberge:
There are a few requests on Merlin's that don't work with `fetch` for some reason. Since raw `XMLHttpRequest` is a pain to work with we use `$.ajax` instead, but it still doesn't fit well into our Promise-based workflow. This PR adds a utility function that wraps $.ajax with a Promise-based interface and centralizes all usage of that function.

## Changes
- Encapsulate $.ajax in a Promise wrapper.

## How to test-drive this PR
- Preview Merlin's
- See that the checkout still works with no 400 errors

### DIFF
--- a/web/app/integration-manager/_merlins-connector/checkout/commands.js
+++ b/web/app/integration-manager/_merlins-connector/checkout/commands.js
@@ -238,6 +238,8 @@ const buildFormData = (formCredentials) => {
         }
     })
 
+    formData.append('form_key', getCookieValue('form_key'))
+
     return formData
 }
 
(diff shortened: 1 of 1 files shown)


web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -270,79 +270,82 @@ const createAddressRequestObject = (formValues) => {
...
+        window.Progressive.$.ajax({
+            ...options,
+            success: (responseData) => resolve(responseData),
+            error: (xhr, status) => reject(status)
         })
+    })
+}
+
+const updateBillingAddress = () => (dispatch, getState) => {
+    const formData = buildFormData({
```
jeremywiebe: What do you think about moving the setting of the `form_key` into `buildFormData`?  We should be submitting it with most (all?) form posts anyways...

DanielBrookRoberge: Yeah that makes sense. We set the form key in `submitForm` for any form that works with `fetch`, but `buildFormData` is only used with `$.ajax`.

PR #618, opened 2017-05-19
Valid AMP

kschmidtdev:
Render valid AMP, use `AmpImage` component

## Changes
- Refactored Home container to be closer to how containers are built/structured in the PWA
- Use `AmpImage` instead of `img`
- Fix button AMP validation issue
- Add AMP validation to test suite/CI (CI has to wait until we get the SDK on npm)
- Add `susy` as dependency

## How to test-drive this PR
- `npm i`
- `cd node_modules/mobify-amp-sdk`
- `npm i && npm run dev:build`
- `cd ../..`
- `npm run dev`
- Go to http://localhost:3000


amp/app/containers/home/container.jsx
```diff
@@ -1,22 +1,43 @@
...
+import AmpLightbox from '../../components/amp-lightbox'
 
 const containerClass = 't-home'
 
+const Home = ({links, title}) => {
+    return (
+        <div className={containerClass}>
+            <div dangerouslySetInnerHTML={{__html: '<button on="tap:my-lightbox">Open lightbox</button>'}} />
+            <AmpLightbox id="my-lightbox">
+                <amp-img src="https://media.giphy.com/media/26ufdipQqU2lhNA4g/giphy.gif" width="500" height="500" />
```
olibrook: Maybe fix this one too? :D

amp/app/containers/pdp/container.jsx
```diff
@@ -1,22 +1,43 @@
...
+import AmpLightbox from '../../components/amp-lightbox'
+
+const containerClass = 't-home'
+
+const PDP = ({links, title}) => {
+    return (
+        <div className={containerClass}>
+            <div dangerouslySetInnerHTML={{__html: '<button on="tap:my-lightbox">Open lightbox</button>'}} />
+            <AmpLightbox id="my-lightbox">
+                <amp-img src="https://media.giphy.com/media/26ufdipQqU2lhNA4g/giphy.gif" width="500" height="500" />
```
olibrook: Fix this one.

amp/app/containers/pdp/container.jsx
```diff
@@ -1,22 +1,43 @@
...
+    /**
+     * A title
+     */
+    title: PropTypes.string
+}
 
 const mapStateToProps = (state) => ({
     links: state.links,
-    title: `PDP! - ${state.title}` || '',
+    title: `Home! - ${state.title}` || '',
```
olibrook: PDP!

amp/app/containers/plp/container.jsx
```diff
@@ -1,22 +1,43 @@
...
+import AmpLightbox from '../../components/amp-lightbox'
+
+const containerClass = 't-home'
+
+const PLP = ({links, title}) => {
+    return (
+        <div className={containerClass}>
+            <div dangerouslySetInnerHTML={{__html: '<button on="tap:my-lightbox">Open lightbox</button>'}} />
+            <AmpLightbox id="my-lightbox">
+                <amp-img src="https://media.giphy.com/media/26ufdipQqU2lhNA4g/giphy.gif" width="500" height="500" />
```
olibrook: Fix this one

amp/app/containers/plp/container.jsx
```diff
@@ -1,22 +1,43 @@
...
+    /**
+     * A title
+     */
+    title: PropTypes.string
+}
 
 const mapStateToProps = (state) => ({
     links: state.links,
-    title: `PLP! - ${state.title}` || '',
+    title: `Home! - ${state.title}` || '',
```
olibrook: PLP!

amp/package.json
```diff
@@ -57,6 +60,7 @@
     "lint:js": "eslint '**/*.{js,jsx}'",
     "lint:sass": "sass-lint -c .sass-lint.yml 'app/**/*.scss' -v",
     "test": "cross-env NODE_ENV=test jest",
+    "test:unit": "tests/unit-tests.sh",
```
olibrook: What is the distinction between tests and unit-tests here?

amp/package.json
```diff
@@ -25,8 +27,9 @@
     "eslint-plugin-react": "6.10.3",
     "git-rev-sync": "1.9.1",
     "jest-cli": "19.0.2",
-    "mobify-amp-sdk": "ssh+git://git@github.com:mobify/mobify-amp-sdk.git#master",
+    "mobify-amp-sdk": "ssh+git://git@github.com:mobify/mobify-amp-sdk.git#add-image-component",
     "mobify-code-style": "2.8.2",
+    "mocha": "^3.4.1",
```
olibrook: Can we stick with Jest, which we are already using as a test runner?

amp/tests/unit/amp-validation.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-env node, mocha */
+
+import chai from 'chai'
+import childProcess from 'child_process'
+import http from 'http'
+
+chai.should()
+
+const handleAMPValidation = (process, done) => {
```
olibrook: I think you can do this in-process, which I think would be a little tidier than spawning a new one for each test.

First example here - https://www.npmjs.com/package/amphtml-validator

olibrook: Also, if anything, I would label these integration tests.

amp/tests/unit/amp-validation.js
```diff
@@ -0,0 +1,61 @@
...
+    before((done) => {
+        done()
+    })
+
+    after((done) => {
+        done()
+    })
+
+    describe('Merlins', function() {
+        const serviceUrl = `http://localhost:3000`
```
olibrook: I don't understand why we need to do the rendering in a separate process here...

amp/tests/unit-tests.sh
```diff
@@ -0,0 +1,14 @@
...
+# The follow makes sure all the processes spawned by 'npm run dev:serve' get cleaned up on exit
+trap "kill 0" EXIT
+
+webpack
+npm run dev:serve > /dev/null 2>&1 &
+
+# dev:serve needs a second to be ready
+sleep 1s
+
+mocha tests/unit --compilers js:babel-core/register
```
olibrook: Two points:

1) I think our experience with the PWA is that there are lots of partners on Windows machines. Trivial bash scripts like this could easily be written as node scripts, which we can assume they have installed more safely. I've been collecting these all together under `./scripts`/.

2) We've already got Jest going as a test runner. Is there a real need to add another one?

amp/package.json
kschmidtdev: Artifact of bringing in the validator tests from the AMP prototype - I'll consolidate with jest.

amp/package.json
kschmidtdev: I think so? I wasn't sure if it was suitable to these kinds of tests.

amp/tests/unit/amp-validation.js
kschmidtdev: That makes sense, I don't recall my reasoning for labelling them otherwise...

Also that in-process stuff is pretty awesome, it wasn't around when I originally set up testing with the prototype 😅

kschmidtdev:
@olibrook I think I addressed all your feedback, thanks for all of it!

kschmidtdev:
@olibrook @cole-sanderson Re-review please 🙏

amp/app/containers/pdp/container.jsx
```diff
@@ -1,8 +1,36 @@
-import Page from '../../components/page/'
+import React, {PropTypes} from 'react'
 import {connect} from 'react-redux'
+import AmpImage from 'mobify-amp-sdk/dist/components/amp-image'
+import AmpLightbox from '../../components/amp-lightbox'
+
+const containerClass = 't-home'
```
cole-sanderson: It should be `t-pdp`

amp/app/containers/plp/container.jsx
```diff
@@ -1,8 +1,36 @@
-import Page from '../../components/page/'
+import React, {PropTypes} from 'react'
 import {connect} from 'react-redux'
+import AmpImage from 'mobify-amp-sdk/dist/components/amp-image'
+import AmpLightbox from '../../components/amp-lightbox'
+
+const containerClass = 't-home'
```
cole-sanderson: it should be `t-plp`

amp/app/main.test.js
```diff
@@ -0,0 +1,52 @@
...
+    test('PDP', () => {
+        return new Promise((resolve, reject) => {
+            const req = {url: '/eye-of-newt.html'}
+            const res = resObject(resolve, reject)
+            const next = () => {}
+
+            return productDetailPage(req, res, next)
+        })
+    })
+})
```
olibrook: This is much, much nicer in-process. I find it a little odd combining the assertion with a mock response object though. Any chance we could split the two, so we have something like

```
  const req = mockRequest({url: 'eye-of-newt.html'})
  const res = mockResponse()
  handler(req, res)
  res.on('end', res => validateAmp(res.text))
```

I haven't used it, but this looks like a pretty good option for mocking Express responses, from the docs https://www.npmjs.com/package/express-mocks-http.


There's a little blog post about it here - https://www.terlici.com/2015/09/21/node-express-controller-testing.html

amp/package.json
```diff
@@ -44,6 +45,7 @@
     "sass-lint": "1.10.2",
     "sass-loader": "6.0.5",
     "source-map-support": "0.4.15",
+    "susy": "^2.2.12",
```
olibrook: I don't see any mention of this new lib in the PR. What is susy for?

https://www.npmjs.com/package/susy

kschmidtdev: The use of this dependency was introduced in #640 here: https://github.com/mobify/platform-scaffold/pull/640/files#diff-a10d7de4307feae95d383e6b24b07999R16 but wasn't added to the list of dependencies

olibrook: Ah. Okay.

amp/app/main.test.js
kschmidtdev: I've updated the tests to use that library - I had a lot of trouble trying to get anything resembling the article and the library README to actually work, please take a look at let me know what you think. FYI there is no real mock data at this point - in the future we'll probably mock the responses from merlinspotions.com but I'd like to address that in a future PR.

olibrook:
👍 Looks good to me

PR #620, opened 2017-05-19
Unify country and region fields

DanielBrookRoberge:
I discovered in preparing to rebuild the schema for checkout locations that the usage of the location (country and region) data was inconsistent between the three places it was used: cart shipping estimate, shipping address, and billing address. This PR splits those out into their own components, so that the logic can be shared and consistent.

 **JIRA**: Preparatory for checkout schema

## Changes
- 

## How to test-drive this PR
- Run `npm run dev`
- Add an item to the cart, and open the cart
- Check that the shipping estimate modal and form work correctly
- Check that the shipping address form works correctly
- Check that the billing address form works correctly


web/app/components/country-select/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+CountrySelect.defaultProps = {
+    countries: []
+}
+
+CountrySelect.propTypes = {
+    /**
+     * Adds values to the `class` attribute of the root element
+     */
+    className: PropTypes.string,
```
jeremywiebe: Is this optional?

web/app/components/country-select/test.js
```diff
@@ -0,0 +1,42 @@
...
+/* eslint-env jest */
+// import React from 'react'
+
+import CountrySelect from './index.jsx'
+
+test('CountrySelect exists', () => {
+    expect(CountrySelect).toBeDefined()
+})
+
+// test('CountrySelect renders without errors', () => {
```
jeremywiebe: Are these still wip or why are they commented out?

web/app/components/country-select/_base.scss
```diff
@@ -0,0 +1,33 @@
+// CountrySelect
+// ===
+//
+// Component description here
```
jeremywiebe: `// TODO ?`

web/app/components/region-field/_base.scss
```diff
@@ -0,0 +1,33 @@
+// RegionField
+// ===
+//
+// Component description here
```
jeremywiebe: `// TODO ?`

web/app/components/country-select/index.jsx
DanielBrookRoberge: It is. We don't need to write it into the doc string since styleguidist will mark `isRequired` props as required automatically.

web/app/components/region-field/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+}) => {
+    const classes = classNames('c-region-field', className)
+
+    if (regions.length === 0) {
+        return (
+            <ReduxForm.Field
+                className={classes}
+                component={Field}
+                name="region"
+                label="Region"
```
jeremywiebe: We're inconsistent with the label if there are regions and if there aren't.  Here we use "Region" but below we use "State/Province"

web/app/components/region-field/test.js
```diff
@@ -0,0 +1,42 @@
...
+/* eslint-env jest */
+// import React from 'react'
+
+import RegionField from './index.jsx'
+
+test('RegionField exists', () => {
+    expect(RegionField).toBeDefined()
+})
+
+// test('RegionField renders without errors', () => {
```
jeremywiebe: More commented out code...

web/app/components/region-field/index.jsx
```diff
@@ -0,0 +1,69 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
```
jeremywiebe: Given how similiar this is to the `CountrySelect` component, what do you think about naming them more closely: `RegionSelect`?

web/app/components/region-field/index.jsx
```diff
@@ -0,0 +1,69 @@
...
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+import React, {PropTypes} from 'react'
+import classNames from 'classnames'
+import * as ReduxForm from 'redux-form'
+import Field from 'progressive-web-sdk/dist/components/field'
+
+/**
+ *
```
jeremywiebe: Missing component description. I think this one is important as this component presents either a `<select>` element or an `<input>` depending on if there's regions or not.

web/app/components/country-select/test.js
DanielBrookRoberge: Since these components render a `ReduxForm.Field`, they throw errors if they aren't mounted within a redux form component. Not sure how to test them in that case, which is why the autogenerated tests are all commented out.

web/app/components/region-field/index.jsx
DanielBrookRoberge: This is on purpose; what we really need is some way to figure out the right term for the given country. IIRC the countries where we're likely to have this enumerated for are CA/US/AU(/MX?) where the State/Province term is relevant, while other countries where this is a free-form field probably don't have the term.

web/app/components/region-field/index.jsx
DanielBrookRoberge: I considered that, but since the region field can be either a select or a text input I thought it would be confusing.

web/app/components/country-select/_base.scss
```diff
@@ -0,0 +1,33 @@
+// CountrySelect
```
jeffkamo: If this file isn't being used, it can be nuked.

web/app/components/country-select/_theme.scss
```diff
@@ -0,0 +1,6 @@
+// CountrySelect Theme
```
jeffkamo: Local components don't need theme files, so this can be nuked.

web/app/components/country-select/README.md
```diff
@@ -0,0 +1,12 @@
...
+import CountrySelect from '../components/country-select'
+
+// SCSS import
+@import '../components/country-select/base';
+```
+
+
+## Example Usage
+
+    <CountrySelect text="Hear me roar!" />
```
jeffkamo: Is it possible to replace this with a real example?

web/app/components/country-select/README.md
```diff
@@ -0,0 +1,12 @@
+```js
+// JS import
+import CountrySelect from '../components/country-select'
+
+// SCSS import
+@import '../components/country-select/base';
```
jeffkamo: If there is no SCSS to import, this can be removed.

web/app/containers/cart/partials/cart-estimate-shipping.jsx
```diff
@@ -42,26 +44,11 @@ export const CartEstimateShippingModal = ({closeModal, isOpen, countries, stateP
             <div className="u-padding-md">
                 <form onSubmit={handleSubmit(submitEstimateShipping)}>
                     <FieldRow>
-                        <ReduxForm.Field component={Field} className="pw--has-select" name="countryId" label="Country">
-                            <select>
-                                {countries.map(({label, value}) => <option value={value} key={value}>{label}</option>)}
-                            </select>
-                        </ReduxForm.Field>
+                        <CountrySelect countries={countries} />
```
jeffkamo: Is there a reason we don't just make the CountrySelect and RegionField components connected instead of passing props down? Seems like these values are all coming from the same place in the store, i.e. `'../../../store/checkout/locations/selectors'`

web/app/components/country-select/README.md
jeffkamo: This gives the following error in the styleguide (`npm run styleguide`):

screen shot 2017-05-23 at 2 53 32 pm

web/app/components/country-select/test.js
jeffkamo: Would wrapping the mounting fields in a `ReduxForm.reduxForm` prevent the errors?

web/app/components/region-field/_base.scss
```diff
@@ -0,0 +1,33 @@
+// RegionField
```
jeffkamo: Like the country select, if this SCSS file isn't used then it can be removed.

web/app/components/region-field/_theme.scss
```diff
@@ -0,0 +1,6 @@
+// RegionField Theme
```
jeffkamo: Local components don't need theme files, so this can be removed.

(I suppose the generator needs to be tweaked to not output theme SCSS files like this for local components)

web/app/components/region-field/README.md
```diff
@@ -0,0 +1,12 @@
+```js
+// JS import
+import RegionField from '../components/region-field'
+
+// SCSS import
+@import '../components/region-field/base';
```
jeffkamo: This SCSS import can be removed since it has no styles.

web/app/components/region-field/README.md
```diff
@@ -0,0 +1,12 @@
...
+```
+
+
+## Example Usage
+
+    <RegionField text="Hear me roar!" />
```
jeffkamo: This example also has an error in the `npm run styleguide`:

screen shot 2017-05-23 at 2 56 55 pm

web/app/containers/cart/partials/cart-estimate-shipping.jsx
DanielBrookRoberge: I'm not connecting the `RegionField` since its selector needs to be provided with the form name to look at for the selected country when it is added to mapStateToProps. Maybe we can do that down the road. CountrySelect works great, though.

web/app/components/country-select/test.js
DanielBrookRoberge: Yes, it would, though that's not a simple thing to get set up. 

I'm going to defer these tests to later in the interest of getting the checkout schema put together. We probably need to set up a `scaffolding` for testing these kinds of components correctly.

web/app/components/country-select/README.md
```diff
@@ -0,0 +1,79 @@
...
+// JS import
+import CountrySelect from '../components/country-select'
+```
+
+## Example Usage
+
+This component must be rendered inside a ReduxForm.reduxForm component
+for correct connection to the store.
+
+    const ReduxForm = require('redux-form');
```
jeremywiebe: Mobify doc standard calls for these code blocks to use [code fences](https://github.com/mobify/mobify-code-style/blob/develop/docs/remarkrc#L5) (```jsx) as opposed to the 4-space indentation style.

web/app/components/region-field/index.jsx
jeremywiebe: Ok, that makes sense. We can look at making this more configurable based on partner feedback going forward as needed.

web/app/components/country-select/README.md
DanielBrookRoberge: I don't know how that interacts with styleguidist; this code isn't included as text but rather is extracted and run by the styleguide.

jeremywiebe:
Is there a way to avoid the CountrySelect showing up as `Connect(CountrySelect)` in Styleguidist?

image

DanielBrookRoberge:
Not as far as I know, it was a bit of a stretch to even get it working, unfortunately.

web/app/components/country-select/README.md
jeremywiebe: You're right.  Styleguidist doesn't like them.  Moving on. :)

PR #622, opened 2017-05-23
SFCC Connector cleanups

DanielBrookRoberge:
This PR contains miscellaneous improvements to the SFCC commands, including more usage of the `makeApiJsonRequest` utility. 

 **JIRA**: Preparatory for checkout schema

## Changes
- Add additional error handling to `makeApiJsonRequest`
- Use it in more places
- Use `requestCartData` more
- Tidy up a bit

## How to test-drive this PR
- Preview on salesforce
- See that the checkout still functions


web/app/integration-manager/_sfcc-connector/login/commands.js
```diff
@@ -107,27 +106,18 @@ export const logout = () => (dispatch) => {
...
-            }
-
-            // Creating a user doesn't sign them in automatically, so dispatch the login command
-            return dispatch(login(email, password, true))
-        })
 
+    return makeApiJsonRequest('/customers', requestBody, {method: 'POST'})
+        .catch(() => { throw new SubmissionError({_error: 'Unable to create account.'}) })
+        // Creating a user doesn't sign them in automatically, so dispatch the login command
+        .then(() => dispatch(login(email, password, true)))
```
jnurse: Should this then go above the catch? We probably don't need to bother trying to log in if the account wasn't created.

jnurse: Ah nevermind I totally read this wrong 😄

web/app/integration-manager/_sfcc-connector/login/commands.js
```diff
@@ -66,22 +65,22 @@ export const login = (username, password) => (dispatch) => {
         .then(() => makeApiRequest(`/customers/${customerID}/baskets`), {method: 'GET'})
```
DanielBrookRoberge: Yes, it should be. Thanks for the catch!

web/app/integration-manager/_sfcc-connector/cart/commands.js
```diff
@@ -6,71 +6,48 @@ import {makeApiRequest, makeApiJsonRequest, getAuthTokenPayload} from '../utils'
...
+            const requestBody = [{
+                product_id: productId,
+                quantity
+            }]
+
+            return makeApiJsonRequest(`/baskets/${basket.basket_id}/items`, requestBody, {method: 'POST'})
         })
-}
+        .catch(() => { throw new Error('Unable to add item to cart') })
+        .then((basket) => dispatch(handleCartData(basket)))
```
jeremywiebe: If the `.catch()` is executed, this `.then((basket) ...` would receive nothing, right?  Should we handle that so we don't dispatch `handleCartData(undefined)`?

DanielBrookRoberge: If the `.catch` is executed, the `.then` doesn't run, since the catch handler throws.

web/app/integration-manager/_sfcc-connector/home/commands.js
```diff
@@ -7,7 +7,5 @@ import {receiveHomeData} from '../../results'
 export const initHomePage = () => (dispatch) => {
     // Banners are being pulled from the bundle right now
     // so we just need an array with the correct number of objects
-    return new Promise(() => {
-        return dispatch(receiveHomeData({banners: [{}, {}, {}]}))
-    })
+    return Promise.resolve(dispatch(receiveHomeData({banners: [{}, {}, {}]})))
```
jeremywiebe: Non-blocking. But would it be clearer to break this into two lines?
```js
dispatch(receiveHomeData({banners: [{}, {}, {}]}))
return Promise.resolve()
```
The two are unrelated and I don't think we're intending to return the return value from `dispatch()` in this case.

web/app/integration-manager/_sfcc-connector/cart/commands.js
jeremywiebe: 🤦‍♂️  Right. I should take a break. I'm not reading code correctly anymore.  lol  

👍

PR #624, opened 2017-05-24
[Checkout Schema] Schematize locations

DanielBrookRoberge:
This PR converts the location data (country and region) in the checkout branch to a schematized form, and makes the necessary modifications in the rest of the app to match.

## Changes
- Add schema types and a new typed action
- Change the connectors to match the type and use the action
- Change the app to expect the new data.

## How to test-drive this PR
- Preview Merlin's and SFCC
- See that the regions and countries still work on the cart estimate shipping, shipping address, and billing address forms.


web/app/components/country-select/index.jsx
```diff
@@ -29,8 +29,8 @@ const CountrySelect = ({
             label="Country"
         >
             <select>
-                {countries.map(({label, value}) => (
-                    <option value={value} key={value}>
+                {countries.map(({label, id}) => (
```
jeremywiebe: This can use the new `CountrySelect` component, can't it?

web/app/components/region-field/index.jsx
```diff
@@ -44,8 +44,9 @@ const RegionField = ({
             label="State/Province"
         >
             <select>
-                {regions.map(({label, value}) => (
-                    <option value={value} key={label}>
+                <option value={null}>Please select a region, state, or province</option>
+                {regions.map(({label, id}) => (
```
jeremywiebe: Same comment about using `RegionField`.

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -53,8 +53,8 @@ const processCheckoutData = ($response) => (dispatch) => {
     const magentoFieldData = extractMagentoShippingStepData($response)
           .getIn(['children', 'shipping-address-fieldset', 'children'])
 
+    dispatch(receiveCheckoutLocations(parseLocations(magentoFieldData)))
     return dispatch(receiveCheckoutData({
```
jeremywiebe: I don't think we need a `return` here as this isn't a thunk we're dispatching.

web/app/integration-manager/_merlins-connector/checkout/parsers.js
```diff
@@ -22,18 +22,34 @@ export const parseShippingInitialValues = (shippingFieldData) => {
...
     if (!shippingStepData) {
         return {}
     }
 
     return {
-        locations: {
-            countries: shippingStepData.getIn(['country_id', 'options']),
-            regions: shippingStepData.getIn(['region_id', 'options'])
-        }
+        countries: shippingStepData
```
jeremywiebe: You have a knack for formatting this type of object construction really cleanly and clearly!  :)  Nice work.

web/app/integration-manager/_merlins-connector/checkout/parsers.js
```diff
@@ -22,18 +22,34 @@ export const parseShippingInitialValues = (shippingFieldData) => {
...
+            .toJS()
+            .map(({value, label, is_region_required, is_zipcode_optional}) => ({
+                id: value,
+                label,
+                regionRequired: !!is_region_required,
+                postcodeRequired: !is_zipcode_optional
+            })),
+        regions: shippingStepData
+            .getIn(['region_id', 'options'])
+            .slice(1)
```
jeremywiebe: Can you add a line-comment why we skip the first one (which is I think what's happening here)?

web/app/components/country-select/index.jsx
DanielBrookRoberge: No it can't (it's the implementation of `CountrySelect`)

web/app/components/region-field/index.jsx
DanielBrookRoberge: Same comment here too 😉

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: No, that's true; these plain actions resolve synchronously.

web/app/integration-manager/_merlins-connector/checkout/parsers.js
DanielBrookRoberge: Thanks!

web/app/integration-manager/_merlins-connector/checkout/parsers.js
DanielBrookRoberge: For sure!

web/app/integration-manager/_sfcc-connector/checkout/constants.js
```diff
@@ -4,318 +4,318 @@
...
         label: 'Armed Forces America (exc. Canada)',
-        value: 'AA'
+        id: 'AA'
     },
     {
-        country_id: 'us',
+        countryId: 'us',
         label: 'Armed Forces Canada',
-        value: 'AE'
+        id: 'AE'
```
jeremywiebe: Can we remove these "states" with duplicated id's?  Or consolidate them?  They continue to cause React `duplicate key` errors and in my experience it's extremely rare to even see these listed as states.  

The original list seems to have come partially from here: https://documentation.demandware.com/DOC2/topic/com.demandware.dochelp/DSS/EAStore-GetCountriesStates.html

DanielBrookRoberge: Yeah, we probably want to have unique IDs for these things. (I could write a runtypes constraint that checks for unique ids, even)

web/app/store/checkout/selectors.js
```diff
@@ -2,7 +2,30 @@
...
 export const getEmailAddress = createGetSelector(getCheckout, 'emailAddress')
+
+export const getLocations = createGetSelector(getCheckout, 'locations', Immutable.Map())
+export const getCountries = createGetSelector(getLocations, 'countries', Immutable.List())
+export const getRegions = createGetSelector(getLocations, 'regions', Immutable.List())
+
+export const getSelectedCountryID = (formKey) => createSelector(
+    getFormValues(formKey),
+    (values) => {
+        return values ? values.countryId : null
```
jeremywiebe: Just double-checking. Are `values` an Immutable.js object here or a plain JS object?

DanielBrookRoberge: It's plain JS, since it comes from the `form` branch which is plain JS.

web/app/components/country-select/index.jsx
jeremywiebe: 🤦‍♂️

web/app/integration-manager/_sfcc-connector/checkout/constants.js
jeremywiebe: Sounds good.  And this doesn't _have_ to be in this PR.  Actually maybe @Cdok can do this.  :)

web/app/integration-manager/_sfcc-connector/checkout/constants.js
```diff
@@ -4,318 +4,283 @@
...
-    },
-    {
-        country_id: 'us',
-        label: 'Armed Forces Pacific',
-        value: 'AP'
-    },
-    {
-        country_id: 'us',
-        label: 'Non-US/Other',
-        value: 'OTHER'
```
jnurse: Is there a reason why these were removed?

DanielBrookRoberge: https://github.com/mobify/platform-scaffold/pull/624/#discussion_r118542176 . I can revert that if you think it's important to keep them.

DanielBrookRoberge: ok apparently linking to Jeremy's comment didn't work.

jnurse: I'm not against removing them, but this content was pulled directly from the equivalent selects on the desktop site. Perhaps we could consolidate them instead? So we just have three armed forces options, one for each of the unique ids (AE, AP and AA). 

Maybe we should explore adding this sort of content to the configuration for a connector since it's possible not all SFCC sites will ship to these locations.

web/app/integration-manager/checkout/types.js
```diff
@@ -0,0 +1,26 @@
...
+import * as Runtypes from 'runtypes'
+import {Identifier, Text} from '../types'
+
+const CountryID = Identifier
+
+const Country = Runtypes.Record({
+    id: CountryID,
+    label: Text,
+    regionRequired: Runtypes.Boolean,
+    postcodeRequired: Runtypes.Boolean
```
jnurse: Are we missing these keys in the SFCC connector? I'm seeing some runType errors when I view the shipping page for SFCC.

web/app/integration-manager/_sfcc-connector/checkout/constants.js
DanielBrookRoberge: Yeah, we'll need to figure out something for configuring the SFCC connector for different shipping ranges; right now we've just got the US hard-coded. I'm going to restore and consolidate the Armed Forces options for now, but we'll probably need to sit down and figure out how we want this to work.

web/app/integration-manager/checkout/types.js
DanielBrookRoberge: Yes, we are. I'll get that fixed.

web/app/integration-manager/_sfcc-connector/checkout/constants.js
jeremywiebe: I was looking in the SFCC docs to see if these could be retrieved by API... there must be a way... haven't found it yet though.

jnurse:
👍

PR #626, opened 2017-05-24
[IM Extensibility] Top level custom results & reducers

jnurse:
This PR adds results & reducers that add a custom branch to the relevant top level keys on the state. The remaining top level branches (products, categories, modals etc) have been intentionally excluded since their shape and usage means a custom branch is most likely unnecessary. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-75
 **Linked PRs**: n/a

## Changes
- Added result & reducer for 'custom' key on the cart branch of the state
- Added result & reducer for 'custom' key on the checkout branch of the state
- Added result & reducer for 'custom' key on the user branch of the state
- Added reducer tests for cart, checkout & user custom branches
- Added selectors for cart, checkout and user custom content

## How to test-drive this PR
- checkout this branch
- run `npm run test`
- the tests should pass


DanielBrookRoberge:
This looks good; maybe it could also include selectors for the custom branch of each store branch?

jnurse:
@DanielBrookRoberge yes good point! I think we had discussed that previously I just forgot about it 😛

PR #633, opened 2017-05-26
A few cleanups from during the schema work

DanielBrookRoberge:
This PR fixes some minor issues that I have found during the schema work

 **JIRA**: Accessory to https://mobify.atlassian.net/browse/WEBDATA-69

## Changes
- Remove console.log statement from `getHighResImage`
- Rename company/apt toggle action to have distinct names on the two pages
- Use `validator` to check the email address on the checkout shipping page
- Clean up 

## How to test-drive this PR
- Preview Merlin's
- See that everything still works, including the company/apt address toggle
- See that you can use an email address on the shipping page with a `+` character in it.


web/app/containers/checkout-payment/actions.js
```diff
@@ -14,7 +14,7 @@ import {receiveCheckoutData} from '../../integration-manager/checkout/results'
 export const receiveContents = createAction('Received CheckoutPayment Contents')
 export const toggleFixedPlaceOrder = createAction('Toggled the fixed "Place Order" container', ['isFixedPlaceOrderShown'])
 export const toggleCardInputRadio = createAction('Toggled the card method radio input', ['isNewCardInputSelected'])
-export const toggleCompanyAptField = createAction('Showing the "Company" and "Apt #" fields', ['isCompanyOrAptShown'])
+export const toggleCompanyAptField = createAction('Toggled the "Company" and "Apt #" fields (Payment)', ['isCompanyOrAptShown'])
```
jeremywiebe: I wonder if we should keep this named exactly the same as the action for the shipping area? In the UI it's not reversible (you can't hide the fields) so the parameter is never passed as `false`.

DanielBrookRoberge: If we keep it named the same then it's interpreted by both reducers; we want to keep the flags separate so the actions need to have different names.

jeremywiebe: Discussed on Slack.  Will do API alignment between Payment and Shipping in a future PR.

PR #634, opened 2017-05-26
Store the app on-boarding carousel data in an array instead of an object

kschmidtdev:
So whether or not messaging is enabled, the order is correct. 

 **JIRA**: https://mobify.atlassian.net/browse/HYB-1044

## Changes
- Store the app on-boarding carousel data in an array instead of an object

## How to test-drive this PR
- `npm run deps` in `native/`
- Run `npm run dev` in `web/
- Run the iOS or Android app
- Observe the onboarding order is correct
- Toggle `messagingEnabled` in `package.json` in both `native/` and `web/`, re-run test, observe the order is still correct

### DIFF
--- a/web/app/containers/onboarding/container.jsx
+++ b/web/app/containers/onboarding/container.jsx
@@ -8,9 +8,9 @@ import CarouselItem from 'progressive-web-sdk/dist/components/carousel/carousel-
 import Carousel from 'progressive-web-sdk/dist/components/carousel'
 import Button from 'progressive-web-sdk/dist/components/button'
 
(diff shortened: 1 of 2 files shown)


jeremywiebe:
Tested and it looks like it mostly works. But when I run it in Chrome, I see these two errors in the console. 

screen shot 2017-05-26 at 3 18 06 pm

kschmidtdev:
@jeremywiebe I had a feeling there was something missing - and amazing that it works at all! Thanks I'll fix that up ✅

PR #639, opened 2017-05-26
[IM Extensibility] Connector Extension Set up

jnurse:
This PR adds a connector-extension to the scaffold and sets up the extensibility code in the integration manager. *Note:* Proper test/sample code will be added in another PR as described in the JIRA tickets

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-78 & https://mobify.atlassian.net/browse/WEBDATA-79

## Changes
- Added registerConnectorExtension function to the integration manager
- Added custom folder/commands file to the integration manager
- Set up basic structure for connector extension folder
- Added sample custom command and command override

## How to test-drive this PR
- Run `npm run dev`
- Preview the [SFCC Site](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Navigate to a PDP
- Update one of the variation selections (size or colour)
- Look in the console, you should see the text "Custom Command Dispatched"
- Click the add to cart button
- Look in the console, you should see the text "Add to Cart Override"

## To Do Before Merging:
- [x] Remove test code
- [x] Set connector back to Merlins


web/app/integration-manager/index.js
```diff
@@ -4,11 +4,27 @@
...
+let registeredConnector = {}
 
 export const registerConnector = (connector) => {
+    registeredConnector = connector
     commands.register(connector.commands)
     reducer.register(connector.reducer)
 }
 
+export const registerConnectorExtension = (extension) => {
+    registeredConnector.commands = merge(
```
jeremywiebe: Could we just use `Object.assign` here?  We already have 'babel-plugin-transform-object-assign' in package.json.

web/app/integration-manager/custom/commands.js
```diff
@@ -0,0 +1,12 @@
...
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+let customCommands = {}
+
+export const register = (commands) => {
+    customCommands = commands
+}
+
+
+export {customCommands}
```
jeremywiebe: I'm not sure this will work. The initial export is an empty object, but once `register` is called, that object changes and anything that's already imported this file will have the empty object instead of the registered commands.

jnurse: That's what I was seeing initially too, but when I tried it again the next day it worked. I must have made some small change between the two times, but I'm not sure what it was. 

Did you try running the steps to test and see an error?

DanielBrookRoberge: If the error comes back, we should be able to do `Object.assign(customCommands, commands)` rather than the the direct assignment and have it work.

web/app/integration-manager/index.js
jnurse: Not with how it's currently written. It needs to be a deep merge, and Object.assign does a shallow one. If we're concerned about importing lodash's merge function we could loop through the keys and merge each branch ourselves.

web/app/integration-manager/custom/commands.js
jeremywiebe: @jnurse I hadn't run the code when I commented, that was based on just reading it. I've tested it and the place it's being used now works. I'd still recommend we use DBR's suggestion otherwise we could have a race condition between importing the command and registering custom commands.

PR #641, opened 2017-05-26
remove hard coded url for removing cart

gsaray:
remove hard coded url for removing cart 

## Changes
- remove URL and navigate to cart using mini cart

## How to test-drive this PR
- Run `npm run dev`
- run `npm run test:e2e --tests`

### DIFF
--- a/web/tests/system/page-objects/cart.js
+++ b/web/tests/system/page-objects/cart.js
@@ -34,6 +34,7 @@ Cart.prototype.removeItems = function() {
     this.browser
         .preview()
         .log('Opening mini cart')
+        .waitForElementVisible(selectors.miniCart)
         .click(selectors.miniCart)
         .element('css selector', selectors.viewCart, (result) => {
             if (result.value && result.value.ELEMENT) {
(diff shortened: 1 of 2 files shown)


web/tests/system/page-objects/cart.js
```diff
@@ -3,11 +3,14 @@
 /* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
 
 const selectors = {
+    miniCart: '.qa-header__cart',
+    viewCart: '.t-mini-cart__content .c--tertiary',
```
jeffkamo: could/should these be replaced with `qa-` classes?

gsaray: I don't think it's necessary here, because there is only one `c--tertiary` button in the modal.

web/tests/system/page-objects/cart.js
```diff
@@ -29,21 +32,24 @@ Cart.prototype.removeItems = function() {
     // Remove all items from the cart
     const self = this
     this.browser
-        .log('Removing item')
-        .url('https://www.merlinspotions.com/checkout/cart/')
-        .waitForElementVisible(selectors.cartTemplateIdentifier)
-        .element('css selector', selectors.removeItem, (result) => {
+        .preview()
+        .log('Opening mini cart')
+        .click(selectors.miniCart)
```
ellenmobify: There can be a `.waitForElementVisible(selectors.miniCart)` above this line for stability.

PR #643, opened 2017-05-29
Add a readme to merlins connector

gsaray:
Make sure developers are not using merlins connector

## Changes
- Add readme to merlins connector

## How to test-drive this PR
- Run `npm run dev`


web/app/integration-manager/_merlins-connector/README.md
```diff
@@ -0,0 +1,2 @@
+# Mobify Merlins Connector
+- The purpose of this connector is for demo only and can be deleted if needed.
```
jeffkamo: Minor: perhaps `...if not needed`, since it will get deleted when in fact it is not needed? Or perhaps that bit can be deleted itself.

web/app/main.jsx
```diff
@@ -24,8 +24,9 @@ import configureStore from './store'
 // added to the markup in `loader.js`
 import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
 
-
+// DO NOT USE! This is an example connector that is for demo only
 import {Connector} from './integration-manager/_merlins-connector'
+
```
jeffkamo: This is probably pretty minor, but I'm not sure if we should add this space here. With the space, the connectors become a little too disassociated. I wonder if we can improve this section to better showcase that the developer "picks" a connector...

jeremywiebe: Initially the developer will be flipping these commented lines, but as soon as possible the generator will ask and maintain this file so the commented out line won't even exist. But I agree with your spacing comment.

PR #645, opened 2017-05-29
[IM Extension] custom commands

jnurse:
This PR adds a custom command that checks if a users email address is in use already to the platform scaffold. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-78
 **Linked PRs**: n/a

## Changes
- Moved isEmailAvailable merlins command to the connector extension
- Removed isEmailAvailable empty command from the SFCC connector

## How to test-drive this PR
- Run `npm run dev`
- add an item to the cart and proceed to the shipping page of the checkout
- enter an email address that is already being used for an account
- the login form should show along with a message asking the user to log in
- go back to the email address field and enter an email that's not being used for an account
- the login form and notification should disappear


jnurse:
Yeah I had the same thought, for now one option would be to comment out registerExtension when you switch to the SFCC connector, but that's not ideal.

PR #646, opened 2017-05-29
Clean up shipping estimate form

DanielBrookRoberge:
The cart estimate shipping form wasn't working correctly and was poorly factored for proceeding with address schematization. This PR fixes that.

 **JIRA**: Preparatory for https://mobify.atlassian.net/browse/WEBDATA-69

## Changes
- Change the input for the `fetchShippingMethodsEstimate` command to be an address rather than a form key
- Move all logic that prepares the address for the estimate requests into the Merlin's connector
- Refactor conditions in the estimate shipping modal UI.

## How to test-drive this PR
- (necessary config changes)
- Run `npm run dev`


web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -24,34 +24,33 @@ import {getCustomerEntityID} from '../selectors'
...
-    const currentState = getState()
-    const isLoggedIn = getIsLoggedIn(currentState)
-    const formValues = getFormValues(formKey)(currentState)
-    const entityID = getCustomerEntityID(currentState)
-    const registeredFieldNames = getFormRegisteredFields(formKey)(currentState).map(({name}) => name)
+const getCartBaseUrl = createSelector(
+    getIsLoggedIn,
+    getCustomerEntityID,
+    (isLoggedIn, entityID) => `/rest/default/V1/${isLoggedIn ? 'carts/mine' : `guest-carts/${entityID}`}`
+)
```
jnurse: Looks like this function is the same as the one in cart/commands.js. Can we refactor this into a utils or selectors file?

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -24,34 +24,33 @@ import {getCustomerEntityID} from '../selectors'
...
-                ...address
-            }
-
+        .then((responseJSON) => parseShippingMethods(responseJSON))
+        .then((shippingMethods) => {
             dispatch(receiveShippingMethods(shippingMethods))
-            dispatch(receiveShippingInitialValues({address: initialValues})) // set initial value for method
+            dispatch(receiveShippingInitialValues({address: {
+                shipping_method: shippingMethods[0].id,
+                postcode: address.postcode
```
jnurse: Can we set the initial value for the region here too?

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: Oops I forgot. I can do that.

web/app/integration-manager/_merlins-connector/checkout/commands.js
DanielBrookRoberge: Setting initial values here is still not really working correctly; I only added the postcode to the state here so that it will show up on the cart page for the time being. 

This will be a focus of the main address schema PR so I thought I'd just leave it this way for now.

DanielBrookRoberge:
The SFCC command seemed to be working with no changes; I guess something changed after I tested it.

DanielBrookRoberge:
The estimate form on SFCC is now partially fixed; the tax calculation fails since it's not implemented, and the shipping methods are dispatched to the store but not displayed because of the same initial value confusion that hit Merlin's above. So that'll be fixed in the address schema work. Thanks for catching the issue!

jnurse:
👍

PR #652, opened 2017-05-29
[Integration Manager] - command override example

Cdok:
# [Integration Manager] - command override example

Overrides the update cart quantity command on Merlin's as an example of command overrides.

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-79
 **Linked PRs**: https://github.com/mobify/platform-scaffold/pull/639

## Changes
- Overrides the update cart quantity command on Merlin's

## How to test-drive this PR
- Use merlin's connector
- Run `npm run dev`
- Add an item to your cart
- Ensure updating cart quantity still works!

### DIFF
--- a/web/app/connector-extension/cart/command-overrides.js
+++ b/web/app/connector-extension/cart/command-overrides.js
@@ -1,3 +1,34 @@
-export const addToCart = () => (dispatch) => {
-    console.log('%cAdd to Cart Override', 'color: green; font-size: x-large')
+// import {submitForm} from '../../utils'
+const UPDATE_ITEM_URL = '/checkout/cart/updatePost/'
+import {getCart} from '../../integration-manager/cart/commands'
+import {makeFormEncodedRequest} from 'progressive-web-sdk/dist/utils/fetch-utils'
+
+const getCookieValue = (cookieName) => {
+    const result = document.cookie.replace(new RegExp(`(?:(?:^|.*;\\s*)${cookieName}\\s*\\=\\s*([^;]*).*$)|^.*$`), '$1')
+    return result
+}
+
+const submitForm = (url, data, options) => {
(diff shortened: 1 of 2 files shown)


web/app/connector-extension/cart/command-overrides.js
```diff
@@ -2,3 +2,23 @@
 /* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
 /* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
 
+import {getCart} from '../../integration-manager/cart/commands'
+import {submitForm} from '../../integration-manager/_merlins-connector/utils'
```
Cdok: Importing directly from merlin's connector here is a bit weird - but since we're overriding commands _specific_ to this connector, and we can expect a partner to want to take advantage of some of the connector's utils in their custom overrides anyways, it makes sense in this case.

Jen and I briefly discussed this, but I'm open to feedback on importing directly from connectors.

jeremywiebe: This raises an interesting point. In addition to the core set of commands a shipping connector has, we'll probably want to export some base set of utilities also so that extensions can leverage them. 

I don't think importing directly from the connector is unnatural though. In my head extensions are very connector-specific. If we want to make an extension that works on all connectors, to me that's a signal that it should become part of the core IM api.

Cdok: It definitely feels like there should be a set of utils from each connector that are exported and documented.

Maybe they could be kept in an`api.md` somewhere in the scaffold where a partner could see some quick documentation on the exposed utils.

I agree that importing directly from the connectors in extensions still makes sense since they're specific to each connector.

PR #653, opened 2017-05-29
Add missing Select style to PDP Variants

jeffkamo:
Ensure that all Select elements are styled correctly. Each `<select>` tag should be wrapped in a `<Field>` (technically a Redux form field) component, with the `pw--has-select` modifier class. For example...

```
<ReduxForm.Field className="pw--has-select" ...>
    <select>{...}</select>
</ReduxForm.Field>
```

## Changes
- Add missing `pw--has-select` class to `select` field

## How to test-drive this PR
- In `app/main.jsx`, comment out the Merlin's Potions connector, and uncomment the SFCC connector
- Run `npm run dev`
- Preview [Merlin's Potions SFCC](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Visit any PDP with product variants
- **Verify** that the variant select dropdowns are styled correctly, and consistently with all other select fields (i.e. as those seen in the PLP/Search sort field, chekcout shipping/payment, and the cart page's estimate shipping modal)


web/app/containers/product-details/partials/product-details-variations.jsx
```diff
@@ -9,31 +9,34 @@ import {createPropsSelector} from 'reselect-immutable-helpers'
...
                 <ReduxForm.Field
                     label={label}
                     name={id}
                     component={Field}
-                    key={id}
-                    className="u-margin-bottom-lg u-margin-top"
-                    customEventHandlers={{
-                        onChange: onVariationChange
-                    }}
+                    className="pw--has-select"
```
tony-dinh: Judging by this, I would guess that local themes for `pw` components also use the `pw--` namespace?

These changes look good to me

tony-dinh: May need to re-run the CI tests though

jeffkamo: @tony-dinh that is correct. The way I see it, SDK components use `pw-` as styling hooks, and if the developer wants to create some of their own custom hooks, they should probably stick to the same prefixes to avoid confusing other developers between truly local components from customized third-party components.

PR #655, opened 2017-05-30
initial commit to fix hardcoded page title

gsaray:
Remove hardcoded header-title

## Changes
- remove hardcoded title and use the one in redux store

## How to test-drive this PR
- Run `npm run dev`


jeremywiebe:
Hmm... In the new world of Integration Manager, we probably shouldn't have the concept of a page title being parsed out of the connector. API-based projects most likely won't be able to provide this. I'd expect the page title to be derived from the category or product that is being rendered.

Cdok:
I think `pageTitle` is a misleading name for the key here - maybe `appName` is more suitable.

Right now it's hard-coded into the header, but I'm not sure that it makes sense to have the appName in the Redux store if it's only going to be used once in the header.

Having the app name hard-coded instead of HTML parsing / hitting an API seems like a fair trade-off, as the app name is unlikely to ever change for a project.

jeffkamo:
As discussed offline, we want to replace this PR with a Generator ([old](https://github.com/mobify/generator-progressive-web), [new](https://github.com/mobify/platform-generator)) change by doing something like a string replace...

```
{/*Generator_Replace_Start*/}Merlin's Potions{/*Generator_Replace_End*/}
```

The above get's replace by the generator into the user defined title at generation time.

PR #659, opened 2017-05-31
Analytics add to cart

sketchytechky:
# Adding `addToCart`, `removeFromCart`

*Open for visibility*

 **JIRA**:  [MTT-631](https://mobify.atlassian.net/browse/MTT-631)
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/650

## Changes
- [x] Add the add to Cart event

## Note on implementation

We ended up hooking onto the `commands` in IM instead on our original plan on `results`.  This is needed since IM only have the 2 type of actions (`receiveCartContents`, `receiveCartCustomContent`) neither of which is specific enough for our need

## How to test-drive this PR

Link to SDK on https://github.com/mobify/progressive-web-sdk/pull/650.  Data sent to EE `s.gif` should not be malformed


web/app/integration-manager/cart/commands.js
```diff
@@ -19,12 +23,26 @@ export const initCartPage = (url, routeName) => connector.initCartPage(url, rout
...
+)
+
 /**
  * Adds a product to the cart
  * @param productId {string} The product's ID
  * @param quantity {number} The quantity to add
  */
-export const addToCart = (productId, quantity) => connector.addToCart(productId, quantity)
+export const addToCart = (productId, quantity) => (dispatch, getState) => {
+    return connector.addToCart(productId, quantity)(dispatch, getState)
```
jeremywiebe: We can use dispatch here instead of manually invoking the thunk. 
```js
return dispatch(connector.addToCart(productId, quantity))
    .then(...)
```

jeremywiebe: I think we'll have to do this a bit more complicated though.  This function has to resolve to whatever the connector thunk resolves to.  So we'll probably have to start by capturing the returned Promise in a variable... then `.then()` on it and at the end of the chain, return that original promise.  Something like this:

```js
    const connectorResult = dispatch(connector.addToCart(productId, quantity))
    return connectorResult.then((cart) => {
        const cartCount = getCartSummaryCount(getState())
        const cartAction = createCartAction(cartCount)
        dispatch(cartAction)
        return connectorResult
    })
}
```

jnurse: I think actually, we just need to make sure that whatever the command returns, the cart object in this case, is returned by the `then` function. So something like this should work:

```
    return dispatch(connector.addToCart(productId, quantity)).then((cart) => {
        const cartCount = getCartSummaryCount(getState())
        const cartAction = createCartAction(cartCount)
        dispatch(cartAction)

        /************ The important line ******************/
        return cart
    })
}f
```

jeremywiebe: Ah, yes.  That's simpler! Good call.

sketchytechky: @jnurse @jeremywiebe - That's some great suggestions there - thanks! 🙏🏻

web/app/integration-manager/cart/commands.js
```diff
@@ -19,12 +23,57 @@ export const initCartPage = (url, routeName) => connector.initCartPage(url, rout
  */
 export const getCart = () => connector.getCart()
 
+const createCartAction = createActionWithAnalytics(
+    'Add to cart complete',
+    ['cart_count'],
```
wizardlyhel: This must be an empty array. We should not be changing redux state as a pure analytics action.

web/app/integration-manager/cart/commands.js
```diff
@@ -19,12 +23,57 @@ export const initCartPage = (url, routeName) => connector.initCartPage(url, rout
...
+        if (cartItems) {
+            const matchedItem = cartItems.find((item) => (item.id === productId))
+            if (matchedItem) {
+                matchedProduct = {
+                    [Product.ID]: matchedItem.id,
+                    [Product.NAME]: matchedItem.title,
+                    [Product.PRICE]: matchedItem.price
+                }
+            }
+        }
```
wizardlyhel: Use the force of SELECTORS~

This whole block can go away with the following:
```jsx
import {getProductById} from '../../../store/products/selectors'

const matchedProduct = getProductById(productId)(getState())
```

web/app/integration-manager/cart/commands.js
```diff
@@ -19,12 +23,57 @@ export const initCartPage = (url, routeName) => connector.initCartPage(url, rout
...
+
+        // find the cartItems that have the same productId
+        const cartAction = createCartAction(
+            cartCount,
+            subtotal,
+            matchedProduct
+        )
+        dispatch(cartAction)
```
wizardlyhel: Do this instead
```
dispatch(sendCartAnalytics(cartCount, subtotal, product))
```
And the above `createCartAction` can be like
```jsx
const sendAddToCartAnalytics = createActionWithAnalytics(
    'Add to cart complete',
    [],
    EVENT_ACTION.addToCart,
    (count, subtotal, product) => ({
        cart: new ShoppingList ({
            [ShoppingList.TYPE]: 'cart',
            count,
            subtotal
        }),
        product: new Product ({
            ...product,
            [Product.NAME]: product.title
        })
    })
)
```

sketchytechky: Nice - that's way cleaner 😄 totally going to steal this

wizardlyhel:
Getting 2 cart items when there should only be 1

screen shot 2017-06-01 at 3 17 33 pm

Not seeing MobifyGA connector firing add to cart analytic properly. Missing product data.

web/app/integration-manager/cart/commands.js
```diff
@@ -21,13 +25,51 @@ export const initCartPage = (url, routeName) => connector.initCartPage(url, rout
  */
 export const getCart = () => connector.getCart()
 
+const sendAddToCartAnalytics = createActionWithAnalytics(
+    'Add to cart complete',
```
wizardlyhel: Change the name of this action to `Send cart analytics`. 'Add to cart complete' is being used elsewhere.

web/app/integration-manager/cart/commands.js
```diff
@@ -21,20 +26,54 @@ export const initCartPage = (url, routeName) => connector.initCartPage(url, rout
...
+
+const dispatchCartAnalytics = (action, dispatch, getState, id) => {
+    const currentState = getState()
+    const cartCount = getCartSummaryCount(currentState)
+    const subtotal = getSubtotal(currentState)
+
+    if (action === EVENT_ACTION.addToCart) {
+        dispatch(sendAddToCartAnalytics(cartCount, subtotal, new Product(getProductById(id)(currentState).toJS())))
+    } else {
+        dispatch(sendRemoveFromCartAnalytics(cartCount, subtotal))
```
sketchytechky: Can we use the same trick with `new Product(getProductById(id)(currentState).toJS()))` here?

sketchytechky: I'm looking at this

wizardlyhel: Not really ... cause it's itemId ... which is the id for the item in your checkout: Unrelated to product id

PR #666, opened 2017-05-31
Add appStart analytics event

jansepar:
**Linked PRs**: https://github.com/mobify/lancome-progressive/pull/950

## Changes
- Adds sandy to the loader, and triggers a "loading" event which also contains the timing point of how long it took to get there from the start of the page load. This is important so that we can compare the % of people who leave to the % of people who leave in AMP pages.

In terms of the size difference, this increases the gzipped size from 4.1KB, which while it would be ideal to be smaller, isn't that large of a change. Alternatively we could create a stripped down sandy client lib, but that would be far more work. The Engagement team is looking to make the sandy client much smaller in the near future.

## How to test-drive this PR
- Run the project (new dependency added so make sure to run npm install)
- Go to the Network tab in DevTools and filter for "action". You should see 2 requests - one of them should contain the "appStart" timing point. Also, you should see that the sessionid and clientid are shared the same between all requests.


web/app/loader.js
```diff
@@ -153,7 +154,25 @@ const attemptToInitializeApp = () => {
         }
     }
 
+    const triggerAppStartEvent = () => {
+        // Gather analytics information indicating loading is happening, so we can determine
+        // the % of people who don't make it to the pageview.
+        Sandy.init(window)
+        Sandy.create('lancome', 'auto')
```
wizardlyhel: Require project slug replacement here

jansepar: lol whooooops copy and pasta fail.

wizardlyhel:
Looks good aside for that mandatory change

sketchytechky:
looks good tested manually.  LGTM - 👍 

image

PR #667, opened 2017-06-01
Install correct version of node-http-mocks, supports event listeners

olibrook:
## Changes
I think `express-http-mocks` was renamed to `node-http-mocks`. That is the latest version and supports event listeners on responses. By listening to those events it's much easier to wrap the response in a promise for easier testing.

Also, I changed the setup code to only initialize the amp validator once per test run.

## How to test-drive this PR
- Eyeball code
- Run `npm run test`


amp/app/main.test.js
```diff
@@ -1,71 +1,56 @@
...
+        return new Promise(resolve => {
+            const res = httpMocks.createResponse({
+                eventEmitter: events.EventEmitter
+            })
+
+            app.handle(req, res)
+
+            res.on('end', () => resolve(res))
+            res.on('send', () => resolve(res))
+
```
kschmidtdev: 🔫

amp/app/main.test.js
```diff
@@ -1,71 +1,56 @@
...
-    const makeResponse = () => {
-        return httpMocks.createResponse({
-            eventEmitter: events.eventEmitter
+    const handle = (req) => {
+        return new Promise(resolve => {
+            const res = httpMocks.createResponse({
+                eventEmitter: events.EventEmitter
+            })
+
+            app.handle(req, res)
```
kschmidtdev: This ends up executing after the next few lines, right? 😄

olibrook: Oh, good point. Only if it's async. I've reordered them.

PR #669, opened 2017-06-01
Fix product variants for item in cart incorrect

Cdok:
# Product variants for item in cart incorrect

Previously, every cart item would show `Maroon, XL` for color and size. This PR shows the correct selected variant in the cart.

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-91

## Changes
- Added selected variant options to the cart product list view

## How to test-drive this PR
- Use SFCC connector
- Run `npm run dev`
- Add some item to cart with variants selected
- Add another item with different variants
- Ensure that the cart is showing the correct variants


web/app/integration-manager/_sfcc-connector/cart/commands.js
```diff
@@ -10,18 +10,17 @@ export const getCart = () => (dispatch) =>
...
-export const addToCart = (productId, quantity) => (dispatch) => (
+export const addToCart = (productId, quantity, variants) => (dispatch) => (
     createBasket()
         .then((basket) => {
             const requestBody = [{
                 product_id: productId,
                 quantity
             }]
-
             return makeApiJsonRequest(`/baskets/${basket.basket_id}/items`, requestBody, {method: 'POST'})
```
Cdok: If there's a way in SFCC to get the variant from passing it in through in the request body here, that would eliminate the need for passing variants along through the `addToCart` pipeline here.

web/app/integration-manager/_sfcc-connector/cart/commands.js
```diff
@@ -21,7 +21,7 @@ export const addToCart = (productId, quantity) => (dispatch) => (
...
```
Cdok: If there's a way in SFCC to get the variant from passing it in through in the request body here, that would eliminate the need for passing variants along through the `addToCart` pipeline. The selected variation would instead be a key on the `basket` object from the response.

web/app/integration-manager/cart/commands.js
```diff
@@ -27,7 +27,7 @@ export const getCart = () => connector.getCart()
  * @param productId {string} The product's ID
  * @param quantity {number} The quantity to add
  */
-export const addToCart = (productId, quantity) => connector.addToCart(productId, quantity)
+export const addToCart = (productId, quantity, variants) => connector.addToCart(productId, quantity, variants)
```
Cdok: Not sure if adding a new param to the IM `addToCart` command makes _perfect_ sense - but we might not be able to capture the selected variant from just `productId` on Merlin's/SFCC (as mentioned in my previous comment)

I'm open to discuss a different approach to this 😄

Cdok:
**NOTE:** This PR should wait for variations to be implemented on Merlin's before merging

PR #671, opened 2017-06-01
Fix checkout confirmation reducer

Cdok:
# Fix checkout confirmation reducer

The reducer for checkout confirmation was incorrectly looking at the thunk action instead of `realReceiveCheckoutConfirmationData`

 **JIRA**: N/A
 **Linked PRs**: https://github.com/mobify/platform-scaffold/pull/556

## Changes
- Exported correct confirmation action
- Added above action to checkout confirmation reducer

screen shot 2017-06-01 at 11 16 13 am

## How to test-drive this PR
- SFCC Connector
- Run `npm run dev`
- Ensure a full checkout flow results in a confirmation page


jnurse:
I'm not getting the order number on the confirmation page, I think it's because the data is being nested under `confirmationData` in the redux store so our selectors don't work.

Cdok:
@jnurse it looks like the order number isn't showing up in Merlin's, but I'm seeing it in SFCC.

> Merlin's 
screen shot 2017-06-01 at 3 03 35 pm

> SFCC
screen shot 2017-06-01 at 3 01 40 pm

web/app/containers/checkout-confirmation/selectors.js
```diff
@@ -9,5 +9,10 @@ import {getUi} from '../../store/selectors'
 export const getCheckoutConfirmation = createSelector(getUi, ({checkoutConfirmation}) => checkoutConfirmation)
 
 export const getIsRegistrationFormHidden = createGetSelector(getCheckoutConfirmation, 'isRegistrationFormHidden')
-export const getOrderNumber = createGetSelector(getCheckoutConfirmation, 'orderNumber')
+export const getConfirmationData = createGetSelector(getCheckoutConfirmation, 'confirmationData')
+
+export const getOrderNumber = createSelector(getConfirmationData, (data) => {
+    return data ? data.get('orderNumber') : null
```
Cdok: Since this `realReceiveCheckoutConfirmationData` is (very?) async, on routechange the `confirmationData` key is doesn't exist when the selectors are first hit. The order number actually flashes on the screen after page load.

Without this check I was getting an error "Cannot `get` from undefined" from immutable. Is there a more elegant way of doing this instead of the ternary @jnurse  @jeremywiebe?

PR #673, opened 2017-06-01
[Checkout Schema] fix saved addresses fetch & update saved address in schema

jnurse:
This PR moves the merlins saved address fetch into the merlins connector, fixes some bugs related to submitting shipping with a saved address and moves saved addresses into the "storedAddresses" key in the store. 

 **Linked PRs**: n/a

## Changes
- Moved saved address fetch into merlins connector
- Called saved address fetch command in initCheckoutShipping command
- Moved logic to get the users saved address details to the submit shipping action, not merlins command
- Renamed the saved address key in the checkout store to "storedAddresses"

## How to test-drive this PR
- Run `npm run dev`
- login as a user with saved addresses (eg. mobifyqa@gmail.com)
- add an item to your cart
- proceed through the checkout (using a saved address on the shipping page)
- log out and try placing an order as an unknown users (it should still work)
- your order should be placed correctly
- switch to the SFCC connector
- preview the [SFCC Site](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- login to the site
- add an item to your cart and proceed through the checkout
*saved addresses have not been implemented in the SFCC connector yet, but you should still be able to checkout after filling in correct details*


DanielBrookRoberge:
I don't know if it's in scope here to fix, but I definitely had a long flash of the plain address form on the shipping page while the saved addresses were loading.

DanielBrookRoberge:
Submitting the payments page appears to be broken: 
screen shot 2017-06-01 at 16 43 27
screen shot 2017-06-01 at 16 43 33

jnurse:
@DanielBrookRoberge I see the same thing, I'd say it's probably out of scope for this work but I could move the saved address request up before the shipping methods estimate one since shipping methods are off screen initially. It won't resolve the issue, but it should make it a little bit better.

jnurse:
I can't reproduce the payment error. We'll need to get someone else to try it and see if they can reproduce that behaviour.

jeremywiebe:
On the Merlin's connector it's working good, except if I select a saved address and moved to billing and then click on the *Shipping* step (from the top "checkout progress" area, the address I had selected isn't selected again.
 
oco4uybueu

web/app/containers/checkout-shipping/actions.js
```diff
@@ -57,9 +59,23 @@ export const submitSignIn = () => (dispatch, getState) => {
...
+
 export const submitShipping = () => (dispatch, getState) => {
-    const currentState = getState()
-    const formValues = getShippingFormValues(currentState)
+    const shippingSelections = submitShippingSelector(getState())
+    let formValues = shippingSelections.formValues
+    const savedAddress = shippingSelections.address
+    if (savedAddress) {
+        formValues = {
+            ...formValues,
```
jeremywiebe: I might not be reading this right, but should the `...formValues` go after the `...savedAddress` so that the form values can override?

jnurse:
@jeremywiebe the correct saved address should show now when you navigate back to the page.

jeremywiebe:
👍  Can confirm.  Works great now!

PR #677, opened 2017-06-01
[Checkout Schema] Add types to checkout results

DanielBrookRoberge:
This PR adds an address type definition to the address action and also cleans up some of the other results. 

## Changes
- Add types to some of the checkout results
- Remove the generic `receiveCheckoutData` action.

## How to test-drive this PR
- Preview Merlin's and SFCC
- See that the checkout still proceeds.


jnurse:
Looks like we're storing the billing address in the wrong format. I get this error when I try submitting my order from the payment page: 
screen shot 2017-06-01 at 4 38 01 pm

DanielBrookRoberge:
Was that with billing === shipping or billing !== shipping?

jnurse:
billing === shipping I think

jeremywiebe:
@jnurse @DanielBrookRoberge We were dispatching the `address` we send to the back-end to `receiveBillingAddress`. It has the `street` array instead of `addressLine1` and `addressLine2`.  

I've removed that from the connector for now as it looks like we typically dispatch the updated data that comes from forms in the "UI actions" instead of in the connector.

web/app/containers/checkout-shipping/actions.js
```diff
@@ -104,8 +104,8 @@ export const submitShipping = () => (dispatch, getState) => {
         telephone,
     }
 
-    if (username) {
-        dispatch(receiveCheckoutData({emailAddress: username}))
+    if (formValues.username) {
```
jnurse: We're destructuring formValues on line 79, so this can probably just be `username` instead of `formValues.username`

PR #682, opened 2017-06-02
Remove double download

jansepar:
## Changes
- Uses HTML from the captured document on first load instead of re-fetching the page the user is already on. This helps to bring the overall load times down!

## How to test-drive this PR
- Run `npm run dev`
- Load the homepage
- Ensure that there isn't a second request for the homepage

### DIFF
--- a/web/app/integration-manager/_merlins-connector/app/commands.js
+++ b/web/app/integration-manager/_merlins-connector/app/commands.js
@@ -21,8 +21,25 @@ import {
     setLoggedIn
 } from '../../results'
 
-export const fetchPageData = (url) => (dispatch) => (
-    makeRequest(url)
+const requestCapturedDoc = () => {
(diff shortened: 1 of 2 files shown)


web/app/integration-manager/_merlins-connector/app/commands.js
```diff
@@ -21,8 +21,25 @@ import {
     setLoggedIn
 } from '../../results'
 
-export const fetchPageData = (url) => (dispatch) => (
-    makeRequest(url)
+const requestCapturedDoc = () => {
```
jeremywiebe: @jansepar I wonder if this should become a utility function in the SDK so that custom connectors using an HTML-based approach can use it too.  Or maybe it's a small enough snippet the they can just copy it in. 🤔   

(☑️  I vote we merge this either way and deal with moving to SDK in a future PR)

PR #686, opened 2017-06-02
Add missing SFCC category routes

drewzboto:
JIRA: WEBDATA-105

How to test: 
Follow bug report in WEBDATA-105


jnurse:
I'm just seeing the skeleton on the gift certificates page, shouldn't this be the empty PLP page instead? 

screen shot 2017-06-02 at 4 57 46 pm

drewzboto:
Yep, that should be a different story/fix to handle empty categories though

PR #687, opened 2017-06-02
PDP 404 fixes & variant fallbacks

Cdok:
Fix image filtering / PDP 404 errors

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-104

## Changes
- Fix PDP productID regex
- add checks for if variants are present or not 

## How to test-drive this PR
- Use sfcc connector
- Run `npm run dev`
- Try navigating to men -> unisex gloves, or grand theft auto under electronics, and ensure the API request doesn't 404 and that you can proceed with checkout

### DIFF
--- a/web/app/integration-manager/_sfcc-connector/parsers.js
+++ b/web/app/integration-manager/_sfcc-connector/parsers.js
@@ -54,7 +54,7 @@ export const parseProductDetails = ({id, name, price, inventory, long_descriptio
         available: inventory.orderable,
         thumbnail: images[0],
         images,
-        initialValues: variants ? setInitialVariantValues(variants, id, variation_attributes) : [],
+        initialValues: variants ? setInitialVariantValues(variants, id, variation_attributes) : {},
         variationCategories: variants ? parseVariationCategories(variation_attributes) : [],
         variants: variants ? variants.map(({product_id, variation_values}) => {
             return {
@@ -91,6 +91,8 @@ export const parseBasketContents = ({product_items, product_sub_total, product_t
         orderTotal: order_total
     }
 }
+
+/* eslint-enable camelcase */
 
 export const getCurrentProductID = (url) => {
     let productID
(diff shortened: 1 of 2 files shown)


web/app/integration-manager/_sfcc-connector/parsers.js
```diff
@@ -41,7 +41,6 @@ const setInitialVariantValues = (variants, id, variationCategories) => {
     return defaultVariant
 }
 
-/* eslint-enable camelcase */
```
jnurse: Are we not re-enabling camelcase at all for this file?

web/app/integration-manager/_sfcc-connector/parsers.js
```diff
@@ -55,19 +54,18 @@ export const parseProductDetails = ({id, name, price, inventory, long_descriptio
         available: inventory.orderable,
         thumbnail: images[0],
         images,
-        initialValues: setInitialVariantValues(variants, id, variation_attributes),
-        variationCategories: parseVariationCategories(variation_attributes),
-        variants: variants.map(({product_id, variation_values}) => {
+        initialValues: variants ? setInitialVariantValues(variants, id, variation_attributes) : [],
+        variationCategories: variants ? parseVariationCategories(variation_attributes) : [],
+        variants: variants ? variants.map(({product_id, variation_values}) => {
```
jnurse: If we're ok with these coming back undefined, you could do this:

`variants && setInitialVariantValues(variants, id, variation_attributes)`

If not the initialValues key should return an empty object instead of an empty array so we don't get an error from runtypes.

web/app/integration-manager/_sfcc-connector/parsers.js
Cdok: Oops. I moved it back to after our last `snake_case` usage.

web/app/integration-manager/_sfcc-connector/parsers.js
Cdok: Switched to returning an empty object for `initalValues` 🎉

Cdok:
@jnurse

I've changed
```diff
- productDetailsData.variants
+ productDetailsData.variants.length
```

in the `initProductDetailsPage` function, which resolved that chrome error for me.

PR #690, opened 2017-06-05
Update sfcc connector storage

jnurse:
Not all browsers support sessionStorage. This PR adds a function to check if session storage is supported and if it's not use cookies instead to track SFCC auth and basket data. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-98
 **Linked PRs**: n/a

## Changes
- Added function to check if sessionStorage is supported
- Replace calls to set/get/remove from sessionStorage with function that will use cookies if sessionStorage is unsupported

## How to test-drive this PR
- Run `npm run dev`
- Preview [SFCC site](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1) in a private Safari window
- Navigate through the site and confirm it works correctly (auth & basket data will be stored in cookies)
-  Preview [SFCC site](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1) again but in a regular chrome window
- Navigate through the site and confirm is works correctly (auth & basket data will be stored in sessionStorage)

### DIFF
--- a/web/app/integration-manager/_sfcc-connector/utils.js
+++ b/web/app/integration-manager/_sfcc-connector/utils.js
@@ -69,24 +69,24 @@ const removeItemFromBrowserStorage = (keyName) => {
 
 export const storeAuthToken = (authorization) => {
     if (authorization) {
-        window.sessionStorage.setItem(AUTH_KEY_NAME, authorization)
+        setItemInBrowserStorage(AUTH_KEY_NAME, authorization)
     }
 }
 
 export const getAuthToken = () => {
-    return window.sessionStorage.getItem(AUTH_KEY_NAME)
+    return getItemFromBrowserStorage(AUTH_KEY_NAME)
 }
 
 export const deleteAuthToken = () => {
-    window.sessionStorage.removeItem(AUTH_KEY_NAME)
+    removeItemFromBrowserStorage(AUTH_KEY_NAME)
 }
 
 export const deleteBasketID = () => {
-    window.sessionStorage.removeItem(BASKET_KEY_NAME)
+    removeItemFromBrowserStorage(BASKET_KEY_NAME)
 }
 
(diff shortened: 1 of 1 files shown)


jeremywiebe:
I noticed @fractaltheory pulled in the `store` npm module in the SDK [here](https://github.com/mobify/progressive-web-sdk/pull/646).  Would it make sense to use that here also so we don't have to keep adjusting how we store these values?

jnurse:
@jeremywiebe it might make sense to use that here, but it will require a bit of work to get it set up the way we need. The store module's cookie storage sets an expiry date and we use session cookies. It might be less brittle to set up an explicit expiry value, but we'll need to make some changes to get it working.

web/app/integration-manager/_sfcc-connector/utils.js
```diff
@@ -8,34 +8,93 @@ import {getApiEndPoint, getRequestHeaders} from './config'
 const AUTH_KEY_NAME = 'mob-auth'
 const BASKET_KEY_NAME = 'mob-basket'
 
+const setCookieValue = (keyName, value) => {
+    document.cookie = `${keyName}=${value}`
```
jeremywiebe: Cause I don't know.. but should we set a `domain` on this cookie?

PR #692, opened 2017-06-05
AMP-45: icon component 

cole-sanderson:
Added `<Icon>` component. 

 **JIRA**: https://mobify.atlassian.net/browse/AMP-45

## Changes
- change sprite prefix to `amp`
- `<Icon>` component 

## How to test-drive this PR
- Run `npm run dev`
- Preview [AMP Merlin's Potions](http://localhost:3000/)


amp/app/components/icon/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+                <use role="img" xlinkHref={`#${prefix}-${name}`} />
+            </svg>
+        )
+    }
+}
+
+Icon.defaultProps = {
+    name: '',
+    title: '',
+    prefix: 'amp',
```
olibrook: Doesn't this need to match the prefix in the SVG icons? You asked that we continue using "pw" the other day...

amp/package.json
```diff
@@ -61,7 +61,7 @@
   "scripts": {
     "add:component": "amp-sdk-add-component",
     "add:container": "amp-sdk-add-container",
-    "build-sprites": "npm run clean:svgs && svg-sprite --symbol --symbol-dest='.' --symbol-sprite='sprite.svg' --shape-id-generator='pw-%s' --dest=./app/static/svg/sprite-dist ../web/app/static/svg/sprite-source/*.svg",
+    "build-sprites": "npm run clean:svgs && svg-sprite --symbol --symbol-dest='.' --symbol-sprite='sprite.svg' --shape-id-generator='amp-%s' --dest=./app/static/svg/sprite-dist ../web/app/static/svg/sprite-source/*.svg",
```
olibrook: ah.

amp/app/components/icon/index.jsx
olibrook: I see you changed it in the build script. Fair enough.

olibrook:
Looks fine. Did anything actually change from the PWA? I think if nothing at all changes, we could just do:

```
import Icon from 'pwa'

export default Icon
```

That would just re-export the same thing, but put it in a namespace - `amp` - that makes it clear to a developer what is safe to use on AMP or not.

amp/app/components/icon/index.jsx
cole-sanderson: Yeah, I had second thoughts about it. Since `amp` have their own `build-sprites` tool and I think it makes more sense to use `amp` prefix. But later will need to deal with it when we decide what to do with "common" work.

olibrook:
👍 With the above comment.

cole-sanderson:
@olibrook there is one small change from PWA's Icon component is that I removed 

```
shouldComponentUpdate(nextProps, nextState) {
        return shallowCompare(this, nextProps, nextState)
    }
```

olibrook:
@cole That method that you removed would be okay server-side - it would just never get called. I don't mind which approach we take (re-export vs. copy), I'm just letting you know that it's possible.

amp/app/components/icon/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+            prefix,
+            name,
+            title,
+            size,
+            className,
+            style
+        } = this.props
+
+        const a11y = (title.length) ? {role: 'img'} : {'aria-hidden': 'true'}
+        const sizeClass = `amp--${size} amp--${size}`
```
cole-sanderson: good catch 👀

PR #693, opened 2017-06-05
Add Stub Connector

cdegit:
This PR adds an extremely basic connector which just stubs out all of the expected commands. This is to give partners who need to create a brand new custom connector a place to start from.

## Changes
- Create stub connector
- Create stubs for all commands the integration manager expects
- Include some utility commands inside the app commands to make things like page fetching, offline mode easier

## How to test-drive this PR
- In main.jsx, register the stub connector 
```
import {Connector} from './integration-manager/_stub-connector'
```
- Run `npm run dev`

### DIFF
--- a/web/app/integration-manager/_stub-connector/actions.js
+++ b/web/app/integration-manager/_stub-connector/actions.js
@@ -9,3 +9,5 @@ import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
 // by the app should go into ./results.js
 
 export const exampleAction = createAction('Some project-specific action')
+
+export const receiveFormInfo = createAction('Receive Form Info')
--- a/web/app/integration-manager/_stub-connector/actions.js
+++ b/web/app/integration-manager/_stub-connector/actions.js
@@ -1,0 +1,11 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
+
+// Add other actions here that are specific to this connector.
+// Actions that are returned out of the connector and reduced
+// by the app should go into ./results.js
+
+export const exampleAction = createAction('Some project-specific action')
--- a/web/app/integration-manager/_stub-connector/actions.js
+++ b/web/app/integration-manager/_stub-connector/actions.js
@@ -8,6 +8,6 @@ import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
 // Actions that are returned out of the connector and reduced
 // by the app should go into ./results.js
 
-export const exampleAction = createAction('Some project-specific action')
+export const exampleAction = createAction('Some connector-specific action')
 
 export const receiveFormInfo = createAction('Receive Form Info')
--- a/web/app/integration-manager/_stub-connector/app/commands.js
+++ b/web/app/integration-manager/_stub-connector/app/commands.js
@@ -2,7 +2,8 @@ import {jqueryResponse} from 'progressive-web-sdk/dist/jquery-response'
 import {makeRequest} from 'progressive-web-sdk/dist/utils/fetch-utils'
 
 import {
-    setPageFetchError
+    setPageFetchError,
+    receiveNavigationData
 } from '../../results'
 
 /**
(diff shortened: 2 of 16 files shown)


web/app/integration-manager/_stub-connector/app/commands.js
```diff
@@ -0,0 +1,48 @@
...
+        return Promise.resolve(capturedDocResponse)
+    })
+}
+
+let isInitialEntryToSite = true
+
+/**
+ * This function is used to fetch data from a desktop page
+ */
+export const fetchPageData = (url) => (dispatch) => {
```
jnurse: The fetchPageData and requestCapturedDoc helper functions are only going to be needed if you're doing HTML scrapping. Maybe a comment here explaining that (or something in the readme) would be helpful? What do you think?

cdegit: Yup, I think that makes sense!

web/app/integration-manager/_stub-connector/app/commands.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    return new Promise((resolve) => {
+        // For more information on the shape of the expected data,
+        // see https://docs.mobify.com/progressive-web/latest/components/#!/Nav
+        dispatch(receiveNavigationData(exampleNavigationData))
+        dispatch(getCart())
+        dispatch(setCartURL('/checkout/cart/'))
+        resolve()
+    })
```
DanielBrookRoberge: This should be equivalent to 
```javascript
dispatch(receiveNavigationData(exampleNavigationData))
dispatch(getCart())
dispatch(setCartURL('/checkout/cart/'))
return Promise.resolve()
```

AFAICT using the Promise constructor when not wrapping a callback API is something of an antipattern.

cdegit: I was originally wrapping them all like this due to the home page carousel needing to be inside a set timeout... but there's not much point here, and it might encourage people to use that antipattern, so I'll change it.

web/app/integration-manager/_stub-connector/home/commands.js
```diff
@@ -0,0 +1,23 @@
...
+            alt: 'Placeholder Image'
+        }]
+    }
+
+    return new Promise((resolve) => {
+        setTimeout(() => {
+            dispatch(receiveHomeData(exampleData))
+            resolve()
+        })
+    })
```
DanielBrookRoberge: Another way to write this would be
```javascript
return Promise.resolve()
    .then(() => dispatch(receiveHomeData(exampleData)))
```
Not sure if that would be clearer?

DanielBrookRoberge:
I think it might be useful for all the stubbed out commands to `console.log` that they've been called. Otherwise they kind of disappear since thunks don't appear in the dev tools.

web/app/integration-manager/_stub-connector/app/commands.js
```diff
@@ -0,0 +1,85 @@
...
+ * You will only need to use this function if you're using HTML scraping as your back end
+ */
+export const fetchPageData = (url) => (dispatch) => {
+    const request = isInitialEntryToSite ? requestCapturedDoc() : makeRequest(url)
+    isInitialEntryToSite = false
+
+    return request
+        .then(jqueryResponse)
+        .then((res) => {
+            const [$, $response] = res
```
DanielBrookRoberge: This probably needs a lint override, since we're not actually using these.

web/app/main.jsx
```diff
@@ -26,7 +26,7 @@ import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
 
 
 // DO NOT USE! Merlins Connector is an example connector that is for demo only
-import {Connector} from './integration-manager/_merlins-connector'
+import {Connector} from './integration-manager/_stub-connector'
```
DanielBrookRoberge: I guess we want to switch this back to Merlin's before merging

web/app/integration-manager/_stub-connector/actions.js
```diff
@@ -0,0 +1,13 @@
...
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
+
+// Add other actions here that are specific to this connector.
+// Actions that are returned out of the connector and reduced
+// by the app should go into ./results.js
+
+export const exampleAction = createAction('Some project-specific action')
```
jeremywiebe: Let's change this slightly.  This is a connector-specific action.  I guess for all purposes that means it's project-specific, but the actions exported here are for managing connector-private data.

web/app/integration-manager/_stub-connector/selectors.js
```diff
@@ -0,0 +1,5 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+export const getIntegrationManager = ({integrationManager}) => integrationManager
```
jeremywiebe: Let's also add a `getFormInfoByKey` here so that they don't have to figure out that piece the first time they need to submit a form back.

PR #694, opened 2017-06-06
WEBDATA-101 Retain email address in checkout flow

kevinxh:
When checking out on Merlin's the confirmation email is not filled in again when going back to the Shipping step. Also, the country field is being reset to `US` during the initialization of the shipping page.

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-101

## Issue

6oyfi4myn2

## Changes
- Update checkout redux form initial value, username(email) is included now
- Update the email field onBlur handler, to avoid sending request with empty email provided
- Added `isInitialized` flag for shipping page

## How to test-drive this PR
- (necessary config changes)
- Run `npm run dev`
- Add your favourite potion to the cart
- Fill out shipping information
- Continue to payment
- Go back to shipping page and verify the form fields


web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -106,7 +106,7 @@ const processCheckoutData = ($response) => (dispatch) => {
           .getIn(['children', 'shipping-address-fieldset', 'children'])
 
     dispatch(receiveCheckoutLocations(parseLocations(magentoFieldData)))
-    dispatch(receiveShippingAddress(parseShippingInitialValues(magentoFieldData)))
+    // dispatch(receiveShippingAddress(parseShippingInitialValues(magentoFieldData)))
```
kevinxh: @jeremywiebe This line is commented out because the parser resets `countryId` to `US` during page initialization, and does nothing else... I'm not clear about how does the magento utility work.. But it looks like the issue is indeed solved by taking this line out.

Is there any other side effects that I am not aware of?

web/app/containers/checkout-shipping/actions.js
```diff
@@ -128,13 +128,15 @@ export const isEmailAvailable = () => (dispatch, getState) => {
...
 
     if (customCommands.isEmailAvailable) {
-        return dispatch(customCommands.isEmailAvailable(formValues.username))
-        .then((emailAvailable) => {
-            if (emailAvailable) {
-                return dispatch(onShippingEmailAvailable())
-            }
-            return dispatch(onShippingEmailRecognized())
-        })
+        if (formValues.username) {
```
jeremywiebe: We could tighten this up a bit by doing both `if` checks in the same statement. 
```js
if (customCommands.isEmailAvailable && formValues.username) { ... } 
```

web/app/integration-manager/_merlins-connector/checkout/commands.js
jeremywiebe: I think it's a bit more complicated than that.  We're dispatching this `processCheckoutData()` method in both `initCheckoutShippingPage` and `initCheckoutPaymentPage`.  It seems like we'd only want to do this in `initCheckoutShippingPage`.  I think you'll need to split out the `receiveShippingAddress` call from this method and call it only in `initCheckoutShippingPage`.

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -109,14 +109,22 @@ const processCheckoutData = ($response) => (dispatch) => {
     dispatch(receiveShippingAddress(parseShippingInitialValues(magentoFieldData)))
 }
 
+const processPaymentData = ($response) => (dispatch) => {
+    dispatch(receiveEntityID(parseCheckoutEntityID($response)))
+    const magentoFieldData = extractMagentoShippingStepData($response)
+          .getIn(['children', 'shipping-address-fieldset', 'children'])
+
+    dispatch(receiveCheckoutLocations(parseLocations(magentoFieldData)))
```
kevinxh: @jeremywiebe Should `receiveCheckoutLocations` be taken out from `processPaymentData ` as well? I assume it is used to populate the country and region lists which is done in the shipping page..

jeremywiebe: Yes, definitely this needs to be here.  We have an address on both pages.

DanielBrookRoberge: Probably nothing will change, but it's definitely safer to do this here too.

jeremywiebe:
When I tested the email works, but the country still reverts to US which then breaks the State/Province field also. 

n3yvyljvmc

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -116,7 +130,7 @@ const shippingDataSelector = createPropsSelector({
 
 export const initCheckoutShippingPage = (url) => (dispatch, getState) => {
     return dispatch(fetchPageData(url))
-        .then(([$, $response]) => dispatch(processCheckoutData($response)))  // eslint-disable-line no-unused-vars
+        .then(([$, $response]) => dispatch(processShippingData($response, getState)))  // eslint-disable-line no-unused-vars
```
jeremywiebe: We shouldn't need to add `getState` as a parameter to `processShippingData`.  `processShippingData` is already a [redux-thunk](https://docs.mobify.com/progressive-web/latest/reference/function-glossary/#thunks).   The `getState` will be provided to `processShippingData` automatically because you're dispatching it.

web/app/integration-manager/_merlins-connector/checkout/commands.js
```diff
@@ -99,14 +99,28 @@ export const fetchSavedShippingAddresses = (selectedSavedAddressId) => {
     }
 }
 
+const isInitializedSelector = createPropsSelector({
+    isInitialized: shippingSelectors.getIsInitialized
+})
```
DanielBrookRoberge: This isn't necessary; you can just do `const isInitialized = shippingSelectors.getIsInitialized(getState())` below.

kevinxh:
@DanielBrookRoberge I agree with you, it resets the flag in `submitPayment` now. However we are facing another issue: if the user does make another order, then all fields are filled with the previous information except the `Country` field, which reverts back to `US`. I wonder if there is a more robust solution to this.

web/app/integration-manager/_merlins-connector/checkout/commands.js
kevinxh: @jeremywiebe I wasn't 100% clear on the thunk actions, now I understand! Thanks for the explanation. 👍

web/app/integration-manager/_merlins-connector/checkout/commands.js
kevinxh: @jeremywiebe Could you review the new changes? Thanks!

kevinxh:
@jeremywiebe Logged it: https://mobify.atlassian.net/browse/WEBDATA-135

PR #698, opened 2017-06-06
Fix worker loading

fractaltheory:
Previewing a bundle from the CDN will attempt to load the main worker module from `https://localhost:8443/worker.js` - which will not work.

## Changes
- Instead of `IS_PREVIEW` boolean testing for a `mobify-path=true` cookie, look at the build origin of the `loader.js` script and evaluate to `false` if `cdn.mobify.com` is found
- Change `IS_PREVIEW` name to `IS_LOCAL_PREVIEW` to be a little more descriptive
- Add to Webpack config. to add DefinePlugin `DEBUG` value to loader file (both dev and production configs)
- Use `DEBUG` value from Webpack build process to pass to Messaging Client init instead, which is more accurate
- Also add a `resolve` handler for `react` so that `npm link progressive-web-sdk` can be used without error

### CDN
- Run `npm run upload`
- Preview the preview URL it returns for the bundle that was created
- Ensure that the main worker is loaded from the CDN and not from `localhost`
### Local
- Run `npm run dev`
- Preview: https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=0
- Ensure the main worker is loaded from your local development server

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -1,4 +1,4 @@
-/* global NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, MESSAGING_ENABLED */
+/* global NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, MESSAGING_ENABLED, DEBUG */
 import {getAssetUrl, getBuildOrigin, loadAsset, initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
 import {isSamsungBrowser, isFirefoxBrowser} from 'progressive-web-sdk/dist/utils/utils'
 import {displayPreloader} from 'progressive-web-sdk/dist/preloader'
@@ -153,7 +153,7 @@ const attemptToInitializeApp = () => {
             // supported and loaded, and messaging is enabled, so we can add a
             // deferred function to load and initialize the Messaging client.
             deferredUntilLoadComplete.push(
-                () => loadAndInitMessagingClient(IS_PREVIEW, MESSAGING_SITE_ID)
+                () => loadAndInitMessagingClient(DEBUG, MESSAGING_SITE_ID)
             )
         }
     }
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -1,5 +1,5 @@
 /* global NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, MESSAGING_ENABLED */
-import {getAssetUrl, loadAsset, initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
+import {getAssetUrl, getBuildOrigin, loadAsset, initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
 import {isSamsungBrowser, isFirefoxBrowser} from 'progressive-web-sdk/dist/utils/utils'
 import {displayPreloader} from 'progressive-web-sdk/dist/preloader'
 import cacheHashManifest from '../tmp/loader-cache-hash-manifest.json'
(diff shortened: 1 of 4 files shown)


web/app/loader.js
```diff
@@ -32,14 +32,20 @@ const attemptToInitializeApp = () => {
...
     const ASTRO_CLIENT_CDN = `//assets.mobify.com/astro/astro-client-${ASTRO_VERSION}.min.js`
-    const SW_LOADER_PATH = `/service-worker-loader.js?preview=${IS_PREVIEW}&b=${cacheHashManifest.buildDate}`
+
+    /**
+     * This needs to be based on whether this is a CDN environment, rather than
+     * a preview environment. `web/service-worker-loader.js` will use this value
+     * to determine whether it should load from a local development server, or
+     * from the CDN.
+     */
+    const IS_LOCAL_PREVIEW = getBuildOrigin().indexOf('cdn.mobify.com') === -1
```
jeremywiebe: I'm not sure this is 100% what we want. This means we **can't** preview non-published bundles that are already pushed up to cloud, right?

DanielBrookRoberge: I definitely remember setting it up with this condition back when we first got the SW up. Don't know when it changed but good to put it back.

web/webpack/base.main.js
```diff
@@ -41,6 +41,7 @@ const config = {
             'lodash.keysin': path.resolve(process.cwd(), 'node_modules', 'lodash', 'keysIn'),
             'lodash.mapvalues': path.resolve(process.cwd(), 'node_modules', 'lodash', 'mapValues'),
             'lodash.throttle': path.resolve(process.cwd(), 'node_modules', 'lodash', 'throttle'),
+            react: path.resolve(process.cwd(), 'node_modules', 'react')
```
DanielBrookRoberge: :boom: So glad you fixed this.

web/webpack/dev.js
```diff
@@ -31,6 +31,12 @@ mainConfig.plugins = mainConfig.plugins.concat([
     })
 ])
 
+loaderConfig.plugins = loaderConfig.plugins.concat([
+    new webpack.DefinePlugin({
+        DEBUG: true
```
jeremywiebe: If we don't have this variable in use anywhere I'd like to see these webpack "DefinePlugin" variables prefixed with `WEBPACK_` going forward.  It's getting harder and harder to tell where they come from when you see them in the code.

jeremywiebe: Ah, I see this isn't the first time we've defined this... Ok, let's not change it for now.

web/app/loader.js
fractaltheory: As far as I know we've never had the capability to test a main worker file from a pushed bundle on the CDN. That said, I don't think that the behavior should be that a bundle pushed to the CDN has a dependency on a local development server running - which is what this fixes.

web/webpack/dev.js
fractaltheory: This is the first use of the `DEBUG` variable via DefinePlugin in the loader file, however it was already being used in both the main and worker files. 

If you're cool with introducing inconsistency between `loader.js` and the rest of the app, that's fine - but ultimately that seems like something ripe for an all-encompassing refactor, not here.

web/app/loader.js
DanielBrookRoberge: Yeah, the setup has always been that all bundles pushed to Cloud get the production service worker. Fixing that has been on our long term TODO list, but nobody's found time to implement it (it has to go in the worker loader and be careful about security)

jeremywiebe: Ok, fair enough.  :)

PR #700, opened 2017-06-06
[WEBDATA-114] Merlins checkout address failure

jnurse:
This PR adds an error message when users enter an incomplete full name (ie. first name only). It also adds validation to the billing address form (including the incomplete full name error), and updates how we split full names. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-114
 **Linked PRs**: n/a

## Changes
- Added full name validation to shipping & billing forms
- Updated split names function to better support double barrelled last names (split Joe De Magio into firstname: Joe lastname: De Magio)

## How to test-drive this PR
- Run `npm run dev`
- follow the steps in the ticket (be sure to enter only one name "Mobify" on the shipping form)

### DIFF
--- a/web/app/containers/checkout-payment/partials/checkout-payment-form.jsx
+++ b/web/app/containers/checkout-payment/partials/checkout-payment-form.jsx
@@ -22,6 +22,34 @@ import CreditCardForm from './credit-card-form'
 import BillingAddressForm from './billing-address-form'
 import OrderSummary from './order-summary'
 
+const REQUIRED_TEXT = 'Required'
+
+const validate = (values) => {
+    const errors = {}
+    const requiredFieldNames = [
+        'name',
+        'addressLine1',
+        'city',
+        'countryId',
+        'regionId',
+        'postcode',
+        'telephone'
+    ]
+
+    if (values.name && !/\w+ \w+/.test(values.name)) {
+        errors.name = 'Please enter a first and last name'
+    }
+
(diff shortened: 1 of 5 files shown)


web/app/containers/checkout-payment/partials/checkout-payment-form.jsx
```diff
@@ -22,6 +22,34 @@ import CreditCardForm from './credit-card-form'
...
+        'name',
+        'addressLine1',
+        'city',
+        'countryId',
+        'regionId',
+        'postcode',
+        'telephone'
+    ]
+
+    if (values.name && !/\w+ \w+/.test(values.name)) {
```
jeremywiebe: Non-blocking: Maybe not realistic, but this regex might be a bit safer if we use `\s+` just so we're super flexible no matter how they separate their first and last name.

jeremywiebe: Also, given that we do this twice in this PR... is that a signal for a utility method?

web/app/utils/utils.js
```diff
@@ -39,9 +39,10 @@ export const getCookieValue = (cookieName) => {
 
 export const splitFullName = (fullname) => {
     const names = fullname.split(' ')
+
```
jeremywiebe: Let's `trim()` all the `names` here so we can't get this:

image

PR #701, opened 2017-06-06
[WEBDATA-120] Fix password confirmation

DanielBrookRoberge:
This PR fixes the bug preventing registration of new user accounts. It also reads the Magento messages cookie to get the correct error message for the register and login forms.

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-120

## Changes
- No longer pass the password confirmation value into the command
- Duplicate the password in the Merlin's connector for registration
- Check that the passwords match in the confirmation page registration form
- Read the error message out of the `mage-messages` cookie rather than using a hard-coded value

## How to test-drive this PR
- Preview Merlin's
- Attempt to register with a new email address (e.g. `yourname+webdata120@mobify.com`) 
- See that it works

### DIFF
--- a/web/app/containers/checkout-confirmation/actions.js
+++ b/web/app/containers/checkout-confirmation/actions.js
@@ -5,6 +5,7 @@
 import {CHECKOUT_CONFIRMATION_MODAL, CHECKOUT_CONFIRMATION_REGISTRATION_FAILED} from './constants'
 import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
 import {createPropsSelector} from 'reselect-immutable-helpers'
+import {SubmissionError} from 'redux-form'
 import {addNotification, removeAllNotifications} from 'progressive-web-sdk/dist/store/notifications/actions'
 import {openModal} from 'progressive-web-sdk/dist/store/modals/actions'
 import * as shippingSelectors from '../../store/checkout/shipping/selectors'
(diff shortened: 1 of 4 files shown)


web/app/containers/login/actions.js
```diff
@@ -120,10 +120,9 @@ export const submitRegisterForm = (formValues) => (dispatch) => {
         firstname,
         lastname,
         email,
-        password,
-        password_confirmation
```
jeremywiebe: Should we be doing the same comparison between `password` and `password_confirmation` here as we do in the checkout register step?

DanielBrookRoberge: The `validateRegisterForm` function already does this, so I just left it alone.

web/app/integration-manager/_merlins-connector/account/commands.js
```diff
@@ -30,25 +31,30 @@ export const initRegisterPage = (url) => (dispatch) => {
...
-                throw new SubmissionError(error)
+                const messages = JSON.parse(decodeURIComponent(getCookieValue(MAGENTO_MESSAGE_COOKIE)))
+
+                throw new SubmissionError({
+                    _error: (messages[0] || defaultErrorMessage).text.replace(/\+/g, ' ')
+                })
             }
             return '/customer/account'
         })
-        .catch((error) => {
```
jeremywiebe: Nice!

web/app/containers/checkout-confirmation/actions.js
```diff
@@ -40,7 +41,15 @@ export const submitRegisterForm = () => {
             billingAddressData
         } = registrationFormSelector(getState())
 
-        return dispatch(registerUser(firstname, lastname, email, password, password_confirmation))
+        /* eslint-disable camelcase */
+        if (password !== password_confirmation) {
```
jnurse: This shouldn't be necessary if we've got a validate function checking the same for the form.

PR #704, opened 2017-06-06
[WEBDATA-112] Check for empty forminfo value in addToCart

Cdok:
# Check for empty forminfo value in addToCart

`formInfo` would return `undefined` in some rare intermittent cases during checkout. When that occurred, the `Add to Cart` button would remain disabled.

This PR fixes the add-to-cart remaining disabled, thus unblocking checkout.

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-112

## Changes
- Add a check for the intermittent empty form info to unblock addToCart

## How to test-drive this PR
- Use merlins connector
- Run `npm run dev`

## Steps to reliably reproduce:
1. Browse to https://www.merlinspotions.com/books/dragon-breeding-for-pleasure-and-profit.html
2. Click Add to Cart (should work)
3. Browse to https://www.merlinspotions.com/potions/unicorn-blood.html
4. Add to cart
5. Go to checkout
6. Remove Unicorn Blood
7. Browse to https://www.merlinspotions.com/books/dragon-breeding-for-pleasure-and-profit.html
8. Add to cart

Ensure that it all works, without Add To Cart button staying permanently disabled

### DIFF
--- a/web/app/integration-manager/_merlins-connector/cart/commands.js
+++ b/web/app/integration-manager/_merlins-connector/cart/commands.js
@@ -61,7 +61,7 @@ export const addToCart = (productId, quantity) => (dispatch, getState) => {
     const hiddenInputs = formInfo.get('hiddenInputs')
 
     if (hiddenInputs === undefined) {
-        Promise.reject('Add to cart failed, form info missing')
+        return Promise.reject('Add to cart failed, form info missing')
     }
 
     const formValues = {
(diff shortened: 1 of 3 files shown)


web/app/integration-manager/_merlins-connector/cart/commands.js
```diff
@@ -56,6 +56,11 @@ export const getCart = () => (dispatch) => {
 export const addToCart = (productId, quantity) => (dispatch, getState) => {
     const product = getProductById(productId)(getState())
     const formInfo = getState().integrationManager.get(urlToPathKey(product.get('href')))
```
jeremywiebe: Instead of doing this like this, let's create a selector like `getFormInfoByKey` so that we don't have to dig around inside the `integrationManager` part of the tree directly.

PR #707, opened 2017-06-07
Fix amp validation errors when using the Sheet component

olibrook:
## Changes
```
    Fix AMP validation errors which are breaking tests.

    - The amp-sidebar must be a direct child of the body tag for it to
      be valid. Render the <body> in react to fix.
    - Provide a Jest mock to stub out webpack imports of SVG assets which
      are not needed during testing.
```
## How to test-drive this PR
- Run `npm run test` locally - not captured by CircleCI at the moment, I guess.

### DIFF
--- a/amp/app/__mocks__/svgMock.js
+++ b/amp/app/__mocks__/svgMock.js
@@ -1,0 +1,1 @@
+module.exports = ''
--- a/amp/app/main.js
+++ b/amp/app/main.js
@@ -61,6 +61,7 @@ const render = (req, res, store, component, css) => {
         <ampSDK.AmpContext declareDependency={scripts.add}>
             <Provider store={store}>
                 <App>
+                    <h1>foo</h1>
                     <Analytics templateName={component.templateName} projectSlug={ampPackageJson.cloudSlug} gaAccount={ampPackageJson.gaAccount} />
                     {React.createElement(component, {}, null)}
                 </App>
--- a/amp/app/main.js
+++ b/amp/app/main.js
@@ -61,7 +61,6 @@ const render = (req, res, store, component, css) => {
         <ampSDK.AmpContext declareDependency={scripts.add}>
             <Provider store={store}>
                 <App>
-                    <h1>foo</h1>
                     <Analytics templateName={component.templateName} projectSlug={ampPackageJson.cloudSlug} gaAccount={ampPackageJson.gaAccount} />
                     {React.createElement(component, {}, null)}
                 </App>
--- a/amp/app/main.test.js
+++ b/amp/app/main.test.js
@@ -13,7 +13,7 @@ describe('Renders valid AMP', () => {
     })
 
     const validateAmp = (html) => {
-        expect(validator.validateString(html).status).toEqual('PASS')
+        expect(validator.validateString(html).errors).toEqual([])
     }
 
     const handle = (req) => {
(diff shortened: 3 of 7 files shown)


amp/app/main.js
```diff
@@ -61,6 +61,7 @@ const render = (req, res, store, component, css) => {
         <ampSDK.AmpContext declareDependency={scripts.add}>
             <Provider store={store}>
                 <App>
+                    <h1>foo</h1>
```
kschmidtdev: 🔫

amp/app/main.test.js
```diff
@@ -13,7 +13,7 @@ describe('Renders valid AMP', () => {
     })
 
     const validateAmp = (html) => {
-        expect(validator.validateString(html).status).toEqual('PASS')
+        expect(validator.validateString(html).errors).toEqual([])
```
kschmidtdev: Does this fix something you ran into?

amp/app/templates/amp-page.js
```diff
@@ -22,11 +22,7 @@ const ampPage = ({title, canonicalURL, body, css, ampScriptIncludes}) => (
                 ${css}
             </style>
         </head>
-        <body>
-            <div id="root">
-                ${body}
-            </div>
-        </body>
+        ${body}
```
kschmidtdev: I think this might break a bunch of styling-related assumptions...we should coordinate with Cole about the severity of this. There are more components that require `<body>` as the direct parent so we'll need a general solution.

amp/app/__mocks__/svgMock.js
```diff
@@ -0,0 +1 @@
+module.exports = ''
```
kschmidtdev: Nice 😆

amp/app/main.js
olibrook: Whoops. Fixed.

amp/app/main.test.js
olibrook: Nope - it just means you can see the description of the error when the test fails.

amp/app/templates/amp-page.js
olibrook: @cole-sanderson You see any problems with this? I think @kschmidtdev might be talking about the `<div id="root">` that is gone, but would be easy to add to, say, the `<App/>` component.

In any event, it's going to be necessary to output the `<body>` tag in a React component if there are amp components that need to be a direct-child of the body - A React component must render one element exactly, so the correct nesting is not possible with the previous version of this template.

amp/app/__mocks__/svgMock.js
olibrook: Thanks.

amp/app/templates/amp-page.js
cole-sanderson: We will need `<div id="root">` to work with utilities styles. But if we can move to `<App/>` then it should be no problem.

olibrook: @cole Can `id="root"` go on the `<body>`? Otherwise, how do you propose fixing the issue that some amp components need to be the direct child of the body?

cole-sanderson: @olibrook Yeah, it should be fine to add `id=root` to `<body>`.

olibrook: @cole-sanderson I've put it on the body.

cole-sanderson:
Looks good 👍

PR #711, opened 2017-06-07
[Refactor] Transition from urlToPathkeys -> product IDs for Products & Category keys

Cdok:
# [Refactor] Transition from urlToPathkeys -> product IDs for Products & Category keys

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-112
**Linked PRs**: https://github.com/mobify/platform-scaffold/pull/704 and https://github.com/mobify/platform-scaffold/pull/733

## Changes
- Transition from urlToPathkeys -> product IDs across both connectors

## How to test-drive this PR
- Tests on both Connectors
- Run `npm run dev`
- Ensure things look good in the Redux store in terms of product-IDs as keys (under categories, cart, products, integrationManager -> formInfo)
- Refresh on PDP - ensure page still works landing directly on PDP
- Try PLP -> PDP flows


web/app/integration-manager/_merlins-connector/categories/parser.js
```diff
@@ -72,8 +71,7 @@ const categoryProductsParser = ($, $html) => {
 
     const products = $
```
jeremywiebe: Are these products or an array of product IDs?  If they're just IDs can we fix this variable name to make that clear?

web/app/integration-manager/_merlins-connector/products/parsers.js
```diff
@@ -95,10 +94,11 @@ export const productListParser = ($, $html) => {
     $products.each((_, product) => {
         const $product = $(product)
         const link = parseTextLink($product.find('.product-item-link'))
+        const productId = $product.find('input[name="product"]').val()
         const thumbnail = parseImage($product.find('.product-image-photo'))
         const available = $product.find('.stock.unavailable').length === 0
 
-        productMap[urlToPathKey(link.href)] = {
+        productMap[productId] = {
             id: $product.find('.price-box').length ? $product.find('.price-box').attr('data-product-id') : '',
```
jeremywiebe: Shouldn't this be the same expression as we used above to create the `productId` ?

Cdok: Ah yes, this one looks safer than grabbing the input value. I've updated it.

It seems like the `productId` is undefined from desktop whenever the item is out of stock. This `.each` might need to be reworked to also `filter` out any items without a productId, otherwise we're getting an extra skeleton product for out of stock items.

web/app/integration-manager/_merlins-connector/categories/parser.js
```diff
@@ -70,14 +69,15 @@ export const priceFilterParser = ($, $html) => {
 const categoryProductsParser = ($, $html) => {
     const $numItems = $html.find('#toolbar-amount .toolbar-number').first()
 
-    const products = $
+    const productIds = $
           .makeArray($html.find('.item.product-item'))
-          .map((product) => $(product).find('.product-item-link').attr('href'))
-          .map(urlToPathKey)
+          .map((product) => {
+              return $(product).find('.price-box').length ? $(product).find('.price-box').attr('data-product-id') : ''
```
DanielBrookRoberge: Should we filter out the entries with invalid product IDs? If we have products with no ID then their data will just step all over each other...

web/app/integration-manager/_merlins-connector/products/parsers.js
```diff
@@ -95,11 +94,12 @@ export const productListParser = ($, $html) => {
     $products.each((_, product) => {
         const $product = $(product)
         const link = parseTextLink($product.find('.product-item-link'))
+        const productId = $product.find('.price-box').length ? $product.find('.price-box').attr('data-product-id') : ''
```
DanielBrookRoberge: We'll probably need to avoid the id-less products here too.

Cdok: Good call! In the latest commit I've filtered out the products that don't contain `.price-box` at the top of this function, before we `.each` on them

web/app/integration-manager/_sfcc-connector/products/commands.js
```diff
@@ -57,3 +63,5 @@ export const getProductVariantData = (variationSelections, variants, categoryIds
 
     return Promise.resolve()
 }
+
+export const setCurrentSelectedProduct = (product) => (dispatch) => dispatch(receiveCurrentProduct(product))
```
jnurse: I don't think a command is needed here, a plain action in `products/results` should be good enough.

web/app/integration-manager/products/results.js
```diff
@@ -25,3 +25,9 @@ export const receiveCartProductData = createTypedAction(
     'Receive Cart product data',
     Products
 )
+
+export const receiveCurrentProduct = createTypedAction(
+    'Receive Product Details UI data',
+    Product,
+    ['currentProduct']
+)
```
jnurse: I'm wondering if it makes sense to store the whole product object, or just the ID for the current product. Storing the whole product object makes our selector pretty simple (just pull all the data off that one key), but it also means we're storing duplicate data. What do you think?

web/app/integration-manager/_sfcc-connector/products/commands.js
Cdok: Ah cool, I wasn't sure if `dispatch`ing directly from the results was okay 👍

web/app/integration-manager/products/results.js
Cdok: I definitely had mixed feelings on this one, but thinking about it a bit more - a little re-work to the selectors to prevent the duplicate data / potentially different sources of truth about the product makes more sense

web/app/containers/product-list/partials/product-list-contents.jsx
```diff
@@ -25,14 +25,21 @@ import ProductTile from '../../../components/product-tile'
...
 
-const ResultList = ({products}) => (
+const ResultList = ({products, setProduct}) => (
     <List className="c--borderless">
-        {products.map((product, idx) => <ProductTile key={product ? product.id : idx} {...product} />)}
+        {products.map((product, idx) => (
+            <ProductTile
+                onClick={product ? () => setProduct(product.id) : null}
+                key={product ? product.id : idx}
+                {...product}
```
DanielBrookRoberge: Are we going to have falsy products here? I think the spread operator would fail if that were the case. 

Maybe instead, we do `key={product.id || idx}`?

web/app/integration-manager/_merlins-connector/cart/commands.js
```diff
@@ -56,7 +56,7 @@ export const getCart = () => (dispatch) => {
 export const addToCart = (productId, quantity) => (dispatch, getState) => {
     const product = getProductById(productId)(getState())
 
-    const formInfo = getFormInfoByPathKey(urlToPathKey(product.get('href')))(getState())
+    const formInfo = getFormInfoById(product.get('id'))(getState())
```
DanielBrookRoberge: You should be able to just pass `productId` here.

web/app/containers/product-list/actions.js
```diff
@@ -12,3 +13,5 @@ export const changeSort = (sort) => (dispatch, getState) => {
         [getCurrentPathKey(getState())]: {sort}
     }))
 }
+
+export const setCurrentProduct = (id) => (dispatch) => dispatch(setCurrentProductId(id))
```
jeremywiebe: I'm confused why we're only passing an `id` to a method called `setCurrentProduct`

web/app/store/products/selectors.js
```diff
@@ -6,12 +6,13 @@ import {createSelector} from 'reselect'
...
 
-import {getProducts} from '../selectors'
-import {getCurrentPathKey} from '../../containers/app/selectors'
+import {getProducts, getUi} from '../selectors'
 
-export const getSelectedProduct = createGetSelector(getProducts, getCurrentPathKey, Immutable.Map())
+const getApp = createSelector(getUi, ({app}) => app)
 
-export const getSelectedProductId = createGetSelector(getSelectedProduct, 'id')
+export const getSelectedProductId = createGetSelector(getApp, 'currentProduct')
```
DanielBrookRoberge: We might want to rename this and the subsequent one `getCurrentProduct` and `getCurrentProductId` to be more consistent with the terminology around here.

web/app/containers/product-list/partials/product-list-contents.jsx
```diff
@@ -166,6 +175,7 @@ const mapDispatchToProps = {
     clearFilters: () => changeFilterTo(null),
     openModal: () => openModal(PRODUCT_LIST_FILTER_MODAL),
     sortChange: changeSort,
+    setProduct: setCurrentProduct
```
jeremywiebe: I'd recommend that we keep `Current` in this method name.  It becomes less clear why we're attaching it to the `ResultList` above if it's just `setProduct`...

web/app/integration-manager/_merlins-connector/cart/commands.js
jeremywiebe: Now that we're just using an `id`....do you think we'll have collisions between different types of products/categories/etc?  Before with using the URL, we couldn't get that.  Wondering if we need `getFormInfoByProductId`? (and other ones as needed).

web/app/containers/product-list/partials/product-list-contents.jsx
Cdok: When the skeletons load, `products` is an array of 5 `undefined`s. Changing to `key={product.id || idx}` ends up throwing `Cannot read property 'id' of undefined` onload of the PLPs. 

It doesn't seem like the spread operator is failing when passed `undefined`

web/app/store/products/selectors.js
```diff
@@ -6,12 +6,13 @@ import {createSelector} from 'reselect'
 import Immutable from 'immutable'
 import {createGetSelector} from 'reselect-immutable-helpers'
 
-import {getProducts} from '../selectors'
-import {getCurrentPathKey} from '../../containers/app/selectors'
+import {getProducts, getUi} from '../selectors'
 
-export const getSelectedProduct = createGetSelector(getProducts, getCurrentPathKey, Immutable.Map())
+const getApp = createSelector(getUi, ({app}) => app)
```
jeremywiebe: I'm not sure this is the right place for this. I _think_ we could create an `/store/ui` folder for this to go into. That keeps the redux tree in sync with the directory structure.

web/app/containers/product-list/partials/product-list-contents.jsx
DanielBrookRoberge: I think I confused this with one of the other placeholder locations. If things are working, I'm not worried.

web/app/integration-manager/_merlins-connector/cart/commands.js
Cdok: From what I've seen on magento & SFCC it seems like ID is always going to be unique regardless of category, so the collisions shouldn't be an issue.

Keying off of the productId with `getFormInfoByProductId` should be safe, if we run into an instance where there are collisions we can discuss changing the keys under `formInfo`

web/app/store/products/selectors.js
Cdok: It looks like I missed it, but a `getApp` selector already existed in `containers/app/selectors'`, so I've updated it to import that one instead.

web/app/containers/product-list/actions.js
```diff
@@ -12,3 +13,5 @@ export const changeSort = (sort) => (dispatch, getState) => {
         [getCurrentPathKey(getState())]: {sort}
     }))
 }
+
+export const setCurrentProductId = (id) => (dispatch) => dispatch(receiveCurrentProductId(id))
```
DanielBrookRoberge: Do we need this proxy function? It doesn't really add anything to the result.

web/app/integration-manager/_merlins-connector/cart/commands.js
```diff
@@ -54,10 +53,7 @@ export const getCart = () => (dispatch) => {
 }
 
 export const addToCart = (productId, quantity) => (dispatch, getState) => {
-    const product = getProductById(productId)(getState())
-
-    const formInfo = getFormInfoByPathKey(urlToPathKey(product.get('href')))(getState())
-
+    const formInfo = getFormInfoByProductId(productId)(getState())
```
DanielBrookRoberge: Nice!

web/app/store/products/selectors.js
```diff
@@ -7,35 +7,32 @@ import Immutable from 'immutable'
...
-export const getProductVariants = createGetSelector(getSelectedProduct, 'variants')
-export const getProductInitialValues = createGetSelector(getSelectedProduct, 'initialValues')
+export const getProductVariants = createGetSelector(getCurrentProduct, 'variants')
+export const getProductInitialValues = createGetSelector(getCurrentProduct, 'initialValues')
 
 // NOTE: These get-something-ByPathKey selectors should only be used within actions/commands
 // Using them within a component will break the performance optimizations selectors normally give us
 export const getProductDetailsByPathKey = (pathKey) => createGetSelector(getProducts, pathKey, Immutable.Map())
 export const getProductThumbnailByPathKey = (pathKey) => createGetSelector(getProductDetailsByPathKey(pathKey), 'thumbnail', Immutable.Map())
 export const getProductThumbnailSrcByPathKey = (pathKey) => createGetSelector(getProductThumbnailByPathKey(pathKey), 'img')
```
DanielBrookRoberge: These selectors should also be changed to talk about IDs rather than PathKeys, or removed if they're unused. (In particular, `getProductDetailsByPathKey` appears to be the same as `getProductById`)

web/app/containers/product-list/actions.js
Cdok: I don't think it's necessary, but it just felt a bit weird dispatching an IM result directly from mapDispatchToProps.

Is it okay to import a result directly into a container?

I had this implemented as a command earlier, but even that was essentially just another proxy layer

web/app/store/products/selectors.js
Cdok: Ah good catch!

web/app/containers/product-list/actions.js
DanielBrookRoberge: Let's leave it then, though it can just become `(id) => receiveCurrentProductId(id)`

web/app/store/products/selectors.js
```diff
@@ -7,35 +7,28 @@ import Immutable from 'immutable'
...
     getProductVariationCategories,
     (categories) => categories.map((category) => category.get('id'))
 )
-export const getProductVariants = createGetSelector(getSelectedProduct, 'variants')
-export const getProductInitialValues = createGetSelector(getSelectedProduct, 'initialValues')
+export const getProductVariants = createGetSelector(getCurrentProduct, 'variants')
+export const getProductInitialValues = createGetSelector(getCurrentProduct, 'initialValues')
 
-// NOTE: These get-something-ByPathKey selectors should only be used within actions/commands
-// Using them within a component will break the performance optimizations selectors normally give us
```
DanielBrookRoberge: This comment is still relevant about these selectors, so we should keep it.

jnurse:
I think we're missing setting the current ID when you click the edit button in the cart. If you navigate to a PDP and add the item to your cart, then navigate to another PDP, then view your cart and edit the first item you'll see the second PDP initially until we finish parsing the new page.

Cdok:
@jnurse good catch 👀  edit cart button is now dispatching the updated product ID

PR #712, opened 2017-06-07
Handle empty Product Listing Pages from SFCC

kevinxh:
Currently the SFCC integration does not properly render an empty PLP when there are 0 products.

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-107

## Changes
- Add response checks for empty product returned

## How to test-drive this PR
- In `/web/app/main.jsx`, import SFCC connector and comment out merlins connector
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1) **Note: use SFCC preview link**
- Navigate to Gift Certificates page, which should show no results instead of skeleton that go on forever.

### DIFF
--- a/web/app/integration-manager/_sfcc-connector/categories/commands.js
+++ b/web/app/integration-manager/_sfcc-connector/categories/commands.js
@@ -48,8 +48,13 @@ export const initProductListPage = (url) => (dispatch) => {
         .then(() => makeApiRequest(makeCategorySearchURL(categoryID), {method: 'GET'}))
         .then((response) => response.json())
         .then(({hits, total}) => {
-            const productListData = parseProductListData(hits)
-            const products = Object.keys(productListData)
+            let productListData = []
+            let products = []
+
+            if (total > 0) {
+                productListData = parseProductListData(hits)
+                products = Object.keys(productListData)
+            }
 
             dispatch(receiveProductListProductData(productListData))
             dispatch(receiveCategoryContents(urlToPathKey(url), {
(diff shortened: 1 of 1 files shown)


web/app/integration-manager/_sfcc-connector/categories/commands.js
```diff
@@ -48,8 +48,13 @@ export const initProductListPage = (url) => (dispatch) => {
...
         .then(({hits, total}) => {
-            const productListData = parseProductListData(hits)
-            const products = Object.keys(productListData)
+            let productListData = []
+            let products = []
+
+            if (total > 0) {
+                productListData = parseProductListData(hits)
+                products = Object.keys(productListData)
+            }
```
DanielBrookRoberge: Since we don't need to dispatch `receiveProductListProductData` if there are no products, we could just do:
```javascript
if (total === 0) {
    dispatch(receiveCategoryContents(urlToPathKey(url), {
        products: [],
        itemCount: total
    })
    return
}
```

kevinxh: @DanielBrookRoberge I was worried that if I don't dispatch `receiveProductListProductData`, the product list component will no re-render. But in fact I tested it and your code is perfectly working. Thanks! 👍

PR #716, opened 2017-06-07
Add ESlint rule to uncomment Merlin's Connector 🎩

Cdok:
# Adds ESlint rule to uncomment Merlin's Connector 🎩

This PR adds a linting error if any connector is left uncommented other than Merlin's, as it should remain the "enabled-by-default" connector.

This plugin and the associated `package.json` import and `.eslintrc` rule should perhaps be removed by the `platform-generator` when a new project is generated.

## How to test-drive this PR
```bash
git checkout add-merlins-connector-eslint-rule
npm i
```

- Change to the SFCC connector
- Go into your editor and disable -> re-enable eslint (restarts eslint server), then check `main.jsx` for errors
- Or run `npm run lint`

screen shot 2017-06-07 at 3 09 43 pm

screen shot 2017-06-07 at 3 11 19 pm

## Resources

ESlint uses an abstract syntax tree (AST) to navigate JavaScript code, to see what the AST looks like for any set of code, paste it into the sandbox editor in the link below
- http://astexplorer.net/


web/eslint-plugin-merlins-connector/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "eslint-plugin-merlins-connector",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "author": "",
+  "license": "ISC",
```
jeremywiebe: I'd probably just omit many of these as we aren't publishing this: `description`, `author`, `license` could all go

web/eslint-plugin-merlins-connector/index.js
```diff
@@ -0,0 +1,19 @@
+module.exports.rules = { // eslint-disable-line
```
jeremywiebe: This plugin is very specific to the `/app` area.  What do you think about pushing this whole folder down into `/web/app/utils/eslint-plugin-merlins-connector`?  I'm a bit hesitant about adding another top-level dir for a utility like this. :)

PR #718, opened 2017-06-07
Fix How we're running Tests with Preview on Production

jasecode:
On master branch, the TEST_PROFILE=production was not being set as start-test-server.sh sets it in the background.

Change this by adding a few lines to check for git branch and export the production variable on a foreground process - run-all-tests.sh

Add run-lighthouse.sh script to have logic of when to run test:pwa-ci vs test:pwa-prod

## Changes
- web/tests/system/run-all-tests.sh 
-web/tests/system/run-lighthouse.sh

##TODOs
- Should also merge to develop to update

## How to test-drive this PR
Change the line 10 on run-all-tests.sh 
"[ "$CURRENT_BRANCH" == "master" ]' to the current branch (for testing purposes to force 'production' to get exported) 
- ./web/tests/system/run-all-tests.sh
- Nightwatch should be testing on the production
- Production tests should fail right now due to us not publishing a bundle for v0.17.0


web/tests/system/run-all-tests.sh
```diff
@@ -2,6 +2,18 @@
 set -o pipefail
 set -o nounset
 
+if git rev-parse ; then
+    # Get the current branch on CircleCI or local
+    CURRENT_BRANCH=${CIRCLE_BRANCH:-$(git branch | grep "*" | awk '{ print $2 }')}
+    # If current branch is master, we eliminate preview because we're testing production'
+    if [ "$CURRENT_BRANCH" == "master" ]; then
+      echo "On production branch, test server not needed."
+      export TEST_PROFILE=production
```
ellenmobify: Do we still need to export `TEST_PROFILE` in `start-test-server.sh`?

jasecode: I left it there to be safe, I'm not sure if we will ever "foreground" it again. 🤔  but I guess I could remove it?

ellenmobify: 👍

PR #719, opened 2017-06-08
Loads scripts synchronously to improve load performance

jansepar:
Loads scripts synchronously to improve load performance

 **JIRA**: N/A
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/673

## Changes
- Loads scripts synchronously in order for the browser to increase the priority of the scripts. The delay due to not being high priority can be up to 5 seconds on a Moto G4.
- Moves a number of things that the tag used to be responsible for into the loader:
  - Adds preview logic into loader. Previously, this was the job of the tag.
  - Adds the "opt-in" logic to the loader. This means going forward if we wanted to opt-in more browsers, we could do it without a tag change.
  - Adds a.js loading for non-supported browsers to the loader, since this is no longer something the tag is responsible for. 

## Notes
- The loader will not parse on IE8 or lower. However, the failure does not prevent the page from continuing to load, or other scripts from running. Tested on SauceLabs with IE6, 7 and 8.

## Loader blocking TODOs
- [x] Fix loading push
- [x] Make loader backwards compatible with V7 tag
- [x] Add conditional logic for loading asynchronously in order for the build to work on 2G (on Chrome, you cannot document.write when on a 2G connection)
- [x] Fix tests

## Loader non-blocking TODOs
- [ ] Refactor logic into SDK

## How to test-drive this PR
You'll need to use Charles to replace the existing tag on Merlin's with the new V8 tag.
- First, ensure you have SSL proxying enabled for www.merlinspotions.com (Proxy -> SSL Proxying Settings -> "Enable SSL Proxying", then add "www.merlinspotions.com").
- Add a rewrite rule (Tools -> Rewrite...), and match the following screenshot (notice how basic the tag is!):

2017-06-07 06 10 28 pm

- You also need to map remote `https://cdn.mobify.com/sites/progressive-web-scaffold/production/*` to `https://localhost:8443/`, since the current live bundle won't have any of the preview logic we need to test. Match the following screenshot:

2017-06-07 06 09 57 pm

- Checkout [this branch](https://github.com/mobify/progressive-web-sdk/pull/673) of the SDK locally, run `npm link && npm run dev:build` in that folder, then in `platform-scaffold/web`, run `npm link progressive-web-sdk`.
- Then run `npm run dev`
- Everything should be working as normal! You should also see in the Network tab that requests should be loaded with "High" or "Medium" priority (to do this, enable the priority column in the Network tab).
- Then you'll need to test with a few other combinations:
  - Try without preview. The site should still work, and priority should still be High/Medium.
  - Try with preview, but with the "Rewrite" rule disabled, to ensure this build still works with the old tag (priority will be low, since it's the old synchronous style of loading).
  - Try without preview, still with "Rewrite" rule disabled, to ensure this build still works with the old tag when not previewing (priority will be low, since it's the old synchronous style of loading).


benlast:
@jansepar it looks to me like this will actually load the Messaging client sooner, since there's no longer any waiting on `Promise`s for the loading of the other scripts. This isn't a problem, but the Messaging client will also make at least one network request when it's initialized, so the original idea was to defer all of the Messaging setup until the app was loaded and stable. This is just an observation, though - as long as you're seeing the improvement you want!

web/app/loader.js
```diff
@@ -1,14 +1,20 @@
 /* global NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, MESSAGING_ENABLED, DEBUG */
 import {getAssetUrl, getBuildOrigin, loadAsset, initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
-import {isSamsungBrowser, isFirefoxBrowser} from 'progressive-web-sdk/dist/utils/utils'
+import {
+    isSamsungBrowser,
+    isFirefoxBrowser,
+    preventDesktopSiteFromRendering,
+    loadScript,
+    loadScriptAsPromise
+} from 'progressive-web-sdk/dist/utils/utils'
```
jnurse: I'm not seeing most of these utils in the SDK branch, and I can't get the merlin's homepage to load with this branch

web/app/loader.js
```diff
@@ -19,301 +25,305 @@ import preloadHTML from 'raw-loader!./preloader/preload.html'
...
+    // loading. This delay can go all the way up to 5 seconds on a Moto G4 on a
+    // 3G connection.
+    loadScript({
+        id: 'progressive-web-vendor',
+        src: getAssetUrl('vendor.js'),
+        docwrite: loadScriptsSynchronously,
+        isAsync: false
+    })
+
+    loadScript({
```
jnurse: Is there any concern with loading these assets before the service worker has been installed?

web/app/loader.js
jansepar: We discovered this was because I hadn't pushed my latest changes.

web/app/loader.js
jansepar: Kinda sorta not really... So the idea behind throwing these all *after* the SW is installed was so that we could make sure the SW caches them. But there apparently is an issue with the SW (at least the last time @DanielBrookRoberge looked) where even when you get the "ready" event, that doesn't guarantee that the SW has actually registered and all future network requests will go through it.

So right now, the only way we'll be guaranteed that this stuff is in the SW cache is by someone doing a second full page load (because the SW will be registered by then, thus the requests will get cached).

The real fix for this is to prefetch this content to it's guaranteed to get in the cache, but due to a limitation of our CDN (as written up by @DanielBrookRoberge [here](https://docs.google.com/document/d/1RhvE0-qHRLQkwrTyU-jyiExjRfGf2PxpyHh1aG4VfG0/edit#heading=h.i9xjchpahnwj)), we can't yet get this optimal setup.

web/app/loader.js
```diff
@@ -19,304 +26,318 @@ import preloadHTML from 'raw-loader!./preloader/preload.html'
...
+    if (isSupportedBrowser() && isPWARoute()) {
+        if (neededPolyfills.length) {
+            neededPolyfills.forEach((polyfill) => polyfill.load(attemptToInitializeApp))
+        } else {
+            attemptToInitializeApp()
+        }
+    } else {
+        // If it's not a supported browser or there is no PWA view for this page,
+        // still load a.js to record analytics.
+        loadScript({
```
benlast: @jansepar a couple of reminder questions here:

1. If I understand the plan, at this point the loader would load `desktop.js` (should we call it `non-pwa.js` to be clearer?)
1. Should this loader **also** pull in `a.js`?
1. Should those scripts load synchronously?

The big difference we have, of course, is that desktop/non-PWA operation has to load everything on every pageview, and I'm guessing that analytics scripts might need to be loaded fast (as opposed to messaging that usually is happy loading later).

jansepar: Re: 1 - that's right! I think `desktop.js` would be fine, although I guess the plan is to have pwas running on desktop so maybe not...

Re: 2 - I think it *could* (and probably should for now to load things like the Analytics SDK (cc @wizardlyhel @sketchytechky )

Re: 3 - Nope these scripts will load async. We don't want to risk blocking rendering of our customers desktop sites and then blame us for slow performance.

jansepar: Oh and as for earlier analytics firing... we actually have `sandy` loaded as part of the loader. If we want to fire some earlier events in desktop.js or in the `else if desktop` condition we totally can.

benlast: Sounds good to me. I'll wait for the Super Analytics League to add their comments...

I would argue for loading `a.js`, and doing it **after** the desktop loader, so that the desktop loader gets a chance to set flags that suppress other code. For example, old messaging (pulled in via a.js) will spot that new messaging is being loaded and do nothing.

web/app/loader.js
```diff
@@ -19,304 +26,318 @@ import preloadHTML from 'raw-loader!./preloader/preload.html'
...
+            id: 'progressive-web-app',
+            src: ASTRO_CLIENT_CDN,
+            rejectOnError: false
+        }).then(() => window.Astro)
+
     }
+
+    // Attempt to load the worker.
+    (('serviceWorker' in navigator)
+        ? loadWorker()
```
benlast: On desktop/non-pwa, my guess is that we should pass a parameter to the worker, in the worker URL, to tell it whether to work in PWA mode or not. That allows the existing SDK bootstrap to remain the same for the PWA and non-PWA sites (important since the customer site has to serve it), and the variation in behaviour can then be done in the Scaffold's `web/worker/main.js` (it should not init the SDK `worker` module in non-PWA mode, just the Messaging worker).

wizardlyhel:
@jansepar Couple things that comes into mind:
**Tag**
* How will the tag handle the case when `loader.js` failed at the CDN level?
* Old tag had a start timing - Let's not lose that in new tag

**`loader.js`**
* How will `loader.js` handle the case when one or more critical bundle file fails? (I don't see logic to fall back to desktop)
* Opt-out logics are missing (We should instrument this at `loader.js` level because `a.js` is requested at this level)

Trying to run this locally but getting error in `loader.js`: Unable to verify analytics
screen shot 2017-07-10 at 11 09 18 am

jansepar:
> How will loader.js handle the case when one or more critical bundle file fails? (I don't see logic to fall back to desktop)
> Opt-out logics are missing (We should instrument this at loader.js level because a.js is requested at this level)

The tag itself has an onerror handler (https://github.com/mobify/mobify-tag/pull/3/files#diff-f4ffaea63771411e93b1a9c766f56ed8R10). But it's not necessary to set an opt-out if other files (like main.js) fail to load because if we've managed to load loader.js, then we can assume our CDN is up or you have the assets in your cache.

jnurse:
I mentioned this to Shawn offline, but I'm getting some errors that are preventing the page from rendering when testing in an iOS simulator. I'm seeing these errors with and without previewing when I have the new tag added using a re-write rule in Charles. 


screen shot 2017-07-24 at 10 47 59 am

web/app/containers/app/container.jsx
```diff
@@ -71,16 +71,21 @@ class App extends React.Component {
...
 
             // Only after we loaded the CSS can confidently unhide the app.
             // This is necessary, because showing the app by default might show
             // a flash of an ugly, unstyled app until the CSS finally loads.
             this.props.toggleHideApp(false)
         } else {
-            setTimeout(this.hidePreloaderWhenCSSIsLoaded, 100)
+            console.log("[Mobify.Progressive] Waiting for CSS to be loaded by checking for window.Progressive.stylesheetLoaded")
+            counter++
+            setTimeout(() => this.hidePreloaderWhenCSSIsLoaded(counter++), 100)
```
jnurse: Should this counter be incremented twice in this block?

web/app/loader.js
```diff
@@ -187,130 +279,118 @@ const attemptToInitializeApp = () => {
...
+     * This must be called before vendor.js is loaded (or before the Webpack
+     * chunk that contains Messaging React components is loaded)
+     *
+     * This creates a Promise: `window.Progressive.MessagingClientInitPromise`
+     * which will be resolved or rejected later by the method `setupMessagingClient`
+     */
+    createGlobalMessagingClientInitPromise(messagingEnabled)
+
+    window.loadCriticalScripts = () => {
+        // The following scripts are loaded async via document.write, in order
```
jansepar: Yep!

jeremywiebe:
Commit 2fea346 applied for https://github.com/mobify/portal_app/blob/8c759c231bdaacd43594b1a996207e9004f280e5/portal/preview/views.py#L86

jeremywiebe:
Talked to @jansepar .  Reverting the supported browser change and instead tweaking the lighthouse script to use a full, regular mobile user agent with `MobifyPreview` tacked on. This is enough to satisfy the portal commit changes noted above.

PR #721, opened 2017-06-08
set loggedin to false when user is signed out

gsaray:
sign out doesn't get updated until you browse to another page

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1391

## Changes
- set loggedin to false in signOut function

## How to test-drive this PR
- Run `npm run dev`
- Open menu and sign in
- Open menu and sign out, check the text is back to sign in


web/app/containers/app/actions.js
```diff
@@ -83,6 +83,7 @@ export const signOut = () => (dispatch) => (
     dispatch(logout()).then(() => {
         // Desktop's message includes 'redirect to home page' message
         // so we'll just hardcode a message instead
+        dispatch(setLoggedIn(false))
```
jnurse: The SFCC connector is already setting this within it's logout command, maybe this should go into the merlins logout command instead of here?

PR #723, opened 2017-06-08
[WEBDATA-127] Edit Cart functionality on SFCC Connector

Cdok:
# Edit Cart functionality on SFCC Connector

This PR adds the functionality to edit / update a cart item on the Salesforce Commerce Cloud Connector

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-127

## Changes
- Adds new command to update cart items in both Merlin's & SFCC connectors
- Changes the configureURL on `Edit Cart` pages to have `item_id` in the URL instead of `basket_id`, as `basket_id` is available through session storage.

## Todos
- [ ] +1
- [ ] Ensure analytics changes make sense for new Merlin's command (maybe a new analytics action for updating cart)

## How to test-drive this PR
- Using both connectors:
- Run `npm run dev`
- Add an item to cart, click Edit Cart, update the item, and ensure it works as intended


web/app/containers/product-details/actions.js
```diff
@@ -51,6 +51,8 @@ const submitCartFormSelector = createPropsSelector({
 
 export const submitCartForm = (formValues) => (dispatch, getStore) => {
     const {productId, qty, variations} = submitCartFormSelector(getStore())
+    const isInCheckout = /configure/.test(window.location.pathname)
+    const itemId = window.location.pathname.match(/(?:\/id\/)(.*?)(?:\/product_id\/)/)
```
Cdok: I feel like it might be a good idea to have some sort of flag in the redux store for if the shopper enters `configure` mode. Right now, the only way to check is via the URL / current route.

It would definitely be more future-proof / connector-proof if we had some action we could dispatch when a user enters and exits "edit" mode for a given cart item, instead of relying so heavily on the URL.


Something along the lines of:

```javascript
{
    cart {
        edit: {
            enabled: true,
            item_id: ID_BEING_EDITED
        }
     }
}

File: web/app/containers/product-details/actions.js [hunk 2]
```diff
@@ -51,6 +51,8 @@ const submitCartFormSelector = createPropsSelector({
 
 export const submitCartForm = (formValues) => (dispatch, getStore) => {
     const {productId, qty, variations} = submitCartFormSelector(getStore())
+    const isUpdatingCart = /configure/.test(window.location.pathname)
+    const itemId = window.location.pathname.match(/(?:\/id\/)(.*?)(?:\/product_id\/)/)
```
Cdok: I feel like it might be a good idea to have some sort of flag in the redux store for if the shopper enters `configure` mode. Right now, the only way to check is via the URL or current route.

It would definitely be more future-proof / connector-proof if we had an action to dispatch when a user enters and exits "edit" mode for a given cart item, instead of relying so heavily on the URL.


Something along the lines of:

```javascript
{
    cart {
        configure: {
            enabled: true,
            item_id: ID_BEING_EDITED
        }
     }
}
```

Definitely open to discuss the approach on this

Cdok: I feel like it might be a good idea to have some sort of flag in the redux store for if the shopper enters `configure` mode. Right now, the only way to check is via the URL or current route.

It would likely be more future-proof / connector-proof if we had an action to dispatch when a user enters and exits "edit" mode for a given cart item, instead of relying so heavily on the URL.

Having a `configure` key where we could keep any `item_id` or associated information that might be required in the `PATCH` request might be useful


Something along the lines of:

```javascript
{
    cart {
        configure: {
            enabled: true,
            item_id: ID_BEING_EDITED
        }
     }
}
```

Definitely open to discuss the approach on this

PR #725, opened 2017-06-08
Fail gracefully throughout checkout when basket expires

jnurse:
This PR updates the commands on the checkout shipping and checkout payment pages to fail gracefully (ie. send the user to the homepage with a notification) when the cart has expired. 

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-108
 **Linked PRs**: n/a

## Changes
- Wrapped checkout shipping and checkout payment init commands in container actions so we can handle expiry error messages
- Added expired basket error throwing to init checkout payment and checkout shipping SFCC commands
- Added expired basket error throwing to submit shipping and submit payment SFCC commands
- Added logic to handle submit shipping & submit payment commands throwing expired basket errors
- Added logic to handle expired basket errors in fetchShippingMethods action 

## How to test-drive this PR
- switch to the SFCC connector (in web/app/main.jsx)
- Run `npm run dev`
- Preview the [SFCC Site](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- add an item to the cart and proceed to the checkout shipping page
- wait on the page for a while so the cart will expire
- refresh the page
- you should be sent to the homepage with a notification stating your cart has expired
- add another item to the cart and proceed to the checkout shipping page
- fill out the form with valid data
- wait on the page for a while so the cart will expire
- try submitting the form to move to the payment page
- you should be sent to the homepage with a notification stating your cart has expired
- add another item to the cart and proceed to the checkout payment page
- wait on the page for the cart to expire
- refresh the page
- you should be sent to the homepage with a notification stating your cart has expired
- add another item to the cart and proceed to the checkout payment page
- fill out the form with valid information
- wait on the page for the cart to expire
- try submitting the order 
- you should be sent to the homepage with a notification stating your cart has expired


web/app/containers/checkout-payment/actions.js
```diff
@@ -62,4 +74,11 @@ export const submitPayment = () => (dispatch, getState) => {
                 pathname: url
             })
         })
+        .catch((error) => {
+            const message = error.message
+            if (message.includes('expired')) {
+                return dispatch(handleCartExpiry())
+            }
+            throw error
+        })
```
DanielBrookRoberge: Maybe it makes sense to split this handler out into a function?

web/app/containers/cart/actions.js
```diff
@@ -46,7 +46,7 @@ export const submitEstimateShipping = () => (dispatch, getState) => {
             const message = error.message
 
             if (message.includes('expired')) {
-                return dispatch(handleCartExpiry())
+                return dispatch(cartExpired())
             }
             return dispatch(addNotification(
```
DanielBrookRoberge: This could still use the `handleCartExpiryError` function as 
```javascript
.catch((error) => dispatch(handleCartExpiryError(error)))
.catch(() => dispatch(addNotification(...)))
```

web/app/integration-manager/_sfcc-connector/checkout/commands.js
```diff
@@ -149,11 +157,11 @@ const addPaymentMethod = (formValues, basket) => (dispatch, getState) => {
...
     }
 
-    return makeApiJsonRequest(
+    return makeApiRequest(
         `/baskets/${basket.basket_id}/payment_instruments`,
-        requestBody,
-        {method: 'POST'}
+        {method: 'POST', body: JSON.stringify(requestBody)}
     )
+    .then((response) => response.json())
```
DanielBrookRoberge: What error handling does `makeApiJsonRequest` do that causes problems here? The error handling code there was really an educated guess based on the code as it existed, maybe we'll need to change it?

jnurse: It throws an error when `!response.ok` and doesn't parse the response as JSON, so when the cart requests return with a 400 error an error gets thrown and the response isn't parsed into JSON so we can't examine the `fault` key. If it does reach the part that parses the response into JSON it again throws an error with the fault message rather than letting us check it ourselves. 

It might be worth checking where we're using it currently, maybe we can remove those checks. It probably makes sense to be less opinionated about how these errors are handled since each request may need a different response to a fault.

DanielBrookRoberge: If we get meaningful JSON payloads in a number of `!response.ok` cases, we can maybe just remove that check and just do `(response) => response.json()`. If we're getting an error with no JSON then this will throw anyways.

jnurse: Good idea!

PR #726, opened 2017-06-08
Fix AMP-specific validation errors (SVGs, forbidden style rules, etc)

olibrook:
Fixes some amp-validation errors that snuck in when we added SVG sprites on AMP. I had to hand-modify some shared SVGs to pass validation - would be good to get @jeffkamo to check that these still look okay for the PWA.

## How to test-drive this PR
- Run `npm run dev`
- Preview at `localhost:3000` and check for validation errors using the amp validator plugin.


cole-sanderson:
screen shot 2017-06-08 at 2 35 58 pm

cole-sanderson:
Looks good to me. 👍

PR #727, opened 2017-06-08
AMP Integration Manager...integration

kschmidtdev:
Sharing parsers/integration manager backend + selectors with the PWA

 **JIRA**: https://mobify.atlassian.net/browse/AMP-43, ninja-fix to https://mobify.atlassian.net/browse/AMP-64

## Changes
- ~~Bunch of shims in `main.js` to make the integration manager codepaths work~~
- ~~Bunch of pass-throughs and whatever else required in `webpack.config.js`~~
- Brought in our own ES6-ified capture.js 'disable' function
- Made `jqueryResponse` externally configurable to the merlins connector
- Implemented our own ES6-ified jqueryResponse
- Made the web scaffold a little more tolerant of running in a browserless environment
- Hooked up selectors in our existing templates

## How to test-drive this PR
- Run `npm i` from the `web` directory - this is a new requirement! (I don't have a solution to this yet)
- Run `npm run dev`
- Check out http://localhost:3000, http://localhost:3000/eye-of-newt.html, http://localhost:3000/potions.html
- Run `npm run test:all`


web/app/integration-manager/_merlins-connector/app/commands.js
```diff
@@ -36,7 +36,9 @@ const requestCapturedDoc = () => {
 let isInitialEntryToSite = true
 
 export const fetchPageData = (url) => (dispatch) => {
-    const request = isInitialEntryToSite ? requestCapturedDoc() : makeRequest(url)
+    const request = isInitialEntryToSite && window.Progressive.capturedDocHTMLPromise
```
jeremywiebe: Is `window.Progressive.capturedDocHTMLPromise` undefined in AMP builds?  If so, could you drop a comment above this line explaining why we have to check for that Promise?

amp/app/containers/product-list/container.jsx
```diff
@@ -4,41 +4,47 @@ import AmpImage from 'mobify-amp-sdk/dist/components/amp-image'
...
+        <h1>Number of items: {numItems}</h1>
+        {
+            products.map((prod) =>
+                <div key={prod.id}>
+                    <h2>{prod.title}</h2>
+                    <AmpImage src={prod.thumbnail.src} width="240" height="300" layout="fixed" />
+                    <p>{prod.price}</p>
+                </div>
+            )
+        }
```
olibrook: I made a little change here sot that it's obvious we're really fetching data using the Integration Manager.

amp/app/data-integration/capturejs-disable.js
```diff
@@ -0,0 +1,94 @@
+// From https://github.com/mobify/capturejs/blob/master/src/capture.js
```
olibrook: I have absolutely no idea why we need this, beyond that the Integration Manager probably references it. We never 'capture' a page on the server, why would we need a `captureDisable` function?

amp/app/data-integration/connectedStore.js
```diff
@@ -0,0 +1,67 @@
...
+
+import appReducer from '../../../web/app/containers/app/reducer'
+import footerReducer from '../../../web/app/containers/footer/reducer'
+import headerReducer from '../../../web/app/containers/header/reducer'
+import homeReducer from '../../../web/app/containers/home/reducer'
+import navigationReducer from '../../../web/app/containers/navigation/reducer'
+import productDetailsReducer from '../../../web/app/containers/product-details/reducer'
+import productListReducer from '../../../web/app/containers/product-list/reducer'
+import categoryReducer from '../../../web/app/store/categories/reducer'
+import productReducer from '../../../web/app/store/products/reducer'
```
olibrook: I'm concerned about the amount of references there are to the PWA here. We had discussed moving the IM code into the SDK. Will these reducers be included, as the operate on Mobify's common data schema?

amp/app/data-integration/connectedStore.js
```diff
@@ -0,0 +1,67 @@
...
+export const createConnectedStore = (window, fullUrl, dataInitFunction) => {
+    // This is okay to pass to both SFCC and Merlin's connectors,
+    // as Merlin's doesn't need a configuration object
+    registerConnector(Connector({
+        siteID: '2017refresh',
+        clientID: '5640cc6b-f5e9-466e-9134-9853e9f9db93'
+    }))
+
+    global.window = window
+    global.Capture = {disable: (...args) => captureDisable(...args)}
```
olibrook: wat?

amp/app/main.js
```diff
@@ -23,34 +21,39 @@ import App from './containers/app/container'
...
 
-export const parse = (window, html) => {
-    const $ = window.$
-    const $html = $(html)
-    return {
-        links: $.map($html.find('a'), (el) => $(el).text()),
-        title: $html.find('h1').text()
-    }
+const getFullUrl = (req) => {
+    return `${ampPackageJson.siteUrl}${req.url}`
```
olibrook: Does this include GET parameters?

amp/app/data-integration/connectedStore.js
olibrook: The global `window` reference is shared and will not be `===` between requests using this pattern.

olibrook:
I'm convinced that this works - that is good. I have some big-picture issues I want to comment on though.

---

For me, this PR is the most important one we will have to merge on AMP because it should demonstrate the fundamental value proposition of our choices to use React and to re-use common code for integration mangers, etc. To succeed it should:

1. *Be safe to reuse code* - I should be able to import the Integration Manage/selectors from an SDK or a common folder, without worrying about changes to the PWA.

2. *Run in a familiar environment*: I should be able run it under NodeJS (or even in a JSDom environment) without patching all over the place to make the environment look like a web-browser.

3. *Use sane concepts*: I shouldn't have to deal with ideas like "capturing" or "un-capturing" - these are totally alien outside the context of a web-browser. AMP should clearly be either a browser or a server-side app.

--- 

Merging this PR sets a precedent for how we deal with the IM. How can we ensure we are moving in the direction of the above aims? My suggestions:

- Work out why the Integration Manager depends on `window` and fix.
- Work out why the Integration Manager depends on `Capture` and fix.
- Take steps to move IM code and related selectors into an SDK.

I think it is legitimate to ask if the IM is fit for our needs at this time, rather than jumping through hoops trying to get it to work.

Thoughts?

amp/app/data-integration/capturejs-disable.js
kschmidtdev: This goes through the markup and rewrites various tags with `x-` prefixed (well, you can pass in whatever prefix you like. We don't actually care about this, but there are parser queries later that depend on the `x-` prefix for some elements. We could update them to be tolerant of `x-` or not - and then we could completely drop this code.

amp/app/main.js
kschmidtdev: According to https://nodejs.org/api/http.html#http_message_url, yes (also I just tested it - confirmed it works)

amp/app/data-integration/capturejs-disable.js
olibrook: That sounds to me to be a design problem in the IM. Ideally "capturing" would only be used to prevent effects on the loaded page. Shouldn't the actual scraping always be done on the HTML as pulled straight off the server?

Ie. on the PWA side

```
  captured = capture(document)
  uncaptured = uncapture(html)
  IntegrationManager.parse(uncaptured)
```

That would give the IM an interface that was usable outside of the context of a PWA.

amp/app/data-integration/connectedStore.js
kschmidtdev: This is a solution to accessing `window.document` and `window.$` from deep within some dependencies where simply `window` is undefined - I don't know how to execute these within the `jsdom` context so it is defined, so resorted to wrapping them via webpack code substitution and redirecting them through things off of `global`..

amp/app/data-integration/capturejs-disable.js
kschmidtdev: This makes sense, I'd like to address it in a follow-up PR since it would require changes to the PWA SDK (there is a util that runs the `disable` against fetched pages)

amp/app/data-integration/connectedStore.js
kschmidtdev: Yes, the reducers will go along with the selectors to the SDK

amp/app/data-integration/capturejs-disable.js
DanielBrookRoberge: The method by which we convert the raw HTML to a jQuery object requires us to "disable" the markup using Capturing. If you just pass the HTML string as `$(myHtml)` the resulting jQuery object only has the contents of the `body` tag, but sometimes parsers need to read attributes off the body tag itself or from the header. The solution involves sticking the HTML into an iframe and reading it that way; if we leave all the `src` attributes and the like intact they would request assets when we do that.

See https://github.com/mobify/progressive-web-sdk/blob/develop/src/get-jquery-html.js for more details.

amp/app/data-integration/connectedStore.js
DanielBrookRoberge: The latter two will be moved into the SDK, but the former two will not be.

kschmidtdev:
@olibrook 

1. When using Integration Manager + Selectors + Reducers, whether they are in the PWA or a SDK, we actually do want to worry about changes. If the desktop markup changes and parsing breaks, we don't want a fix to go out for the PWA and AMP is left broken (or we have to fix every breakage twice). Most of this stuff is scheduled to make its way to a SDK anyways.

2 + 3. This is the _Merlin's Connector_ (desktop parsing) rather than issues with Integration Manager + selectors + reducers. Just to clarify and focus what we're talking about.

https://github.com/mobify/progressive-web-sdk/blob/develop/src/jquery-response.js#L36
https://github.com/mobify/progressive-web-sdk/blob/develop/src/get-jquery-html.js#L8

^ The specific issue surrounds getting a ('disabled') jQuery object of the document. The parsers actually pass around the document and jQuery, so they don't depend on `window` themselves. They do depend on some stuff being named a certain way due to `disable`-ing. 

We could provide our own way of getting a 'disabled' jQuery object of the document? It would reduce the amount of patching required. I'd prefer to incrementally making changes to the SDK + web scaffold + remove patching rather than attempt to do it in one step with this PR.

amp/app/data-integration/connectedStore.js
olibrook: Yep - I understand the motivation. I was pointing out that this is mutable, global state shared across handlers in event-driven code.

Are the two requests below racing to set/read the window object. Is this a problem in our code?

```
// Obviously a stupid example
const handler(req) => {
  global.window = JSDOM()
  global.window.doSomethingStateful()
  fetch(url)
    .then(content => parse(window.$))
}

handler(new Request())
handler(new Request())
```

kschmidtdev: That's a great point. The only stateful thing we are doing is temporarily creating an iframe element for the whole 'jQueryfication' thing. 

So we could actually just instantiate `jsdom` once - on start-up.

amp/package.json
```diff
@@ -9,9 +10,11 @@
   },
   "devDependencies": {
     "amphtml-validator": "^1.0.20",
+    "archiver": "1.3.0",
```
olibrook: Why?

amp/app/data-integration/connectedStore.js
```diff
@@ -0,0 +1,61 @@
...
+import navigationReducer from '../../../web/app/containers/navigation/reducer'
+import productDetailsReducer from '../../../web/app/containers/product-details/reducer'
+import productListReducer from '../../../web/app/containers/product-list/reducer'
+import categoryReducer from '../../../web/app/store/categories/reducer'
+import productReducer from '../../../web/app/store/products/reducer'
+
+import {jqueryResponse} from './capturejs'
+import {PAGE_TITLE, DATA_INIT_FUNCTION} from './constants'
+
+export const createConnectedStore = (window, fullUrl, dataInitFunction) => {
```
olibrook: This is waaaay nicer than the `window` hacks from before. Some more thoughts:

1. I'd combine this function with `main.initializeStore` which doesn't really do anything anymore.
   Optional - that's just what I'd do.

2. It is bad-practice to put anything that is not serializable in the Redux store. We can fix this AND improve 
   ergonomics, I think:
   
   ```
   // If we change `initializeStore` to this
   const initializeStore = (req, components) => {
        // ...snip
        const renderProps = {
            location: {},
            components: components,
            history: {}
        }
        return waitForResolves(renderProps, createdStore)
   }
   
   // Then each top-level component can do this
   const Home () => {}
   Home.resolves = (dispatch, getState) => dispatch(getHomePageData)

   // And we can remove the DATA_INIT_FUNCTION slice of the app state completely.
   ```

   That keeps the Redux store serializable and keeps the data-fetching/dispatching code closer
   to the components that actually depend on it.

   See - http://redux.js.org/docs/faq/OrganizingState.html#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

amp/app/data-integration/connectedStore.js
```diff
@@ -0,0 +1,61 @@
...
+        thunk,
+    ]
+
+    const noop = (f) => f
+
+    const initialState = ({ui: {app: fromJS({
+        [CURRENT_URL]: fullUrl,
+        [PAGE_TITLE]: 'Merlins AMP', // Fetch the page again and get title?
+        [DATA_INIT_FUNCTION]: dataInitFunction
+    })}})
```
olibrook: This is a mutable object containing a mutable object containing an immutable object. Shouldn't it be Immutable all the way down?

Pretty sure you can just do:

```
fromJS({ui: {app: {
  [CURRENT_URL]: fullUrl,
  ...
}}})
```

amp/webpack.config.js
```diff
@@ -29,6 +28,14 @@ module.exports = {
...
             },
             {
+                test: /\.js(x?)$/,
+                include: /progressive-web-sdk/,
+                use: [
+                    {loader: "imports-loader?window=>{location: {href: ''}},\
+                                             navigator=>{userAgent: ''}"}
+                ],
+            },
+            {
```
olibrook: Is this still needed with the `window` fixes?

Why does `/progressive-web-sdk/` need to be included? Can't we import the built assets from `dist`?

amp/package.json
kschmidtdev: I guess bad merge? Removing

amp/webpack.config.js
kschmidtdev: These are patches specific to `progressive-web-sdk` because there is code that is evaluated immediately upon import that requires these objects to exist. I will be following up with the web team to see if we can get updates in the future so we can remove these patches.

amp/app/data-integration/connectedStore.js
kschmidtdev: I tried that. I don't think that's how we have things set up - I think we keep the high level structure as JS objects and then data that is more 'leafy' are Immutable (see https://github.com/mobify/platform-scaffold/blob/develop/web/app/containers/cart/partials/cart-estimate-shipping.test.js#L21)

web/app/containers/app/reducer.js
```diff
@@ -12,7 +12,7 @@ import {setCheckoutShippingURL, setCartURL, setCurrentURL} from '../../integrati
 import {CURRENT_URL} from './constants'
 
 export const initialState = fromJS({
-    [CURRENT_URL]: window.location.href,
+    [CURRENT_URL]: typeof window !== 'undefined' ? window.location.href : '',
```
kschmidtdev: @jeremywiebe any thoughts on my ugly attempt at making code not depend on the browser?

web/app/integration-manager/_merlins-connector/app/commands.js
```diff
@@ -35,12 +34,18 @@ const requestCapturedDoc = () => {
 
 let isInitialEntryToSite = true
 
+const runningInBrowser = () => typeof window !== 'undefined' && window.Progressive
```
kschmidtdev: @jeremywiebe another one

web/app/utils/astro-integration.js
```diff
@@ -4,7 +4,7 @@
 
 import isRunningIn from '../vendor/astro-detect'
 
-export const isRunningInAstro = isRunningIn.app()
+export const isRunningInAstro = typeof window !== 'undefined' && isRunningIn.app()
```
kschmidtdev: another one

kschmidtdev: @jeremywiebe another 'not a browser' compatibility change

amp/webpack.config.js
kschmidtdev: Working on a way we might be able to avoid patching this...

kschmidtdev: Unfortunately though a chain of imports leading to code we never use, we can't get rid of the `window.location.href` patch yet - will require a change to the PWA SDK.

web/app/containers/app/reducer.js
kschmidtdev: We're not using `app/containers/app/selectors.js` directly, but `app/containers/product-details/selector.js` and others depend on it.

Maybe this is a good candidate for making our own reducer since most of the contents don't apply to AMP. We have our own state that we initialize the server's redux store with, and it populates `CURRENT_URL`. This one is evaluated in import, which breaks for us.

web/app/integration-manager/_merlins-connector/app/commands.js
kschmidtdev: Maybe some simple wrapper util - maybe limited to the connector scope for now?

web/app/containers/app/reducer.js
kschmidtdev: Made our own minimal reducer and no longer need to make this change to the scaffold 🎉

web/app/integration-manager/_merlins-connector/app/commands.js
kschmidtdev: There is only one case right now so maybe this is enough until we use it more.

olibrook:
@kschmidtdev I think you've fixed the bits I was worried about, at least to the point at which I'm happy to merge this and see where we get to using the integration manager.

👍

PR #732, opened 2017-06-09
Build AppVeyor only on master and develop

johnboxall:
Okay, this time for real _reals_:

We found that AppVeyor seems? to ignore the settings configured through the web ui if you provide a appveyor.yaml file. So, we've moved how to ignore builds to this file.

We also need to disable AppVeyor's "Pull Request" webhoook as per appveyor/ci#882

Changes

AppVeyor builds only on master/develop


ellenmobify:
👍

PR #741, opened 2017-06-13
Update Lighthouse to Run test:pwa-prod on Master

jasecode:
Update Lighthouse to Run test:pwa-prod on Master

Currently, we have no differentiation of when to use our lighthouse tests and we're always running test:pwa-ci on circle regardless of it being on master or not. This PR is to add logic so we know to run test:pwa-prod on master.

## Changes
- add run-lighthouse.sh
-- Checks for the branch and runs the appropriate lighthouse command
- check-score.js
-- Refactored
-- Added additional condition to check for the audit-prod.report command

## How to test-drive this PR
- rerun build
- Edit run-lighthouse.sh , to run on your current branch to test if the test:pwa-prod works


web/scripts/lighthouse-prod.sh
```diff
@@ -6,6 +6,9 @@ URL=${1-$npm_package_siteUrl}
 
 lighthouse \
 	--view \
+	--output json \
 	--output=html \
```
ellenmobify: Let's make it consistent and have this be `--output html \`

web/lighthouse/check-score.js
```diff
@@ -1,33 +1,53 @@
 #!/usr/bin/env node
 
 const fs = require('fs')
-
+// import fs from 'fs'
```
ellenmobify: remove

johnboxall:
@jasecode What errors are we hoping to catch running `npm run test:pwa-prod` on CI?

jasecode:
@johnboxall 
We do the same check-score.js against the threshold just like in test:pwa-ci.

Previously, we're running test:pwa-ci when the branch is on master and the test-server not does run when on master while we're running lighthouse with the preview url.

johnboxall:
@jasecode we discussed this IRL, but over time I'd like to see us move away from implicitly deciding which test to run based on the branch we are on and instead rely on the caller to specify the test.

PR #742, opened 2017-06-13
Log requests on the AMP server

olibrook:
Enable request logging for easier debugging. Prompted by the observation that the AMP validator makes x5 duplicate requests for each resource.

## How to test-drive this PR
- Run `npm i`
- Run `npm run dev`
- App should run as normal and log requests in the terminal.

### DIFF
--- a/amp/app/main.js
+++ b/amp/app/main.js
@@ -16,6 +16,7 @@ import path from 'path'
 import Promise from 'bluebird'
 import fetch from 'node-fetch'
 import express from 'express'
+import morgan from 'morgan'
 import React from 'react'
 import ReactDOMServer from 'react-dom/server'
 import _jsdom from 'jsdom'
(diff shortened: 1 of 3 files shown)


amp/app/main.js
```diff
@@ -106,8 +107,12 @@ const homePage = (req, res, next) => {
         .catch(next)
 }
 
+const onLambda = process.env.hasOwnProperty('AWS_LAMBDA_FUNCTION_NAME')
+
 const app = express()
+const logger = morgan(onLambda ? 'short' : 'dev')
 
+app.use(logger)
```
kschmidtdev: For some reason this line causes the validator tests to fail 🤔 (they time out before rendering a page)

olibrook: Urgh, how annoying. Nice catch.

kschmidtdev: Looks like you can wrap the logger setup with `if (process.env.NODE_ENV !== 'test') {`

olibrook: Thanks. Fixed.

PR #743, opened 2017-06-13
Simplify start-test-server

ellenmobify:
## Changes
- Cleans up the `start-test-server.sh` script a bit

## How to test-drive this PR
- Rebuild on CI: 
https://circleci.com/gh/mobify/platform-scaffold/tree/simplify-start-test-server


web/tests/system/start-test-server.sh
```diff
@@ -2,22 +2,19 @@
...
-    npm run prod:build
-    npm run test:server
+    if [ "$CURRENT_BRANCH" != "master" ]; then
+        echo "Running tests against local build."
+        # Kill background processes when this script exits.
+        trap 'kill $(jobs -p)' EXIT
+        echo "Running Test Server."
+        npm run prod:build
+        npm run test:server
+    fi
```
jasecode: I ran into this issue in run-all-tests.sh yesterday. 
If the branch is on master I assume want to do nothing however we should have the echo there instead. Can we move the "else" within the branch check block?

web/tests/system/start-test-server.sh
```diff
@@ -2,22 +2,19 @@
...
 else
     echo "On production branch, test server not needed."
```
jasecode: I ran into this issue in run-all-tests.sh yesterday.
If the branch is on master I assume want to do nothing however we should have the echo there instead. Can we move the "else" within the branch check block?

ellenmobify: Good catch. Let me think about this.

PR #744, opened 2017-06-13
Add <meta charset='utf-8'> tag

jeremywiebe:
Feedback from Sapient/Debenhams via @cdegit was that our PWA's are missing a `<meta charset='utf-8'>` tag (which should be added to avoid XSS attacks).  More info on this page: https://developer.mozilla.org/en/docs/Web/HTML/Element/meta

image

 **JIRA**: N/A
 **Linked PRs**: N/A

## Changes
- As it says in the title.

## Todos

## How to test-drive this PR
- Run `npm run dev`
- Open dev tools and view the *Elements* tab. Verify there is a `<meta>` tag that has the `charset` property set to `'utf-8'`.

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -201,6 +201,11 @@ const attemptToInitializeApp = () => {
         loadAsset('meta', {
             name: 'theme-color',
             content: '#4e439b'
+        })
+
+        loadAsset('meta', {
+            name: 'charset',
+            content: 'UTF-8'
         });
 
         // Attempt to load the worker.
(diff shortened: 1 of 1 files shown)


web/app/loader.js
```diff
@@ -201,6 +201,11 @@ const attemptToInitializeApp = () => {
         loadAsset('meta', {
             name: 'theme-color',
             content: '#4e439b'
+        })
+
+        loadAsset('meta', {
+            name: 'charset',
+            content: 'UTF-8'
```
kschmidtdev: Completely pedantic but the PR description and Mozilla site has `utf-8` in all low-caps - any reason why it's in uppercase here?

jeremywiebe: @kschmidtdev Cause I messed up.

542c9bd1

PR #745, opened 2017-06-13
Group up test dependencies

jasecode:
Grouping up Test Dependencies in attempt to further decrease buildtimes on circleci

## Changes
- Decreased the bottleneck of npm run test:pwa:ci by adding add-self-signed-cert.sh in the circle.yml step and removed building the project and server during this step since we've already done it during setup
- configured circle.yml for more effective backgrounding
- Add wait-for-dependencies for a failsafe to prevent tests from running before backgrounded installations are finished

## Goal
- 5 Minutes of build time on platform-scaffold with all tests enabled.
## Result
- 5:11 ~ from what I see being regular runtime.

## How to test-drive this PR
- rebuild on circleci


web/scripts/wait-for-dependencies.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+echo "Waiting for npm prod:build to complete"
+while [ ! -f build/loader.js ]; do sleep 2; done
+echo "npm prod:build is complete"
+
+echo "Waiting for dev server to become active"
+while ! nc -z localhost 8443; do sleep 2; done
+echo "8443 dev server is now active"
```
ellenmobify: nitpick: we're waiting for the **test** server

web/tests/system/start-test-server.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 echo "Running tests against local build."
```
ellenmobify: I forgot to remove one of these `echo` lines. Could you do that please?

jasecode: addressed.

web/scripts/wait-for-dependencies.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+echo "Waiting for Chrome to finish installing"
+application="Google Chrome"
+while [ "$application" != "$(google-chrome --version | awk '{print $1, $2}')" ]; do sleep 2; done
```
ellenmobify: Does this wait for the _latest_ Chrome to finish installing? I wonder if this will always be true, since the CI container ships with Chrome 54.

jasecode: It's more of a wait with whether or not chrome exists because while we're installing chrome, chrome is not accessible until it finishes causing nightwatch to fail.

ellenmobify:
Future: This PR removed the start/stop test server logic that was in `lighthouse-ci.sh`. Now, `lighthouse-ci.sh` is looking more like `lighthouse-local.sh`. Maybe we can combine these scripts.

web/scripts/wait-for-dependencies.sh
jasecode: ...actually just did some testing. confusingly enough, it seems we're not using chrome 59. Grrr. Need to rethink this if we always want the up to date chrome.

jasecode: Done! We're now waiting for at least Chrome 59.

web/scripts/add-self-signed-cert.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
johnboxall: I really like this change of pulling out this script and isolating it as it has a lot of platform specific logic in it... on thing that isn't clear is you're only going to be able to run this script on ubuntu/linux/ci – so we might want to name it something to that effect.

johnboxall:
🤔 This is a really interesting approach:

1. Install Chrome / Setup Self Signed Certs / Start the Test Server
2. Run lint and/or unit tests
3. Wait for #1 to complete
4. Run e2e and/or performance (lighthouse) tests

This is probably the fastest we're going to get with running things in parallel without revisiting tools we use for testing or the CI environment itself 🐎 

One thing I'm a bit concerned about how is the tradeoff of complexity for speed. 

A few bad ideas:

1. Would it make sense to pack all the "dependancies" into a single script rather than having a # of separate deps steps in the `circle.yml` file?
2. Can you `run-all-tests.sh` locally? Does it make assumptions about `CIRCLE_NODE_TOTAL` being set? What would it look like as two scripts? One for running on multiple nodes and one for running on a single machine?
3. I'm still uneasy about assumptions we make when testing on master branch. Does the `circle.yml` format actually allow us to explicitly launch different commands on different branches? Or even higher level, what are we hoping to get out of running different tests on different branches?
4. I'd love to look closer at the output of this command – what do logs look like when we pass? when we fail? Are we being consistent with our formatting / naming throughout our tests?

web/scripts/add-self-signed-cert.sh
jasecode: Good point. Since these scripts are meant for use on circle-ci only I'll rename it.

jasecode:
@johnboxall 
1) This might be addressed in a different PR, but I had the same idea. The reason for splitting them was for more easy debugging and tracking step by step... need to discuss pros / cons
2) The run-all-tests script is meant for running on circle-ci only... I guess i should rename it as such as well. We still have the regular separate tests and smoke-test for those who want to run the e2e tests locally as written in the readme.
3) The approach we've taken on testing on master branch is to account for them into the shell scripts. There are only two different flows when testing on master that are accounted in the scripts, the circle.yml file does not change.
4) Lets take a look this together our QA meeting?

PR #746, opened 2017-06-13
Use walk to check the build folder for files larger than 2MB

ellenmobify:
Through discussions it was decided that the best way to verify that _minified_ build files are less than 2MB (to get gzipped on the CDN) is to examine the files in the `build` folder. 

## Changes
- Use [walk](https://www.npmjs.com/package/walk) to traverse the `build` folder and make sure each file is less than 2MB

## How to test-drive this PR
- Rebuild on CI https://circleci.com/gh/mobify/platform-scaffold/tree/verify-build-files-size and verify there is no change (all files are currently less than 2MB)
- To test the failure state locally, change the threshold on line 33 and 
```
npm run prod:build
node lighthouse/check-score.js
```


web/lighthouse/check-score.js
```diff
@@ -1,46 +1,80 @@
...
     }
 }
 
+/**
+* Display some important performance metrics.
+*/
 const checkTTI = function(jsonResults) {
-	// Tell me about some performance metrics that are important to me.
     console.log(`Time to interactive: ${jsonResults.audits['time-to-interactive'].displayValue}`)
     console.log(`${jsonResults.audits['total-byte-weight'].displayValue}`)
```
jasecode: This file size is a little confusing... should we print out exactly what file size we're looking at? I guess a simple console.log on the top to say: "Analyzing total file pre-minified / minified size" or something like that would be fine.

web/lighthouse/check-score.js
```diff
@@ -1,46 +1,80 @@
...
-	process.exit(0);
+    console.log('Error Lighthouse report not found.')
+    process.exit(0);
+}
+
+/**
+* Traverse the build folder and verify the bundle size. CDN will not compress 
+* files larger than 2MB.
+*/
+let options = {
```
jasecode: It might be confusing to the web-team if there's no output - I know I was confused when it ran.
I'd suggest we put more console.log like:
"Beginning analyzing bundle size. Each file must be less than It must be less than 2MB to get gzipped on the CDN."
and
"Test Pass."

For them to know it passed.

jasecode:
A future iteration would be probably extracting the bundle analyzer logic from check-score into a test command.
Currently - we can't run check-score.js without a lighthouse report.

johnboxall:
@ellenmobify 

A few late notes here:

Here we're putting two jobs in a single script: we're checking a Lighthouse score, and we're asserting that all files in a particular folder are under a certain size.

They have different requirements: checking a Lighthouse score requires an audit file, checking file size requires the `build` folder.

Since these two things are so different, they deserve different scripts – this should be split into two separate test scripts that.

cc @jasecode

web/lighthouse/check-score.js
```diff
@@ -1,46 +1,87 @@
...
+    listeners: {
+        file: function (root, fileStats, next) {
+            let filePath = path.join(root, fileStats.name)
+            let fileStat = fs.statSync(filePath)
+            if (fileStat.size > 2000000) {
```
johnboxall: This constant should be factored out. The 2MB limit is maximum threshold, but I'd love to shoot even lower.

This should work similarly to `min_lighthouse_score`.

web/lighthouse/check-score.js
```diff
@@ -1,46 +1,87 @@
...
- * Surface some useful information from the HTML and JSON reports generated by Lighthouse
+ * Verify that built files are small enough to be gzipped on the CDN and
+ * surface some useful information from the HTML and JSON reports generated by Lighthouse
  */
 
-
 let fileName
+let failure = false
 
 if (fs.existsSync('lighthouse/audit-local.report.html')) {
```
johnboxall: In some future version of this script, the path to the file should be passed as an argument so its behaviour is explicit rather than implicit.

web/lighthouse/check-score.js
```diff
@@ -1,46 +1,87 @@
...
     fileName = 'audit-local'
 } else if (fs.existsSync('lighthouse/audit-prod.report.html')){
     fileName = 'audit-prod'
 } else {
-	console.log('Error Lighthouse report not found.')
-	process.exit(0);
+    console.log('Error Lighthouse report not found.')
+    process.exit(0);
```
johnboxall: I'm not sure whether linting is currently enabled on this file, otherwise I'd think it would catch the trailing semicolon.

web/lighthouse/check-score.js
johnboxall: The Gzip'd on the CDN part is probably a detail that partners don't need to be aware of. This is a file size test – we're checking the build folder for things over a threshold... anything over that size is too big and you should do something about it! (part of that is b/c it won't be gzip'd but the other part is b/c its criminal to have a 2MB asset)

web/lighthouse/check-score.js
```diff
@@ -1,46 +1,87 @@
...
+            if (failure === false) {
+                console.log(`Success! All build files are below 2MB.`)
+            }
+        }
+    }
+}
+
+if (fs.existsSync('build')) {
+    console.log(`Verifying individual file sizes in the build...`)
+    walk.walkSync('build', options);
```
johnboxall: I'd pull this whole thing out into a function – we're checking that files in a folder are under a certain size threshold... and failing/reporting if any are!

PR #749, opened 2017-06-13
[WEB-1393] Performance Fix: Delay `getCart`

jeffkamo:
In an effort to avoid unnecessary network requests, I have delayed fetching for cart content (i.e. content for the mini cart) until after the user scrolls. There are other actions that trigger fetching of cart content, such as opening the `miniCart`, adding products to the cart, etc.

 **JIRA**: [WEB-1393](https://mobify.atlassian.net/browse/WEB-1393)

## Changes
- Prevent `getCart` from being called on initial page load
- Add window scroll handler to run `getCart`
- Add placeholders into the MiniCart when cart content hasn't been loaded yet

## How to test-drive this PR
- Run `npm run dev`
- Visit any product page and add it to the cart
- Return to the homepage and refresh the page
- After the site reloads, notice that the mini-cart does not get updated to reflect that you have an item in the cart
- Scroll down the page
- **Verify** that after scrolling does the mini cart update

**Bonus**: verify the above steps for Salesforce too!

## Results

| Load Cart only on Scroll | Placeholders in `MiniCart` |
| --- | --- |
| cart-load | cart-load-placeholder |


web/app/containers/mini-cart/container.jsx
```diff
@@ -39,7 +40,7 @@ const MiniCartMain = ({hasItems, closeMiniCart, checkoutShippingURL}) => {
     const buttonClasses = 'c--primary u-width-full u-text-uppercase'
 
     return (
-        <div className="t-mini-cart__content u-flexbox u-direction-column u-padding-md">
+        <div className={MINI_CART_CONTENT_CLASSES}>
```
kevinxh: If the className is a constant, why not just type out the classes? What's the reason to use a variable here?

web/app/containers/cart/container.jsx
```diff
@@ -113,12 +115,17 @@ class Cart extends React.Component {
 Cart.propTypes = {
     cartLoaded: PropTypes.bool,
     hasItems: PropTypes.bool,
-    removeItemID: PropTypes.string
+    removeItemID: PropTypes.string,
+    requestCartContent: PropTypes.func,
```
kevinxh: I noticed that in some of the other files, there are comments for each of the props... Maybe we should do it in all files?
eg.
```javascript
...
/**
  * Whether the cart is finished loading
*/
cartLoaded: PropTypes.bool,
...
```

web/app/store/cart/selectors.js
```diff
@@ -4,12 +4,14 @@
...
-import {createGetSelector, createHasSelector} from 'reselect-immutable-helpers'
+import {createGetSelector} from 'reselect-immutable-helpers'
 import {getCart, getProducts} from '../selectors'
 
 export const getCartCustomContent = createGetSelector(getCart, 'custom')
 
-export const getCartLoaded = createHasSelector(getCart, 'items')
+export const getCartLoaded = createSelector(getCart, (cart) => {
+    return cart.get('items') !== undefined
+})
```
tony-dinh: Curious as to what the difference in the change here is :O?

jeffkamo: I want an empty array to be truthy (it was falsey before) as far as this check is concerned.

Basically, our indication for whether or not the cart is loaded is no longer whether the `items` array is populated, because you can have a fetched cart with no items.

web/app/containers/cart/container.jsx
jeffkamo: This is something we've not been super consistent about. These comments are more important for UI components rather than containers, because these comments are used to generate the Prop tables seen in the docs. We don't generate docs for containers, so these comments aren't technically necessary.

In this case, I was just following the style used in this particular file. We can look to make things consistent across all containers at a later date.

web/app/containers/mini-cart/container.jsx
jeffkamo: It's used in two places that must match styles. If you change one (i.e. the real content container), it should effect the other (i.e. the placeholder container)

web/app/store/cart/selectors.js
tony-dinh: Ah, gotcha :+1:

web/app/containers/mini-cart/container.jsx
kevinxh: Ahh, I didn't see the second one..

PR #753, opened 2017-06-14
AMP-45: main nav

cole-sanderson:
Finalize AMP Main Navigation

 **JIRA**: https://mobify.atlassian.net/browse/AMP-45

## Changes
- Add `<HeaderBar>` to Main Navigation Sidebar
- Update Sign In List Tile to match PWA 

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions](http://localhost:3000/)
- Open Main Navigation and make sure it looks same as PWA Main Navigation


amp/app/components/icon-label-button/index.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+    const classes = classNames('c-icon-label-button', className)
+
+    const attrs = {
+        href, id, on, disabled, name, value, role,
+        className: classes
+    }
+
+    // Add all aria attributes
+    Object.keys(props).forEach((key) => {
```
kschmidtdev: This code is also in `components/button/index.jsx` - can we have it in one place (some sort of utils) and then use that version in both files?

cole-sanderson: It's a little tricky because `components/button/index.jsx` will move to AMP SDK. But if in future we reuse this block in MP project and we can move to utils?

kschmidtdev: Ok - maybe when that happens we can look at putting it into a utils in the SDK, and then they can both use that 😄

olibrook:
Nav is basically working. I need to make a decision on how to handle special cases in the parsed nav items returned by the connector. For merlins that is the `sign-in` link that has different styles and doesn't correspond to a page on the AMP version of the site.

olibrook:
I think this is good to go!

cole-sanderson:
:awyeah: looks good 👍

web/app/components/icon-label-button/index.jsx
```diff
@@ -1,28 +1,24 @@
-/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
-/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
-/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
-
 import React from 'react'
```
olibrook: Wrong file!

PR #754, opened 2017-06-14
AMP-61: ProductItem and ProductTile Components

yourpalsonja:
This PR adds components that are needed for the _PLP_ in **AMP-61**, it isn't the build-out of the _PLP_ itself.

**JIRA**: https://mobify.atlassian.net/browse/AMP-61

## Changes
- Added in `ProductItem` component, styles, test and README
- Added in `ProductTile` component, styles, test and README
  - Removed skeleton since we won't be needing it

## Todos
- [ ] Engi + 1

## How to test-drive this PR
- Run `npm run dev`
- Visit `localhost:3000`
- I added this code to a template to ensure that it worked, along with styles pulled in: `<ProductTile className="hi" href="/" price="$200" title="product" thumbnail={{alt: "test", src: "/static/mobify.png"}}/>`

### DIFF
--- a/amp/app/components/product-item/index.jsx
+++ b/amp/app/components/product-item/index.jsx
@@ -12,8 +12,7 @@ const ProductItem = ({
     className,
     image,
     price,
-    title,
-    customWidth
+    title
 }) => {
     const classes = classNames('amp-product-item', 'u-flexbox', 'u-direction-row-reverse', className)
 
@@ -43,7 +42,7 @@ const ProductItem = ({
 
 
             {image &&
-                <div className="u-padding-end u-flex-none" style={{width: customWidth}}>
+                <div className="u-padding-end u-flex-none">
                     {image}
                 </div>
             }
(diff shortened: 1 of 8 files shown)


amp/app/components/product-item/index.jsx
```diff
@@ -0,0 +1,92 @@
...
+                {!!children &&
+                    <div>
+                        {children}
+                    </div>
+                }
+            </div>
+
+
+            {image &&
+                <div className="u-padding-end u-flex-none" style={{width: customWidth}}>
```
kschmidtdev: This isn't allowed in AMP:
```The inline 'style' attribute is not allowed in AMP documents. Use 'style amp-custom' tag instead. ```

yourpalsonja: Oops, good catch! Should this be something that gets checked in the amp validator, @kschmidtdev? I thought I ran `npm run test:all`, but perhaps I didn't.

kschmidtdev: @yourpalsonja The validator only checks our existing templates - so the component would have to be used within one of them to trip that. I did that manually and then noticed the browser extension validator complained 😬

yourpalsonja: Okay, cool! Thanks! I've removed this `style` and the `customWidth` prop since they aren't of any use in this component for AMP.

PR #756, opened 2017-06-14
Move search into IM & implement across connectors

Cdok:
# Search -> Integration Manager

## Related SDK PR: https://github.com/mobify/progressive-web-sdk/pull/714

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-73

## Changes

 - Merlin's search connector is migrated to use Integration Manager
 - The SFCC connector is expanded to add the required search OCAPI calls for executing and displaying a Search Results Page and sorting a Search Results Page
 - Does not include filtering

## How to test-drive this PR
```
cd progressive-web-sdk
git checkout im-search-commands
npm link
npm install
npm run dev:build
cd ../platform-scaffold/web
git checkout sfcc-search-suggestions
npm link progressive-web-sdk
npm run dev
```
- Test both SFCC connector & Merlin's Connector
- Search for `Shirt` on SFCC or `potion` on merlins
- Click through suggestions & results


Cdok:
I'm getting pretty bad latency when I'm typing a search suggestion - might need to confirm if SFCC just doesn't like me today or if this is a real problem

web/app/integration-manager/commands.js
```diff
@@ -34,3 +34,5 @@ export const submitNewsletter = (...args) => connector.submitNewsletter(...args)
 
 /** @function */
 export const getSearchSuggestions = (...args) => connector.getSearchSuggestions(...args)
+
+export const submitSearch = (searchQuery) => connector.submitSearch(searchQuery)
```
jeremywiebe: Can you add JSDoc comments to this?

web/app/integration-manager/_sfcc-connector/commands.js
```diff
@@ -10,12 +10,36 @@ import * as appCommands from './app/commands'
...
 export default {
     home: homeCommands,
     products: productsCommands,
     categories: categoriesCommands,
     cart: cartCommands,
     app: appCommands,
     checkout: checkoutCommands,
-    account: accountCommands
+    account: accountCommands,
+    getSearchSuggestions,
```
Cdok: I was thinking about this myself. Since search is accessible in any part of the application, would it make sense to keep a `search` key under `app` to have the search information?

web/app/integration-manager/_merlins-connector/commands.js
```diff
@@ -34,6 +34,10 @@ export const getSearchSuggestions = (query) => (dispatch) => {
         .then((responseJSON) => dispatch(receiveSearchSuggestions(parseSearchSuggestions(responseJSON))))
 }
 
+export const submitSearch = (query) => (dispatch) => {
```
Cdok: I've updated it to `searchProducts` (on the off-chance that there's another command in the future to search for something _other_ than products).

Let me know if this makes sense 😄

web/app/integration-manager/_sfcc-connector/commands.js
```diff
@@ -10,12 +10,36 @@ import * as appCommands from './app/commands'
...
+import {buildQueryString} from '../../utils/utils'
+import {browserHistory} from 'progressive-web-sdk/dist/routing'
+import {SUGGESTION_URL} from './config'
+
+export const getSearchSuggestions = (query) => (dispatch) => {
+    // SFCC API requires min length of 3
+    if (query.length < 3) {
+        return dispatch(receiveSearchSuggestions(null))
+    }
+    const queryURL = `/search_suggestion?q=${query}`
```
Cdok: I wasn't sure, most of the `makeApiRequest` calls in the SFCC commands are passing in hardcoded URL strings (ex. `removeFromCart` is being passed `/baskets/${basket.basket_id}/items/${itemId}`). 


Do you think we should be moving these types of URL strings into `config.js`, or can we assume that these endpoints will always be the same?

web/app/integration-manager/_sfcc-connector/categories/commands.js
```diff
@@ -42,10 +44,30 @@ const extractCategoryId = (url) => {
...
+
+    } else {
+        const categoryID = extractCategoryId(url)
+        searchUrl = makeCategorySearchURL(categoryID, '')
+        dispatch(fetchCategoryInfo(categoryID))
+    }
 
-    return dispatch(fetchCategoryInfo(categoryID))
-        .then(() => makeApiRequest(makeCategorySearchURL(categoryID), {method: 'GET'}))
+    return makeApiRequest(searchUrl, {method: 'GET'})
```
DanielBrookRoberge: Are we OK to not wait for `fetchCategoryInfo` above before making this request?

web/app/integration-manager/_sfcc-connector/commands.js
jeremywiebe: I think that makes sense (move them into the `app` key)

web/app/integration-manager/_sfcc-connector/categories/commands.js
jeremywiebe: I feel like we should keep the behaviour the same as before for now just to be safe..... so no.

web/app/integration-manager/_sfcc-connector/commands.js
jeremywiebe: Ok, I noticed `SUGGESTION_URL` was, but I guess that's because it's re-used in `initProductListPage`.  Don't bother with moving this to config... I'd say if the URLs change in a new version we probably only have one place to update it anyways.

web/app/integration-manager/_merlins-connector/parsers.js
```diff
@@ -1,23 +1,3 @@
 /* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
 /* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
 /* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
```
jnurse: This file can probably be deleted now.

web/app/integration-manager/_sfcc-connector/categories/commands.js
```diff
@@ -8,15 +8,16 @@ import {receiveCategoryContents, receiveCategoryInformation} from '../../categor
 import {receiveProductListProductData} from '../../products/results'
 import {parseProductListData} from '../parsers'
 
-import {getCategoryPath} from '../config'
+import {getCategoryPath, SUGGESTION_URL} from '../config'
 
 const makeCategoryURL = (id) => `/categories/${id}`
-const makeCategorySearchURL = (id) => `/product_search?expand=availability,images,prices&q=&refine_1=cgid=${id}`
+const makeCategorySearchURL = (id, query) => `/product_search?expand=availability,images,prices&q=${query}&refine_1=cgid=${id}`
```
jnurse: What about making query default to '' here, then we don't have to actually pass an empty string when there's no query value.

web/app/integration-manager/_sfcc-connector/parsers.js
```diff
@@ -158,3 +158,23 @@ export const parseProductListData = (products) => {
     })
     return productListData
 }
+
+export const parseSearchSuggestions = ({product_suggestions}) => {
```
jnurse: You could try this instead: 
`export const parseSearchSuggestions = ({product_suggestions: {products}}) => {`

Then you can remove line 163

web/app/integration-manager/app/commands.js
```diff
@@ -8,6 +8,17 @@ export const register = (commands) => {
     connector = commands
 }
 
+/** @function */
```
jnurse: Can you add a description of the function and the params to this JSDoc comment?

web/app/integration-manager/_merlins-connector/app/commands.js
```diff
@@ -67,7 +67,7 @@ export const getSearchSuggestions = (query) => (dispatch) => {
         return dispatch(receiveSearchSuggestions(null))
     }
 
-    const queryURL = `${QUERY_URL}${buildQueryString(query)}&_=${Date.now()}`
+    const queryURL = `${buildQueryURL(query)}&_=${Date.now()}`
```
jnurse: Can this `Date.now` piece go into the buildQueryURL function too?

PR #761, opened 2017-06-15
[WEB-1392] Delay Default Ask to second page view

jeffkamo:
In an attempt to ~~reduce the number network calls (faster to downloading everything) and~~ avoid extraneous renders (faster visually complete time), we've delayed the default ask to only show on the second page view.

 **JIRA**: [WEB-1392](https://mobify.atlassian.net/browse/WEB-1392)

## Changes
- Change `showOnPageCount` value from `1` to `2`

## How to test-drive this PR
- Run `npm run dev`
- Refresh the page
- *Verify* that the second page load caused the DefaultAsk to show


tony-dinh:
Works as advertised but I believe the performance improvement here is just to reduce the amount of things rendered on first view. I think the network requests are mandatory in order to check how many page views we have, so not sure if there is any reduction there :O.

Changes look good to me, and works! No complaints :+1:

jeffkamo:
@tony-dinh I think you're right. Looks like we download the push assets regardless if the ask is shown or not. Description updated.

PR #762, opened 2017-06-15
AMP: Font changes

cole-sanderson:
Font Changes

## Changes
- Add Google fonts
- `wf-active` class same as PWA
- remove `amp-border-box` class, no longer need it

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions](http://localhost:3000/)
- Check if Oswald font is working

### DIFF
--- a/amp/app/templates/amp-page.js
+++ b/amp/app/templates/amp-page.js
@@ -2,7 +2,7 @@ const ampPage = ({title, canonicalURL, body, css, ampScriptIncludes}) => (
     /*eslint-disable */
     `
     <!doctype html>
-    <html amp lang="en" class="amp-border-box">
+    <html amp lang="en" class="wf-active">
         <head>
             <!-- Standard AMP markup -->
             <meta charset="utf-8">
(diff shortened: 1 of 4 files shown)


cole-sanderson:
In PWA they added font with `componentDidMount` and we cannot use it. 

```
componentDidMount() {
        this.hidePreloaderWhenCSSIsLoaded()
        this.props.fetchSvgSprite()
        this.props.initApp()
        WebFont.load({
            google: {
                families: ['Oswald:200,400']
            }
        })

        // Lazy load other containers when browser is at the end of frame
        // to prevent jank
        registerPreloadCallbacks()
    }
```

amp/app/templates/amp-page.js
```diff
@@ -2,14 +2,15 @@ const ampPage = ({title, canonicalURL, body, css, ampScriptIncludes}) => (
     /*eslint-disable */
     `
     <!doctype html>
-    <html amp lang="en" class="amp-border-box">
+    <html amp lang="en" class="wf-active">
```
cole-sanderson: `wf-active` class is from PWA, on font load `wf-active` will be added to `body`.

amp/app/templates/amp-page.js
```diff
@@ -2,14 +2,15 @@ const ampPage = ({title, canonicalURL, body, css, ampScriptIncludes}) => (
...
         <head>
             <!-- Standard AMP markup -->
             <meta charset="utf-8">
             <script async src="https://cdn.ampproject.org/v0.js"></script>
             <title>${title}</title>
             <link rel="canonical" href="${canonicalURL}" />
             <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1, user-scalable=no">
+            <link href="https://fonts.googleapis.com/css?family=Oswald:200,400" rel="stylesheet">
```
kschmidtdev: Technically would could use export the `App` container with `ampSDK.ampComponent`, and to inject this line via that system. I'm not sure if it's worth it in this case, but personally I think we should try to keep partners from messing around in this file. What do you think?

cole-sanderson: Yeah, that's great idea 👍  I ll try to fix that. Thanks

olibrook: @kschmidtdev @cole-sanderson `ampComponent` is really for when it is not known what is going to be rendered. In this case, the app *is* always going to be rendered - this is a reasonable global as it is, I think.

olibrook:
I see Oswald, looks good to me.

👍

amp/app/templates/amp-page.js
kschmidtdev: Yes - I think we're just going to stick the font tag in another file and bring it into the template, to minimize changes to the template.

amp/app/templates/amp-page.js
```diff
@@ -1,14 +1,18 @@
...
-    <html amp lang="en" class="amp-border-box">
+    <html amp lang="en" class="wf-active">
         <head>
             <!-- Standard AMP markup -->
             <meta charset="utf-8">
             <script async src="https://cdn.ampproject.org/v0.js"></script>
             <title>${title}</title>
             <link rel="canonical" href="${canonicalURL}" />
+
+            <!-- Font Serve: Insert <link rel="styleshet" href="/path/to/font-serve"> in <head> tag. It will only work with Typography.com, Fonts.com, Google Fonts, or Font Awesome. For more information, please read: https://www.ampproject.org/docs/guides/responsive/custom_fonts. If you want to use custom font then please use <AmpFont> component. @TODO add/fix component name when we add it. -->
```
kschmidtdev: What do you think about this comment being in `font-serve.js` instead 🤔  I'm not sure we strip these comments out of the eventual markup so all shoppers will be downloading it 😆

cole-sanderson: good call, moved comments in font-serve file.

PR #764, opened 2017-06-15
Remove dependency code patching for integration manager usage

kschmidtdev:
Make our own minimal reducers or simply stop using reducers that we don't need, in order to not end up importing code that relies on `window` and other web browser facilities to exist. 

## Changes
- Add our own `product-details` reducer
- Stop importing the PWA's `header` reducer
- Remove code rewriting config from webpack config

## How to test-drive this PR
- Run `npm run dev`
- Check out http://localhost:3000, http://localhost:3000/eye-of-newt.html, http://localhost:3000/potions.html

### DIFF
--- a/amp/app/data-integration/connectedStore.js
+++ b/amp/app/data-integration/connectedStore.js
@@ -15,11 +15,10 @@ import {CURRENT_URL} from '../../../web/app/containers/app/constants'
 
 import appReducer from './app-reducer'
 import footerReducer from '../../../web/app/containers/footer/reducer'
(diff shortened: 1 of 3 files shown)


olibrook:
Great! Works. I'm really glad to see the back of those patches.

Any hope of getting something like this into the IM, or is it very AMP-specific?

👍

kschmidtdev:
@olibrook We'll have to see what the reducers that end up the SDK look like - this technically isn't AMP-specific, just a subset of the PWA.

PR #766, opened 2017-06-15
Fix selector names referencing productPaths

Cdok:
# Fix selector names referencing productPaths

## Changes
- Rename productPath selectors missed from this PR https://github.com/mobify/platform-scaffold/pull/711

### DIFF
--- a/web/app/store/categories/selectors.js
+++ b/web/app/store/categories/selectors.js
@@ -17,7 +17,7 @@ export const getSelectedCategory = createGetSelector(
 
 const PLACEHOLDER_URLS = Immutable.List(new Array(5).fill(PLACEHOLDER))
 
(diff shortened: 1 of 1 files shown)


web/app/store/categories/selectors.js
```diff
@@ -47,6 +47,6 @@ export const getCategoryParentHref = createGetSelector(
 
 export const getCategoryProducts = createSelector(
     getProducts,
-    getCategoryProductPaths,
-    (products, productUrls) => productUrls.map((path) => products.get(path))
+    getCategoryProductIds,
+    (products, productIds) => productIds.map((path) => products.get(path))
```
jnurse: One more! The path variable in the map should probably be id instead.

PR #770, opened 2017-06-15
Accessibility Improvement: Modal Manager

kevinxh:
Problem: when a modal is open, user (particularly Screen Reader users) should only be able to interact with the modal, not the background. Currently, the user can still focus on text, buttons and everything else that is in the background by using accessibility features such as screen readers. 

In this PR, a new component is introduced : `ModalManager`, it's modelled after the `NotificationManager` component in the scaffold. `ModalManager` helps to extract all modals that are nested deep in the component hierarchy, and it is placed as a sibling to `<div class="app-wrap">`. When a modal is open, the attribute `aria-hidden=true` will be added to the app wrap, thus hide the background from accessibility tools.

Document for discussion: https://docs.google.com/document/d/17GvCBYp7KmF_2EIH5W14Qoh-dbropTxwZgAkP6kL9_8/edit

 JIRA: **[WEB-1404](https://mobify.atlassian.net/browse/WEB-1404)**

State: **ready for review**

## Changes
- Added a new component `ModalManager`
- The render of modals are controlled by the `modal` store in redux, except `<Navigation>` and `<OfflineModal>` which are always rendered.
- Modals are moved from `web/app/containers/...`to `web/app/modals`
- Modals have new class prefix: `m-...`

## Todos
- [x] Add accessibility attributes to the background when modal is open.
- [ ] Update Generator
- [ ] Docs

## How to test-drive this PR
- Run `npm run dev`
- Test modals~!


web/app/components/modal-manager/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+import CartWishlistModal from '../../containers/cart/partials/cart-wishlist'
+import CartRemoveItemModal from '../../containers/cart/partials/cart-remove-item'
+import CheckoutConfirmationModal from '../../containers/checkout-confirmation/partials/checkout-confirmation-modal'
+import RememberMeModal from '../../containers/login/partials/remember-me-modal'
+import OfflineModal from '../../containers/offline/partials/offline-modal'
+import ProductListFilterModal from '../../containers/product-list/partials/product-list-filter-modal'
+
+const modals = {
+    // [MODAL.NAVIGATION_MODAL]: <Navigation />
+    // [MODAL.OFFLINE_MODAL]: <OfflineModal />
```
jeffkamo: I'm not particularly a fan that some modals need to be treated different.

Are we still exploring the possibilit of making these modals connected and getting their required props via the Redux state?

web/app/components/modal-manager/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+    const {history, reload, isOpen} = props
+    let openedModal
+    for (const modal in isOpen) {
+        if (isOpen[modal]) {
+            openedModal = modals[modal]
+        }
+    }
+
+    return (
+        <div>
```
jeffkamo: This component, as a local component, should have its wrapper class be `c-modal-manager`.

jeffkamo: That said, if we move this into it's own directory, like `app/modals` then I'd suggest we adopt a new prefix `m-`

web/app/containers/app/container.jsx
```diff
@@ -62,6 +62,12 @@ class App extends React.Component {
         registerPreloadCallbacks()
     }
 
+    getChildContext() {
```
kevinxh: @jeffkamo I'm not too sure about this, but this is the neatest way to pass the `reload` handler from `App` container to its children components..

web/app/containers/offline/container.jsx
```diff
@@ -30,12 +29,11 @@ const Offline = ({
...
 }
 
-
-Offline.propTypes = {
+Offline.contextTypes = {
     /**
      * Method that attempts to fetch the page again
      */
-    reload: PropTypes.func.isRequired
+    reload: PropTypes.func
```
jeffkamo: If we stick with the approach of using context, perhaps we should clarify in this file where the reload is coming from since it's no longer being explicitly passed into it (i.e. backtracing might be harder now, maybe?)

web/app/modals/offline/container.jsx
```diff
@@ -38,16 +37,21 @@ const OfflineModal = ({
...
-    isOpen: PropTypes.bool.isRequired,
-    reload: PropTypes.func.isRequired
+    isOpen: PropTypes.bool.isRequired
+}
+
+OfflineModal.contextTypes = {
+    /**
+     * Method that attempts to fetch the page again
+     */
+    reload: PropTypes.func
```
jeffkamo: Same here, as mentioned above, we should indicate where this context is coming from.

web/app/containers/app/container.jsx
jeffkamo: I'm not aware of the general best practices around exposing context in this way. Perhaps to be a bit defensive, let's include a comment here to clarify why `reload` is being exposed in this way (i.e. so it can be used in the Offline modal)

PR #772, opened 2017-06-16
Add amp development guidelines as discussed today

olibrook:
## How to test-drive this PR
- Read the readme


kschmidtdev:
I think the main README is considered partner-facing documentation - is this correct @drewzboto ? I'm thinking we put this in a DEVELOPING.md that we strip out, or maybe in the mobify-amp-sdk repo instead? What do you think?

olibrook:
Fixed @kschmidtdev

amp/DEVELOPING.md
```diff
@@ -0,0 +1,59 @@
+# Mobify AMP SDK Scaffold - Developer Readme
+
+
+## Development Guidelines
+
+We had a bit of back-and-forth about how AMP relates to the PWA and
+decided on these guidelines.
```
jeffkamo: This may be true, but is maybe not all that relevant of information in the long run. Perhaps a brief explanation on why this process might be important instead?

amp/DEVELOPING.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Naming conventions
+
+The UI Kit is our spec for building mobile sites. We will follow its naming
```
jeffkamo: Perhaps add a link to the UI Kit repo?

olibrook:
Fixed. I don't think we are shooting for 100% quality guidelines at this point. We just want to keep people roughly on the same page.

PR #777, opened 2017-06-16
[WEB-1395]: Prefetch Template Chunks

tony-dinh:
Previously, our code-split chunks were fetched AND loaded.

We now prefetch them instead, so that they are not loaded until they are needed.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1395
 **Linked PRs**: N/A

## Changes
- Prefetch instead of preload code-split chunks

## Todos
- [x] Test changes on iOS and Android Devices
- [ ] +1

## How to test-drive this PR
- Run `npm run dev`
- Before navigating... to ensure that the coded chunks were loaded correctly use the chrome dev tools:
    1. Check the network tab to see the requests for template files: 
        - `cart.js`
        - `checkout-confirmation.js`
        - `checkout-payment.js`
        - `checkout-shipping.js`
        - `login.js`
        - `product-details.js`
        - `product-list.js`
    2. Check that they are all cached `Application` > `Cache` > `progressive-web-scaffold-<version>-bundle`
- Cut the network connection.
- Navigate to a PLP, and it should still fail gracefully.

## Results
Webpage tests result in a more consistent measurement of TTI ~ 9.9 - 10.5.
https://www.webpagetest.org/result/170619_WM_1PFE/

The fetched code chunks no longer execute when fetched from the network.
Turns out something else is blocking the main thread so we are still in the 10s TTI range because of it.


appveyor.yml
```diff
@@ -11,7 +11,6 @@ branches:
   only:
     - master
     - develop
-    - hotfix-tests-on-windows
```
tony-dinh: Just removing something that slipped into develop.

web/app/containers/templates.js
```diff
@@ -6,23 +6,34 @@
...
+// A list of filenames that we want to be prefetched by the browser
+// these filenames come from the chunk names defined below.
+const prefetchFilenames = [
+    'cart.js',
+    'checkout-confirmation.js',
+    'checkout-payment.js',
+    'checkout-shipping.js',
+    'login.js',
+    'product-details.js',
+    'product-list.js'
```
jansepar: Can we do this without hardcoding names?

web/app/containers/templates.js
```diff
@@ -6,23 +6,34 @@
...
         loader,
         LoadingComponent: ContainerPlaceholder
     })
-    loadableList.push(loadable)
     return loadable
 }
 
-export const registerPreloadCallbacks = () => {
-    loadableList.forEach((loadable) => {
-        requestIdleCallback(() => loadable.preload())
```
jansepar: We still probably want to do this the old way for iOS since it doesn't support prefetching.

tony-dinh: Dammit Safari... the one browser.

web/app/containers/templates.js
tony-dinh: Hmm.. if you have any suggestions that would be great.

The named chunks comment don't accept variables unfortunately...
So i can't really think of a way where we can avoid hardcoding them...

But perhaps we can make the PWALoadable method take in the chunk file name method as a parameter and so this array can be built as we need.

jansepar: Could they be templated in at compile time? Would love to not have repetition, it would be so easy for someone to update it in the chunks section and not here.

tony-dinh: Do you mean something like... defining those template names here:
https://github.com/mobify/platform-scaffold/blob/develop/web/webpack/base.loader.js#L64

and have that replaced during build time?
Those constants don't have an effect on comments/comment blocks

jansepar: Yup something like that.

tony-dinh: Since comment blocks such as `/* webpackChunkName */` don't get filled in at build time, it would mean we still need to manually define them in the comment block and so... we would still end up having to edit them in two separate places.

Hm any other ideas? For now, I've added `chunkFilename` as a parameter to the PWALoader method so that at least the array is built during execution and if a template is removed, then it is also removed from the array.

jansepar:
👍

PR #779, opened 2017-06-16
Group Test dependencies into one file, Clean up circle.yml

jasecode:
Our circle.yml increased in complexity and has a lot of commands. Wanted to group all our installation steps together and clean up the circle.yml file overall.

## Changes
- Cleaning up our Circle.yml
- add setup-ci-env.sh 
-- Script sets up all our dependencies
- update wait-for-dependencies.sh
-- Added logic of a timeout for the installation steps and print the log
- combined update chrome and accepting the cert to prevent apt-get conflicts (which causes one of the two (mainly installing chrome to fail)
- add upload-artifacts.ci.sh

## How to test-drive this PR
- Rebuild on circleci


web/scripts/update-chrome-accept-cert-ci.sh
```diff
@@ -1,7 +1,18 @@
 # Install the latest stable version of Chrome.
 # https://circleci.com/docs/build-image-trusty/#google-chrome
```
ellenmobify: Comment here why we had to combine the Chrome installation and cert scripts.

web/lighthouse/check-score.js
```diff
@@ -4,15 +4,16 @@ const fs = require('fs')
 const chalk = require('chalk')
 
 /**
- * Surface some useful information from the HTML and JSON reports generated by Lighthouse
+ * Verify that built files are small enough to be gzipped on the CDN and
```
ellenmobify: Can update this comment as this script no longer does `Verify that built files are small enough to be gzipped on the CDN`

ellenmobify:
👍

web/scripts/setup-env-ci.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+#Install all dependencies on circleci
+
+mkdir logs
+printf "\nInstall Chrome: "
+nohup bash -c "./scripts/update-chrome-accept-cert-ci.sh &> logs/apt-get.log &"
```
kschmidtdev: ✨

PR #781, opened 2017-06-16
Add the option to pick your connector from the CLI

Cdok:
# Pick your connector with a command line flag 🇨🇦 !

This PR is currently a proof-of-concept of passing command line args to pick your connector, thus removing the ambiguity of how exactly connectors work, as well as making the connector selection substantially easier to document / explain to someone new to the platform.

## Changes
- Add a global webpack compile-time variable `WEBPACK_CONNECTOR_NAME` to select your connector
- By default, `mp` and `sfcc` are supported for Merlin's and SFCC respectively

### Test SFCC

```bash
git checkout connector-selection-from-flag
npm start sfcc
```

### Test Merlin's

```bash
npm start mp
```

### Test that Merlin's is the default

```bash
npm start
```

## ... but what about `npm run dev`

Fear not, `npm start` does exactly what `npm run dev` does!

However, this PR might be a good time to start the transition over to `npm start` 😈 

## Idea 💡 

Wouldn't it be nice to run
```bash
npm start mp
```
in one terminal and
```bash
npm start sfcc
```
in another?

Now you can :tada:

Having simultaneous tabs to test across both connectors prevents you from having to play around with that pesky `main.jsx`.

screen shot 2017-06-16 at 3 51 57 pm

> Call now and we'll include a second connector free of charge!

## Side effects include:

- The connector-specific ports will break everything that points at `localhost:8443`


Cdok:
```javascript
let connectorPort
 
if (process.argv.find((arg) => arg === 'sfcc')) {
    connectorPort = 8080
} else if (process.argv.find((arg) => arg === 'mp')) {
    connectorPort = 8443
} else {
    connectorPort = 8443
}
```
Note the above code is mostly repeated for picking the connector init as well, if this feature ends up being something that we're interested in I could DRY these up and make them less brute force-y.

There's also some potential here to provide different preview links with the correct port / URL per connector.

web/app/main.jsx
```diff
@@ -26,12 +26,14 @@ import Stylesheet from './stylesheet.scss' // eslint-disable-line no-unused-vars
...
 // DO NOT USE! Merlins Connector is an example connector that is for demo only
-import initConnector from './init-merlins-connector'
+const initConnector = require(WEBPACK_CONNECTOR_NAME).default // eslint-disable-line
+// import initConnector from './init-merlins-connector'
 // import initConnector from './init-sfcc-connector'
 // import initConnector from './init-stub-connector'
 
 initConnector()
 
+
```
jeremywiebe: 🤔

web/dev-server/index.js
```diff
@@ -9,8 +9,18 @@ const config = require('../webpack/dev.js')
 const addMiddleware = require('./middlewares/frontendMiddleware')
 const logger = require('./logger')
 
+let connectorPort
+
+if (process.argv.find((arg) => arg === 'sfcc')) {
+    connectorPort = 8080
```
jeremywiebe: Why does SFCC listen on 8080?

web/webpack/base.main.js
```diff
@@ -16,6 +16,16 @@ const webPackageJson = require('../package.json')   // eslint-disable-line impor
 
 const analyzeBundle = process.env.MOBIFY_ANALYZE === 'true'
 
+let connector
+
+if (process.argv.find((arg) => arg === 'sfcc')) {
```
jeremywiebe: I'm debating in my head if this needs a `--connector` option before it. 
```
# this?
npm start --connector sfcc

# of this?
npm start sfcc
```

Cdok: I was thinking about both options. I feel like it would be a bit _safer_ if we had a `--connector` flag, but I do like how concise `npm start mp/sfcc` is

web/dev-server/index.js
Cdok: No specific reason haha - just used `8080` as an example, but is there another standard HTTPS port we could use other than `8443`?

jeremywiebe:
Talked with @jasecode about this idea. We're going to opt for a lower-impact solution to the scaffold code itself by building a small shell script that just modifies `main.jsx` to comment out the unwanted connectors and enable the one we want.  

Eg
```sh
$ pwd
.../platform-scaffold/web

$ cat main.jsx
...
import initConnector from './init-merlins-connector'
// import initConnector from './init-sfcc-connector'
// import initConnector from './init-stub-connector'

initConnector()
...

$ scripts/enable-connector sfcc
Updating main.jsx to enable sfcc-connector

$ cat main.jsx
...
// import initConnector from './init-merlins-connector'
import initConnector from './init-sfcc-connector'
// import initConnector from './init-stub-connector'

initConnector()
...
```

Going to close this for now and @jasecode will open a new PR for the utility and supporting npm scripts (if needed)

PR #788, opened 2017-06-19
WIP: Reorganize test folders and files

ellenmobify:
* Move test scripts under `tests/system/test-scripts`
* Move `lighthouse` folder under `tests/system/test-scripts`


web/tests/system/verify-build-size.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env node
```
ellenmobify: Can we remove this file?

jasecode: My bad. Removing...

ellenmobify:
👍

PR #792, opened 2017-06-19
`export default customCommands` so importing is more natural

jeremywiebe:
Initial feedback from Debenhams is that using custom commands from the IM feels different than built-in commands. 

## Before

The following example contrasts importing built-in commands and custom commands:
```js
// Built-in
import {login} from 'progressive-web-sdk/dist/integration-manager/account/commands'
dispatch(login('jeremy@mobify.com', 'supersecretpassword', true))

// Custom
import {customCommands} from 'progressive-web-sdk/dist/integration-manager/custom/commands'
return dispatch(customCommands.isEmailAvailableCommand(formValues.username))
```

## After
This PR changes the export of custom commands slightly so that importing them is more like the built-in commands.

```js
// Built-in (unchanged)
import {login} from 'progressive-web-sdk/dist/integration-manager/account/commands'
dispatch(login('jeremy@mobify.com', 'supersecretpassword', true))

// Custom (this works)
import {isEmailAvailableCommand} from 'progressive-web-sdk/dist/integration-manager/custom/commands'
return dispatch(isEmailAvailableCommand(formValues.username))

// Custom (this also works)
import * as customCommands from 'progressive-web-sdk/dist/integration-manager/custom/commands'
return dispatch(customCommands.isEmailAvailableCommand(formValues.username))
```

## Missing Custom Command

If the custom command is _not_ registered, it is still possible to check if the custom command is defined and skip calling it. A bonus is that the console warns that the command is also undefined which is a good signal that the UI is dispatching a custom command that no longer exists.

```js
import {isEmailAvailableCommand} from '../../integration-manager/custom/commands'
if (isEmailAvailableCommand) {
    return dispatch(isEmailAvailableCommand(formValues.username))
}
```

image

 **JIRA**: N/A
 **Linked PRs**: N/A

## Changes
- Export the `customCommands` as the default export instead of being a named export
- Update the single custom command usage to work with new export model

## Todos

## How to test-drive this PR
- Run `npm run dev`
- Add an item to the cart and start checkout
- Enter an email address in the Email Confirmation field and verify that a network call is made to the `isEmailAvailable` endpoint


jeremywiebe:
Upon further investigation this won't be possible the way things are architected. The integration manager is imported as the main bundle loads but only after it loads is the connector extension registered. For many uses, this is too late as objects that import the IM commands have already been evaluated. 

For now we have to settle with the approach that's a bit different than regular IM commands. 

**Closing this PR for now.**

PR #793, opened 2017-06-20
Automatically include the right CSS for components used on each page

olibrook:
Open for feedback.


olibrook:
I'm going to leave separate comments for things I think are worth thinking about...

olibrook:
These changes compile the CSS for each component's `base.scss` and `theme.scss` separately. That's not much of a problem, but it means the SASS compiler can't optimize cases like this:

```
// base.scss
   .component {color: green}
// theme.scss
   .component {color: blue}
// config
  styles.set(Component, [base.scss, theme.scss])
```

That would output two style rules, one overriding the other. If we want to fix it and let the sass compiler optimize, we'd have to change how we write the stylesheets, eg:

```
// base.scss
   .component {color: green}
// theme.scss
    import base.scss
   .component {color: blue}
// config
  styles.set(Component, [theme.scss])
```

The second option would reduce redundant CSS rules, but reduce copy-pastability from the PWA.

olibrook:
If you look carefully, you'll notice that you have to import dependent sass files explicitly in each component's file. I don't think this is a problem, but I'm looking for feedback. Eg.

```
@import '../../styles/variables';

// ProductItem
// ===

.amp-product-item__category {
    font-size: $font-size - 2;
    line-height: 1em;
}

```

olibrook:
The `config/styles.js` file is a pain in the butt, but it allows us to override styles per-project fairly easily. The other option I thought about was this:


```JavaScript
// Each time a user wants to override styles, etc they could wrap the existing component
// in the SDK and override a styles property, like this:

// in lancome-button/index.js
import Button from 'amp-sdk/button'
import ampComponent from 'amp-sdk'
import base from 'amp-sdk/button/_base.scss'
import customTheme from './_theme.scss'

const LancomeButton = ampComponent(Button)
LancomeButton.styles = [base, theme]
export default LancomeButton

```

Is that easier to understand?

cole-sanderson:
@olibrook 
I don't see problem of including base and theme styles as long as we import in proper order to ensure that theme styles override base styles. 

I am okay with import variables in component file but I would suggest to create new file that includes variables, lib and animation and include to component file instead of just variable. 

Yeah, I think the `config/style.js` is pretty pain in the butt and I like your option on import base and theme style IN component rather than `config/style.js` file.

amp/app/config/scripts.js
```diff
@@ -0,0 +1,25 @@
+import AmpForm from '../components/amp-form/index'
+import AmpLightbox from '../components/amp-lightbox/index'
+import Analytics from '../components/analytics/index'
+import Sheet from '../components/sheet/index'
+
+
+const config = new Map()
```
kschmidtdev: I'm not clear what the advantage is to moving to this design for our script includes. I thought the other one was pretty good, and it was clear how it would work with a combination of components from the SDK as well as within the project (scaffold in our case). How does this work if you bring in components used in the SDK? You have to know to add the right data to your particular project?

olibrook: @kschmidtdev The CSS included per component needs to be modifiable from project-to-project. The `Map()` allows us to track renders and make associations between Component->Styles that can be changed between projects.

That said, I think I've come up with a way to do this that will be usable for CSS and scripts that will work much the same way as the original.

yourpalsonja:
I don't have strong opinions one way or another on any of this.

The only reason why I would be for importing `base.scss` into the `theme` file might be to potentially reduce the overall CSS file size by avoiding redundant styles, but I doubt it will need to come to that.

amp/app/config/styles.js
```diff
@@ -0,0 +1,124 @@
...
+styles.set(Link, [LinkBase, LinkTheme])
+styles.set(List, [ListTheme])
+styles.set(ListTile, [ListTileBase, ListTileTheme])
+styles.set(Nav, [NavBase, NavTheme])
+styles.set(NavItem, [NavItemBase, NavItemTheme])
+styles.set(NavMenu, [NavMenuBase, NavMenuTheme])
+styles.set(ProductItem, [ProductItemBase])
+styles.set(ProductTile, [ProductTileBase, ProductTileTheme])
+styles.set(Sheet, [SheetBase, SheetTheme])
+styles.set(SkipLinks, [SkipLinksBase, SkipLinksTheme])
```
jeffkamo: Is this file maintained manually, or is it generated?

Would it be easier to maintain if these `set.set()` calls are grouped with their respected component? Example...

```jsx
import SkipLinks from '../components/skip-links/index'
import SkipLinksBase from '../components/skip-links/_base.scss'
import SkipLinksTheme from '../components/skip-links/_theme.scss'
styles.set(SkipLinks, [SkipLinksBase, SkipLinksTheme])
```

jeffkamo:
One thing to be wary of when grouping your styles in this way is how they compile in the end. We use to group our style imports by component too:

```scss
@import 'button/base';
@import 'button/theme';

@import 'carousel/base';
@import 'carousel/theme';

@import 'icon/base';
@import 'icon/theme';

// etc.
```

But ran into situations where theme styles for one component could not override the styles of a shared component of another. Using the above example, the button theme cannot apply custom theme styles to its built in icon (i.e. `<Icon className="pw-button__icon">` which has the base `pw-icon` class). To address this, we had to import the styles this way instead:

```scss
// Bases
@import 'button/base';
@import 'carousel/base';
@import 'icon/base';
// etc.

// Themes
@import 'button/theme';
@import 'carousel/theme';
@import 'icon/theme';
// etc.
```

This ensured theme styles could override all base styles of any component.

It's not super clear to me what order your styles will be compiled, but I assume it will be the more like my first example?

jeffkamo:
@olibrook I'm not sure I understand how this works. Maybe I am missing something:

> _The second option would reduce redundant CSS rules_

Are you using another library to help Sass optimize the compiled CSS the way you described? I thought Sass would compile all defined selectors and rules regardless of redundancy?

olibrook:
@jeffkamo I thought SASS used to be able to optimize that case - maybe I'm mixing up my compilers though. I tested this morning and it doesn't make a difference after all.

amp/app/config/styles.js
olibrook: 1) Manually
2) I really don't mind.

olibrook:
@jeffkamo With respect to grouping - yes, this would work as your first example does. That is based on the assumption that components are isolated, don't depend on the cascade and can be included in any order. If they do depend on the cascade significantly I can't really propose a good fix to this problem at all.

cole-sanderson:
looks good to me, great work 👍

PR #802, opened 2017-06-22
AMP-71: Carousel Component 

cole-sanderson:
`<Carousel>` for PDP.

 **JIRA**: https://mobify.atlassian.net/browse/AMP-71

## Changes
- `<Carousel>` component
- added chevron left and right SVG icons

## How to test-drive this PR
- Run `npm update` to update `mobify-code-style`
- Run `npm run dev`
- Preview [Merlin's Potions](http://localhost:3000/potions/eye-of-newt.html)


amp/app/components/carousel/index.jsx
```diff
@@ -0,0 +1,294 @@
...
+
+        let layoutItemValue
+
+        if (typeCheck(typeCarousel)) {
+            layoutItemValue = layoutItem
+        } else {
+            throw new Error(`'${typeCarousel}' type does not support '${layoutItem}' layout`)
+        }
+
+        const childList = [prevChild, currentChild, nextChild]
```
olibrook: We need to do some work here to remove quite a lot of very browser-centric stuff. This line means that this `<Carousel />` works with at most 3 items!

olibrook:
I think this looks good to go now, except for that I think the prop names `controls` and `showControls` are not very intuitive on `Carousel`. Can we come up with better names?

I would suggest `showArrows` and `showPips`, since that's what they actually do. @cole-sanderson  - thoughts?

cole-sanderson:
@olibrook yeah I like that idea, will change that right now 👍

amp/app/components/carousel/_theme.scss
```diff
@@ -0,0 +1,59 @@
+@import '../../styles/component-base';
```
kschmidtdev: Is this file used? It seems similar to another one but I don't see what references it

cole-sanderson: We will need this file to use variables.

kschmidtdev: Sorry I mean `amp/app/components/carousel/_theme.scss` - I don't see it referenced anywhere.

PR #810, opened 2017-06-23
Web 1439: Cleanup the empty cart's layout

gsaray:
Cleanup the empty cart's layout

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1439

## Changes
- add padding top to empty cart
- make error color consistent and match the SDK 

## How to test-drive this PR
- Run `npm run dev`
- check empty cart and form errors


gsaray:
@jeffkamo looks like footer height was out of date for empty cart, it's fixed now.

web/app/containers/cart/container.jsx
```diff
@@ -41,7 +41,7 @@ export const continueShopping = () => {
 }
 
 const EmptyCartContents = ({hide, isLoggedIn}) => {
-    const emptyCartClassnames = classNames('t-cart__empty u-flexbox u-flex u-direction-column u-align-center u-justify-center', {
+    const emptyCartClassnames = classNames('t-cart__empty u-flexbox u-flex u-direction-column u-align-center u-justify-center u-padding-top-xlg', {
```
jeffkamo: Minor-ish: instead of adding a new class to the CSS, we could use two existing classes to achieve a similar effect: `u-padding-top-lg u-margin-top`

I feel like we should only add new utility classes like this if we feel like there's a strong need for them. For example, if we start seeing padding and margins doubled up in this manner might be a good indicator that we need larger paddings or margins.

Thoughts?

gsaray: agreed, I think it makes sense to add more utilities if it's needed in multiple places.

PR #816, opened 2017-06-26
use passwordInput component in login and registration forms

gsaray:
use passwordInput component in login and registration forms

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1419

## Changes
- use passwordInput in sign in and registration form

## How to test-drive this PR
- Run `npm run dev`
- Go to login page and check sign in and register forms

##Note
- We need to merge and publish https://github.com/mobify/progressive-web-sdk/pull/716 first then merge this

### DIFF
--- a/web/app/containers/login/partials/common.jsx
+++ b/web/app/containers/login/partials/common.jsx
@@ -43,7 +43,7 @@ export const LoginField = ({label, type, forgotPassword, name, tooltip, analytic
             component={Field}
             >
             {isPassword ?
-                <PasswordInput isText buttonTextHide="Hide" buttonTextShow="Show" />
+                <PasswordInput isText buttonTextHide="Hide" buttonTextShow="Show" data-analytics-name={analyticsName} />
             :
                 <input type={type} data-analytics-name={analyticsName} />
             }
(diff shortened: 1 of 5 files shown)


web/app/containers/login/partials/common.jsx
```diff
@@ -34,14 +35,18 @@ LoginFieldLabel.propTypes = {
...
+export const LoginField = ({label, type, forgotPassword, name, tooltip, analyticsName, isPassword}) => (
     <FieldRow>
         <ReduxFormField
             name={name}
             label={<LoginFieldLabel label={label} forgotPassword={forgotPassword} />}
             component={Field}
             >
-            <input type={type} data-analytics-name={analyticsName} />
+            {isPassword ?
+                <PasswordInput isText buttonTextHide="Hide" buttonTextShow="Show" />
```
jeffkamo: interestingly, even though `data-analytics-name` isn't being included on this element, I am still seeing analytics events being triggered. I suppose that is technically because it's added to the parent "component" (LoginField) in the `register-form.jsx` file.

@wizardlyhel is this intended behaviour?

gsaray: @jeffkamo We have the analytics on PasswordInput component, so we don't have separately call it here.

wizardlyhel: Double checked with analytics behaviour, this is slightly incorrect.

We need to pass in the `analyticsName` into `PasswordInput`. You can see the error when you go to register user page where there are 2 password inputs: password and confirm password

password field should have analytics name `password`
confirm password should have analytics name `confirm_password`

jeffkamo: Sounds like this will require a new SDK release, because those analytic values are hardcoded in the SDK component:

https://github.com/mobify/progressive-web-sdk/blob/develop/src/components/password-input/index.jsx

PR #817, opened 2017-06-26
AMP-46 - bundle upload command

olibrook:
**JIRA**: https://mobify.atlassian.net/browse/AMP-46

## Changes
- Change build script to build zip files (which Cloud expects)
- Provide  a `bundles.js` script for management of bundles on cloud.
- Include basic upload and list commands for the endpoints we have.

## Note
The list command is a little bare at the moment - we need it to show which version is the active one, which we can only do once we have deployments actually going through Cloud.

## How to test-drive this PR
- Set up and run Cloud locally.
- Set your machine up with the superuser credentials for deployment:

```
➜  amp git:(amp-46--upload-command) cat ~/.mobify
{
  "username": "fixture+super.user@test.com",
  "api_key": "asdf"
}
```
- Run:
```
npm run build
./scripts/bundles.js upload project-41 build/[commit].zip
./scripts/bundles.js list project-41
```

You should see a list of uploaded bundles.

### DIFF
--- a/amp/scripts/bundles.js
+++ b/amp/scripts/bundles.js
@@ -57,9 +57,9 @@ const list = ({projectSlug}) => {
 const upload = ({projectSlug, file}) => {
     const url = listCreateURL(projectSlug)
     step(`Uploading bundle "${file}" to "${url}"`)
-    // const parsed = path.parse(file)
-    // parsed.ext !== '.zip' && abort('Expected a zip file')
-    // parsed.name.length !== gitRevisionLength && abort(`Expected a ${gitRevisionLength}-character git revision`)
+    const parsed = path.parse(file)
+    parsed.ext !== '.zip' && abort('Expected a zip file')
+    parsed.name.length !== gitRevisionLength && abort(`Expected a ${gitRevisionLength}-character git revision`)
 
     const buildRequestBody = () => (
         fs.readFileAsync(file)
--- a/amp/scripts/bundles.js
+++ b/amp/scripts/bundles.js
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
-
 const process = require('process')
 const yargs = require('yargs');
 const path = require('path');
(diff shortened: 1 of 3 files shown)


amp/scripts/bundles.js
```diff
@@ -0,0 +1,112 @@
...
+        }))
+    )
+
+    return Promise.all([getCredentials(), buildRequestBody()])
+        .then(([credentials, body]) => fetch(url, {
+            method: 'POST',
+            headers: headers(credentials),
+            body: JSON.stringify(body)
+        }))
+        .then(res => res.json())
```
noahadams: the intention here is to pretty print the output? that might do well as a function

amp/scripts/bundles.js
```diff
@@ -0,0 +1,112 @@
...
+        .then(data => info(JSON.stringify(data, null, 4)))
```
noahadams: A catch for handling network errors might be nice

noahadams:
This looks basically functional, we'll want to think about what we output in future, but that's not a blocker until we're doing deploys as well, address comments and ths will have my plus one, nice work!

noahadams:
I lied, one last ask, some commandline or environment variable overrides for supplying API key would be useful, especially for testing or CI

amp/scripts/bundles.js
olibrook: What do you want that catch to do? I've got a top-level catch-all in `main`, so I could add something there.

amp/scripts/bundles.js
olibrook: Fixed

olibrook:
@n Added your environment variable, too.

amp/scripts/bundles.js
noahadams: I was thinking more `.then(res => pprint(res.json()))` but i guess .json() is promise returning?

olibrook: Ah, good shout - Cloud often responds with HTML on an error, so this is often not safe. Fixed by `rejecting()` on a bad status.

PR #818, opened 2017-06-26
remove remember me from sign in and register

gsaray:
remove remember me from sign in and register

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1440

## Changes
- remove remember me from sign in and register

## How to test-drive this PR
- Run `npm run dev`
- check sign in and register form


web/app/containers/login/partials/register-form.jsx
```diff
@@ -12,7 +12,7 @@ import {submitRegisterForm} from '../actions'
 import Button from 'progressive-web-sdk/dist/components/button'
 import FieldSet from 'progressive-web-sdk/dist/components/field-set'
 
-import {LoginField, RememberMeTooltip} from './common'
+import {LoginField} from './common'
```
jeffkamo: Minor non-blocking: the `LoginField` export could be made default so it doesn't need the curly braces when being imported.

PR #819, opened 2017-06-27
Fix Stretched Product Thumbnails

jeffkamo:
Certain images were getting stretched due to a mismatch in height/width ratios. The assets that we were using and the values that we were forcing upon them did not quite match.

## Changes
- Remove from our product parser where we override the product thumbnail with the images from the PDP carousels (they aren't the right dimensions for use in the thumbnails)

## Todos
- ~~(if applicable) analytics events instrumented to measure impact of change~~

## How to test-drive this PR
- Run `npm run dev`
- Visit a Product List page and note that height/width of the product thumbnails (they should look good)
- Click any of the products to visit its PDP and let it load, then hit the back button to return to the product list page
- **Verify** that the product thumbnails still look correct (don't stretch or deform)

## Result

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/web/app/connectors/_merlins-connector/products/parsers.js
+++ b/web/app/connectors/_merlins-connector/products/parsers.js
@@ -47,8 +47,7 @@ export const productDetailsParser = ($, $html) => {
         price: getTextFrom($mainContent, '.product-info-price .price-wrapper .price'),
         description: getTextFrom($mainContent, '.product.info.detailed .product.attibute.description p'),
         available: getAvailabilityFrom($mainContent),
-        images,
-        thumbnail: images[0]
+        images
     }
 }
 


web/app/connectors/_merlins-connector/products/parsers.js
```diff
@@ -47,8 +47,7 @@ export const productDetailsParser = ($, $html) => {
         price: getTextFrom($mainContent, '.product-info-price .price-wrapper .price'),
         description: getTextFrom($mainContent, '.product.info.detailed .product.attibute.description p'),
         available: getAvailabilityFrom($mainContent),
-        images,
-        thumbnail: images[0]
```
jeffkamo: The image asset coming from `images[0]` is a different asset than the thumbnail was before. Indeed, higher resolution, but it has different dimensions.

PR #821, opened 2017-06-27
Integration manager/cleanup left behind files

jeremywiebe:
During a merge conflict resolution, one of the Integration Manager files moved to the SDK was re-introduced to the scaffold.  This removes that file.

 **JIRA**: N/A
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/718

## Changes
- Removed leftover `cart/types.js` file from integration manager (which is now in the SDK)

## Todos

## How to test-drive this PR
- Run `npm run dev`
- The steps to test this PR can be found in the [related SDK PR](https://github.com/mobify/progressive-web-sdk/pull/718)


web/app/integration-manager/cart/types.js
```diff
@@ -1,67 +0,0 @@
-/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
```
jnurse: Are there any changes to this file that should be ported over to the SDK version?

jeremywiebe: @jnurse There was one change, but I actually don't agree with it.  

The `label` key on the Shipping object was made optional, but instead I did [this](https://github.com/mobify/platform-scaffold/pull/821/files#diff-297ac389b0066004475316289307a1caR61). 

Let me know if you agree. :)

jeremywiebe: Update. The exact change that was applied to this file has now been applied to the one in the SDK. 

https://github.com/mobify/progressive-web-sdk/blob/develop/src/integration-manager/cart/types.js#L37

PR #829, opened 2017-06-28
MTT-647 - more analytics instrumentation

wizardlyhel:
**JIRA**: https://mobify.atlassian.net/browse/MTT-647
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/723

## Changes
Add analytics instrumentation to the following:
* [x] Share component - https://mobify.atlassian.net/browse/WEB-1416
* [x] Carousel
* [x] Sort By select
* [x] Quantity steppers
* [x] Update to cart (Currently it’s triggering add_to_cart button event but not addToCart event)
* [x] Promo Code accordion in the card
* [x] modal dismiss button

## Todos

## How to test-drive this PR
1. In SDK branch, run `npm link`
2. In Scaffold branch, run `npm link progressive-web-sdk`
3. Run `npm run dev:build` in the SDK repo
4. Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)


wizardlyhel:
It's not a rename. `analyticsName` is a prop that gets pass down to this react component. `data-analytics-name` is the HTML DOM attribute to use.

If it's not passed, the console will error. This is just a react component type checking on components (I think)

wizardlyhel:
IRL Conversation with @sketchytechky 
The confusion comes in with the following example:
```jsx
<Button data-analytics-name={UI_NAME.menu} />
<IconLableButton analyticsName={UI_NAME.men} />
```
`Button` is a component that accepts any valid props it is given while `IconLableButton` is a component that accepts very specific props.

We have to use `data-*` on html dom elements because react will strip out any attributes that it doesn't recognize.

We cannot define a `data-analytics-name` prop because it will trigger complex code convention cases we need to handle.

web/app/containers/offline/partials/offline-banner.jsx
```diff
@@ -21,7 +21,7 @@ const OfflineBanner = ({
         <header className="t-offline-banner__wrapper">
             <div className="t-offline-banner u-flexbox u-align-center u-width-full">
                 <div className="u-flex u-margin-start-md">Currently browsing in offline mode</div>
-                <IconLabelButton iconName="info" label="" onClick={openModal} />
+                <IconLabelButton iconName="info" label="" onClick={openModal} analyticsName={UI_NAME.info} />
```
jeffkamo: Minor: I don't suppose we could take this opportunity to add an accessible label, could we?

PR #834, opened 2017-06-28
Web 1437: Add account dashboard (ACCOUNT BASE BRANCH)

gsaray:
Add account dashboard

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1437
 **Related PR**: https://github.com/mobify/progressive-web-sdk/pull/725

## Note:
- We need to merge https://github.com/mobify/progressive-web-sdk/pull/725 first and do a release before merging this PR
- This is going to be the base branch for all account pages, **DO NOT MERGE** until all pages are reviewed and merged into this branch
- Dashboard links are going to broken pages on SFCC until we add those page later

## Changes
- Add account dashboard page
- Only links that exist on both SFCC and merlins are displayed in dashboard

## Todos
- ~(if applicable) analytics events instrumented to measure impact of change~

## How to test-drive this PR
- checkout `WEB-1437__account` in the sdk
- run `npm link` (in SDK)
- run `npm install` (is SDK)
- checkout this branch 
- run `npm run dev`
- [Preview](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2Fcustomer%2Faccount%2Flogin%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- login with mobifyqa@gmail.com to land on dashboard


web/app/connectors/_merlins-connector/account/commands.js
```diff
@@ -120,8 +130,9 @@ export const navigateToSection = (router, routes, sectionName) => {
     }
 }
 
-export const logout = () => (dispatch) => (
-    makeRequest('/customer/account/logout/')
+export const logout = () => (dispatch) => {
+    browserHistory.push({pathname: '/'})
```
gsaray: I added this because user stays on account page after logging out.
what would be the best way to redirect users to homepage only when they logout from one of account pages?

jeffkamo:
Looks like the Nav's sign in link lost its background. Was that intentional?

| Before | This PR |
| --- | --- |
| screen shot 2017-06-29 at 12 09 50 pm | screen shot 2017-06-29 at 12 13 02 pm |

web/app/containers/account-dashboard/container.jsx
```diff
@@ -0,0 +1,70 @@
...
+import template from '../../template'
+
+import {getTitle, getLink} from './selectors'
+import SkeletonText from 'progressive-web-sdk/dist/components/skeleton-text'
+import ListTile from 'progressive-web-sdk/dist/components/list-tile'
+import Icon from 'progressive-web-sdk/dist/components/icon'
+
+
+const containerClass = 't-account-dashboard'
+const titleClass = 't-account-dashboard__title u-padding-md u-text-uppercase'
```
gsaray: yea, this was how the page was generated so I didn't change it, but it's updated now

web/app/containers/app/actions.js
```diff
@@ -70,16 +70,17 @@ export const fetchSvgSprite = () => (dispatch) => {
 }
 
 
-export const signOut = () => (dispatch) => (
-    dispatch(logout()).then(() => {
+export const signOut = () => (dispatch) => {
+    browserHistory.push({pathname: '/'})
```
Cdok: Jumping in here - this is how we had it implemented previously and there was a solid 1-2 second pause before the logout resolved & the user was redirected, so we moved the navigation before the command

jnurse: My only concern with it is what happens if the logout command fails? If we redirect before we run the command then they'll be sent to the homepage but won't actually be logged out. If we're concerned about the pause causing confusion then it might be better to show some sort of loading overlay first, dispatch the command and when it's successful redirect the user .

web/app/containers/account-dashboard/container.jsx
```diff
@@ -0,0 +1,76 @@
...
+    return (
+        <ListTile
+            className="t-account-dashboard__link"
+            href={href}
+            endAction={<Icon name="chevron-right" />}
+        >
+            {text ?
+                <div>{text}</div>
+            :
+                <SkeletonText
```
jeffkamo: With the Dashboard links being hardcoded as they are now, it's probably not necessary to provide placeholders anymore.

PR #835, opened 2017-06-29
Fix JsonpAsyncAsync Webpack Console Error

Cdok:
# Fix JsonpAsyncAsync Webpack Console Error

 **JIRA**: https://mobify.atlassian.net/browse/WEBDATA-149
 ~~**Linked PRs**: (links to corresponding PRs, optional)~~

## Changes
- Updated regex to no longer match against any string instance of `webpackJsonp`, and instead also optionally match against the entire `webpackJsonpAsync` string.

## Todos
- [x] ~~(if applicable) analytics events instrumented to measure impact of change~~

## How to test-drive this PR
- Test the regex against `webpackJsonp`, `webpackJsonpAsync`, and `webpackJsonpAsyncAsync` strings

### DIFF
--- a/web/webpack/base.main.js
+++ b/web/webpack/base.main.js
@@ -128,7 +128,7 @@ config.plugins.push(function() {
                 }
 
                 const source = children[0]
-                source._value = source._value.replace(/^webpackJsonp/, 'webpackJsonpAsync')
+                source._value = source._value.replace(/^webpackJsonp(.*)?/, 'webpackJsonpAsync')
             }
         }
 
--- a/web/webpack/base.main.js
+++ b/web/webpack/base.main.js
@@ -128,7 +128,7 @@ config.plugins.push(function() {
                 }
 
                 const source = children[0]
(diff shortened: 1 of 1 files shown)


web/webpack/base.main.js
```diff
@@ -128,7 +128,7 @@ config.plugins.push(function() {
                 }
 
                 const source = children[0]
-                source._value = source._value.replace(/^webpackJsonp/, 'webpackJsonpAsync')
+                source._value = source._value.replace(/^webpackJsonp(.*)?/, 'webpackJsonpAsync')
```
jeremywiebe: I'm a bit concerned with how greedy this regex is.  It's basically matching _any_ character following `webpackJsonp`   Really we only want to match any characters following that string, right? 

So
```javascript
 source._value = source._value.replace(/^webpackJsonp\w*/, 'webpackJsonpAsync')
```

Also, `*` in a regex is 0, or more, so the `?` following that isn't really needed.

PR #841, opened 2017-06-29
Loads app in parallel with the service worker

jansepar:
Loads app in parallel with the service worker

## Changes
- Currently, all of our files necessary for the rendering of the app don't get loaded until *after* the Service Worker downloads. This hurts TTI by [what seems to be upwards of 2.5 seconds](http://www.webpagetest.org/result/170626_ZN_1CE1/)!
- Also loads web push in parallel with the app so that it's compilation + execution doesn't cause blocking later.
- Also ensures that you don't need the SW to successfully install in order to use the app.

## How to test-drive this PR
- Run `npm run dev`
- @Cdok  ensure that your issue doesn't happen anymore.

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -205,16 +205,16 @@ const attemptToInitializeApp = () => {
         loadAsset('meta', {
             name: 'charset',
             content: 'utf-8'
-        });
+        })
 
         loadAsset('link', {
-                href: getAssetUrl('main.css'),
-                rel: 'stylesheet',
-                type: 'text/css',
-                // Tell us when the stylesheet has loaded so we know when it's safe to
-                // display the app! This prevents a flash of unstyled content.
-                onload: 'window.Progressive.stylesheetLoaded = true;'
-            })
+            href: getAssetUrl('main.css'),
+            rel: 'stylesheet',
+            type: 'text/css',
+            // Tell us when the stylesheet has loaded so we know when it's safe to
+            // display the app! This prevents a flash of unstyled content.
+            onload: 'window.Progressive.stylesheetLoaded = true;'
+        })
 
         loadAsset('link', {
             href: getAssetUrl('static/manifest.json'),
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -10,6 +10,7 @@ import {
     loadAndInitMessagingClient,
     loadScript,
     loadScriptAsPromise,
+    prefetchLink,
     updateMessagingSWVersion
 } from './utils/loader-utils'
 import {getNeededPolyfills} from './utils/polyfills'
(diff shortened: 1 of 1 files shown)


jansepar:
Test results:

https://www.webpagetest.org/result/170629_YH_1HXZ/

web/app/loader.js
```diff
@@ -27,162 +28,154 @@ window.Progressive = {}
...
+    // will find nothing in localStorage, and return the URL without
+    // any Messaging-worker parameters, but we'll do an asynchronous
+    // fetch to update the parameters, which will then be used on the
+    // next run.
+
+    // We expect supported browsers to have local storage. If the browser
+    // does not, then we may assume we're running in some situation
+    // like incognito mode, in which case there is no point getting
+    // Messaging worker version data, we can just use the base URL.
+    if (!isLocalStorageAvailable()) {
```
jeremywiebe: I'm a bit concerned that we exit early here. The rest of this _currently_ deals with code that expects the existence of local storage, but in the future I could easily see people adding other code that extends the URL and is unrelated to the existence of local storage... and then it breaking in incognito. 

Could we reverse this `if` and nest the `swVersion` in it and only return at the bottom?

web/app/loader.js
```diff
@@ -27,162 +28,154 @@ window.Progressive = {}
...
+/**
+ * Load the service worker
+ * @returns Promise.<Boolean> true when the worker is loaded and ready
+ */
+const loadWorker = () => (
+    navigator.serviceWorker.register(getServiceWorkerURL())
+        .then(() => navigator.serviceWorker.ready)
+        .then(() => true)
+        .catch(() => {
+        })
```
jeremywiebe: I know this isn't your code, but empty `catch()` blocks in any language always look suspect. We should have a comment here that at least says we're swallowing errors here on purpose.

web/app/loader.js
```diff
@@ -27,162 +28,154 @@ window.Progressive = {}
...
+)
+
+const asyncInitApp = () => {
+    window.webpackJsonpAsync = (module, exports, webpackRequire) => {
+        const runJsonpAsync = () => {
+            if (window.webpackJsonp) {
+                // Run app
+                window.webpackJsonp(module, exports, webpackRequire)
+            } else {
+                setTimeout(runJsonpAsync, 50)
```
jeremywiebe: Should we have an upper bound on how long we wait?  Not sure that's a realistic situation to have a "hang" waiting for `webpackJsonp`

jansepar: Hmm. It's tough because if someone is on a very flaky connection, `vendor.js` could take suuuuper long to load, so it's hard to put a hard coded upper bound without some kind of heuristic based on the connection. And that doesn't seem worth it :)

jeremywiebe: Fair enough... but... 

giphy

PR #846, opened 2017-06-30
Add delay to modal dismiss

kevinxh:
Modal manager removes modal right away when user tries to close the modal, that stops the close animation. This PR introduces a delay for removing modals.

It also introduces a way to unify the transition time for modals throughout a project:
`<ModalManager duration={200} />`

as well as set individual modal transition time (inside ModalManger):
`[MODAL.NAVIGATION_MODAL]: {content: <Navigation />, customDuration: 2000}`

 **JIRA**: none
 **Linked PRs**: none

## Changes
- Add delay for close animation
- Unify transition time for all modals in Modal Manager

## Todos

## How to test-drive this PR
- Run `npm run dev`
- Open and close modals to verify the animation is working


web/app/modals/index.jsx
```diff
@@ -30,20 +30,44 @@ const modals = {
...
+        const isOpen = this.props.isOpen
+
+        for (const nextModal in nextIsOpen) {
+            // Open Modal
+            if (nextIsOpen[nextModal] === true) {
+                return true
+            } else {
+                // Close Modal
+                // Set a delay for modal dismiss animation
+                // `delay` should be larger than transition time
```
jeffkamo: Which transition time is this referring to? Is this referring to a modal's `duration`?

kevinxh: @jeffkamo yes, it refers to `duration` for `Sheet`, I should have make this more clearer. Here the delay is hardcoded and I wonder if there is a way to collect the duration of all modals or the opened modal..

jeffkamo: +1, it would be much more ideal if we could get the duration from the modal being closed

web/app/modals/index.jsx
```diff
@@ -19,35 +19,82 @@ import OfflineModal from './offline/container'
...
+                return true
+            } else {
+                // Close Modal
+                // Set a delay for modal close animation
+                if (isOpen[nextModal] !== nextIsOpen[nextModal]) {
+                    const delay = modals[nextModal].customDuration || duration
+                    setTimeout(() => this.forceUpdate(), delay)
+                    return false
+                }
+            }
```
jeffkamo: Minor, non-blocking comment on the formatting of this `if` statement and it's comments:

Since the first `if` either returns or doesn't, you can safely not have an `else`. The plus side is that this also saves you some indentations.

```js
for (...) {
    // Open Modal
    if (const nextModal in nextIsOpen) {
        return true;
    }

    // Close Modal
    if (isOpen[nextModal] !== nextIsOpen[nextModal]) {
        // Set a delay for modal close animation
        // ...
    }
}
```

kevinxh: Fixed.

PR #849, opened 2017-07-03
Add stub connector unit tests

jeremywiebe:
This PR introduces a very basic set of tests for the stub connector.

 **JIRA**: N/A
 **Linked PRs**: N/A

## Changes
- Uses a generator type function to test every command exported by the connector
- Skips `fetchPageData` as that uses the `fetch` API which isn't supported in node

## Todos

## How to test-drive this PR
- Run `npm test`


web/app/connectors/_stub-connector/commands.test.js
```diff
@@ -0,0 +1,48 @@
...
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+/* eslint-env jest */
+import commands from './commands'
+
+export const testStubCommand = (command, ...args) => {
+    it(`${command.name} executes without errors`, () => {
+        const dispatch = () => Promise.resolve()
+        return command(...args)(dispatch)
```
cdegit: One thing I've run into with this test on Debs is that it would be helpful to also mock getState

jeremywiebe: @cdegit Are you using this generic tester for testing the custom connector?  I was thinking this would be used to cover the stub commands generically, but once a custom connector was started each command would have it's own specific test(s).

cdegit: No, we're just using it for the stub commands. The main thing for Debs right now is that the devs are frequently running into issues with the tests. When they implement a command, the test for that command will fail, which is slowing them down a bit.

jeremywiebe:
@cdegit Check out the tiny `getState` stub I added.  Do you think that'll address the issue?

web/app/connectors/_stub-connector/commands.test.js
```diff
@@ -0,0 +1,50 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+/* eslint-env jest */
+import commands from './commands'
+
+export const testStubCommand = (command, ...args) => {
+    it(`${command.name} executes without errors`, () => {
+        const dispatch = () => Promise.resolve()
```
fractaltheory: We could potentially use something like `redux-mock-store` to stub out these Redux arguments - but that's overkill here.

jeremywiebe: My thinking for this was that the stub connector shouldn't really be accessing the store here. I could see it as you build out a custom connector so maybe a connector test helper might be useful in the Scaffold.  I'm going to avoid using the `redux-mock-store` here though.

PR #801, opened 2017-06-21
DRY-up routing

kschmidtdev:
(brief description / PR title)

~~**JIRA**: (link to JIRA ticket)~~

## Changes
- DRY-up routing
- Make our README.md meet the guidelines for [minimal-viable-documentation](https://github.com/mobify/developer-values#%EF%B8%8F-write-great-documentation)

## How to test-drive this PR
- Check that everything works as before

PR #851, opened 2017-07-04
Fix for building for lambda breakage

kschmidtdev:
Don’t run our `postinstall` logic when building for production, because it won’t work, and we don’t need it

 **Linked PRs**: #801 

## Changes
- Conditionally run `postinstall` logic based on node env

## How to test-drive this PR
- `npm i`
- Make sure everything works as normal
- `npm run build`
- Make sure build is successful


olibrook:
👍

PR #855, opened 2017-07-04
Fix messaging client load and init promise

fractaltheory:
Recent changes to `loader.js` broke the loading of Messaging Client as well as the order of app load and Init Promise initialization.

## Changes
- Since `loadAndInitMessagingClient` is no longer added to an array of deferred promises, remove "thunk" and just call the script load directly
- Add new method `addMessagingClientInitPromise` that sets up the Init Promise on window before `vendor.js` is loaded
- Modify error conditions in the script loading function - to avoid throwing multiple errors in console on error

## Todos
- [x] Test
- [x] +1

## How to test-drive this PR
- Run `npm run dev`
- On the second page view, ensure you are shown the `DefaultAsk` messaging ask prompt (on supported devices, i.e. Chrome)

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -8,6 +8,7 @@ import {
     getMessagingSWVersion,
     isLocalStorageAvailable,
     loadAndInitMessagingClient,
+    addMessagingClientInitPromise,
     loadScript,
     loadScriptAsPromise,
     prefetchLink,
@@ -256,6 +257,10 @@ const attemptToInitializeApp = () => {
             src: getAssetUrl('main.js')
         })
 
+        // This must come before vendor.js, or before the Webpack chunk that contains
+        // Messaging React components
+        addMessagingClientInitPromise(messagingEnabled)
+
         loadScriptAsPromise({
             id: 'progressive-web-vendor',
             src: getAssetUrl('vendor.js')
@@ -275,10 +280,6 @@ const attemptToInitializeApp = () => {
          ? loadWorker()
          : Promise.resolve(false)
         ).then((serviceWorkerSupported) => {
-
-            // Set up the Messaging client integration - this must be
-            // done now, but the work is deferred until after script
-            // loading is complete.
             setupMessagingClient(serviceWorkerSupported, messagingEnabled)
         })
 
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -257,8 +257,13 @@ const attemptToInitializeApp = () => {
             src: getAssetUrl('main.js')
         })
 
-        // This must come before vendor.js, or before the Webpack chunk that contains
-        // Messaging React components
+        /**
+         * This must be called before vendor.js is loaded (or before the Webpack
+         * chunk that contains Messaging React components is loaded)
+         *
+         * This creates a Promise: `window.Progressive.MessagingClientInitPromise`
+         * which will be resolved or rejected later by the method `setupMessagingClient`
+         */
         createGlobalMessagingClientInitPromise(messagingEnabled)
 
         loadScriptAsPromise({
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -8,7 +8,7 @@ import {
     getMessagingSWVersion,
     isLocalStorageAvailable,
     loadAndInitMessagingClient,
-    addMessagingClientInitPromise,
+    createGlobalMessagingClientInitPromise,
     loadScript,
     loadScriptAsPromise,
     prefetchLink,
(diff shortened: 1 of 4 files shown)


web/app/utils/loader-utils.js
```diff
@@ -72,23 +72,29 @@ export const isLocalStorageAvailable = () => {
 
 const MESSAGING_PWA_CLIENT_PATH = 'https://webpush-cdn.mobify.net/pwa-messaging-client.js'
 
+// Creating an early promise that users of the Messaging Client can
+// chain from means they don't need to poll for its existence
+let clientInitResolver = () => {}
+let clientInitRejecter = () => {}
+export const addMessagingClientInitPromise = (messagingEnabled) => {
```
jansepar: `add` is a bit confusing to me. "Add to what" was the first thing that came to my mind.

How about something like `createGlobalMessengingClientInitPromise` - it's a bit more descriptive about what it's actually doing.

fractaltheory: Fixed: https://github.com/mobify/platform-scaffold/pull/855/commits/2d082be93e04e306990a2c0993b887ab55308013

web/app/utils/loader-utils.js
```diff
@@ -72,23 +72,29 @@ export const isLocalStorageAvailable = () => {
...
+    if (!messagingEnabled) {
+        return
+    }
+
+    window.Progressive.MessagingClientInitPromise = new Promise((resolve, reject) => {
```
jansepar: What actually checks for this promise? I can't find anything doing a `then`/`catch` on it.

web/app/utils/loader-utils.js
jansepar: I think it still needs to be updated here?

web/app/utils/loader-utils.js
fractaltheory: Messaging React components in the SDK.

web/app/utils/loader-utils.js
fractaltheory: Oops good catch, https://github.com/mobify/platform-scaffold/pull/855/commits/a1bd04e597ff8d8f4094694f35a6ce1654fadc41

web/app/utils/loader-utils.js
```diff
@@ -72,23 +72,29 @@ export const isLocalStorageAvailable = () => {
 
 const MESSAGING_PWA_CLIENT_PATH = 'https://webpush-cdn.mobify.net/pwa-messaging-client.js'
 
+// Creating an early promise that users of the Messaging Client can
+// chain from means they don't need to poll for its existence
+let clientInitResolver = () => {}
+let clientInitRejecter = () => {}
```
jansepar: Why are these declared outside of the function? They don't seem to be needed by anything else in this `loader-utils` module.

fractaltheory: They are called at https://github.com/mobify/platform-scaffold/pull/855/files#diff-b4393702fa401717bdebc13228dfc3d9R108 and https://github.com/mobify/platform-scaffold/pull/855/files#diff-b4393702fa401717bdebc13228dfc3d9R116 - essentially the `messagingClient.init` promise chain will resolve the InitPromise, or an error thrown at any point in the load or `messagingClient.init` promises will be caught by rejecting the InitPromise.

jansepar: Ah yeah okay they totally are, not sure how I missed that!

jansepar:
👍

web/app/loader.js
```diff
@@ -256,6 +257,10 @@ const attemptToInitializeApp = () => {
             src: getAssetUrl('main.js')
         })
 
+        // This must come before vendor.js, or before the Webpack chunk that contains
+        // Messaging React components
+        createGlobalMessagingClientInitPromise(messagingEnabled)
```
benlast: Could we have a comment here to explain that the `Promise` is created by this call, but resolved when `setupMessagingClient` is called? Just to make it easier for a dev to follow the flow?

web/app/utils/loader-utils.js
benlast: Should the default implementations throw and/or log an error, so that a failure to call `createGlobalMessagingClientInitPromise` and `setupMessagingClient` in the correct pattern will be more obvious?

web/app/loader.js
fractaltheory: Great idea, added: https://github.com/mobify/platform-scaffold/pull/855/commits/1c0c65a61112f776aa8988e51a9a6d0f070d93b8

fractaltheory:
Can't reply to your inline comment, but I've modified the initial resolve/reject variables to `console.error`: `LoaderUtils.createGlobalMessagingClientInitPromise must be called before setupMessagingClient`

PR #860, opened 2017-07-06
Upgrade test:build-size

jasecode:
Web-team would like to verify the build files sizes do not get any bigger than they are now so we're upgrading our test:build-size!

With this test, we will fail if any file in the build folder exceeds the threshold of the size we set it to be.

We have a configurable file: file-size-config.json of which we will list a bunch of files and their gzipped filesize. We will be running this test through every CircleCI build.

Please note that we are comparing it against the predicted gzipped file-size. We are using the 'gzip-size' to do this.

## Changes
- updated verify-build-size.js
- add file-size-config.json
- removed the initial config from package.json

## Todos
-  Get more of a list of build file sizes that we're interested in and their threshold.

## How to test-drive this PR
- npm run prod:build
- change file-size-config.json to add/edit a file you would like to check if it's over the threshold
- npm run test:build-size


web/tests/system/test-scripts/verify-build-size.js
```diff
@@ -3,30 +3,59 @@
...
             const fileStat = fs.statSync(filePath)
+
+            /* Checks each file - if it's over size limit */
             if (fileStat.size > FILE_SIZE_LIMIT) {
                 failure = true
                 console.log(chalk.red(`${filePath} is ${fileStat.size} bytes. It is too big!\n`))
             }
+
+            /* Get the gzipped file size and parse file-size-config.json*/
+            const source = fs.readFileSync(filePath, 'utf8')
```
ellenmobify: Move lines 45-47 to inside the `if` below, so that we only check the gzipped size of the files that are listed in `file-size-config.json`

jasecode: Nice catch.

web/tests/system/test-scripts/verify-build-size.js
```diff
@@ -3,30 +3,57 @@
...
+* The test will fail if it goes above the threshold.
 */
+
 const options = {
     listeners: {
         file: (root, fileStats, next) => {
             const filePath = path.join(root, fileStats.name)
             const fileStat = fs.statSync(filePath)
+
+            /* Checks each file - if it's over size limit */
```
ellenmobify: Mention that this check is for minified files

web/tests/system/test-scripts/verify-build-size.js
```diff
@@ -44,6 +71,7 @@ const options = {
 
 if (fs.existsSync('build')) {
     console.log(`Verifying individual file sizes in the build are less than ${FILE_SIZE_LIMIT} bytes...`)
```
ellenmobify: Mention again down here that one check is for minified and the second is for gzipped files.

web/tests/system/test-scripts/file-size-config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "bundleSize": {
+        "max": 2000000,
+        "files": {
+            "vendor.js": 174000,
+            "main.js": 40000,
```
jeffkamo: Are there any other JS files we should track, @jansepar? Perhaps `loader.js`?

jeffkamo: cc @jeremywiebe

jansepar: +1 to loader.

web/tests/system/test-scripts/file-size-config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "bundleSize": {
+        "max": 2000000,
+        "files": {
+            "vendor.js": 174000,
```
jansepar: What are the current build sizes? We're going to need to leave some buffer room here to start if these are the current sizes.

jasecode: @jansepar We're at the moment  giving like a 1kb of buffer. What would the recommended buffer be?

jansepar: It's definitely gonna get bigger then 1kb. I would say 200KB would be the max upper limit. If we switch to Preact we can lower it.

jasecode: Alright. will change the thresholds to be 200kb plus the current size. These are configurable in this file if needed.

PR #861, opened 2017-07-07
Relocate the setup of messaging promise in loader

fractaltheory:
So, I thought that `vendor.js` contains everything in `node_modules` but that doesn't seem to be the case. For safety, set up the Messaging promises before both `main.js` and `vendor.js` are loaded.

**Linked PR**: https://github.com/mobify/platform-scaffold/pull/855

## Changes
- Move `createGlobalMessagingClientInitPromise` above `main.js` and `vendor.js` loading in loader.js

## Todos
- [x] +1

## How to test-drive this PR
- Run `npm run dev`
- Messaging should be loaded, and DefaultAsk shown on the second page visit

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -252,19 +252,19 @@ const attemptToInitializeApp = () => {
             src: getAssetUrl('static/js/jquery.min.js')
         })
 
-        loadScript({
-            id: 'progressive-web-main',
-            src: getAssetUrl('main.js')
-        })
-
         /**
-         * This must be called before vendor.js is loaded (or before the Webpack
-         * chunk that contains Messaging React components is loaded)
+         * This must be called before the Webpack chunk that contains Messaging
+         * React components is loaded - right now that's main.js.
          *
          * This creates a Promise: `window.Progressive.MessagingClientInitPromise`
          * which will be resolved or rejected later by the method `setupMessagingClient`
          */
         createGlobalMessagingClientInitPromise(messagingEnabled)
+
+        loadScript({
+            id: 'progressive-web-main',
+            src: getAssetUrl('main.js')
+        })
 
         loadScriptAsPromise({
             id: 'progressive-web-vendor',


web/app/loader.js
```diff
@@ -252,20 +252,20 @@ const attemptToInitializeApp = () => {
...
-         * chunk that contains Messaging React components is loaded)
+         * This must be called before the Webpack chunk that contains Messaging
+         * React components is loaded - right now that's main.js.
          *
          * This creates a Promise: `window.Progressive.MessagingClientInitPromise`
          * which will be resolved or rejected later by the method `setupMessagingClient`
          */
         createGlobalMessagingClientInitPromise(messagingEnabled)
 
+        loadScript({
```
jeffkamo: Has this been tested with WebpageTest to verify if it has any impact on how the scripts load?

fractaltheory: No, but it shouldn't - all this method does is attach a Promise to `window`. Both `main.js` and `vendor.js` are loaded asynchronously anyway.

fractaltheory: That said, though, are there instructions for running WPT with preview anywhere?

jeffkamo: There are no instructions, but I'll throw something together right now and link you.

jeffkamo: Turns out there was already an answer! 

https://mobify.atlassian.net/wiki/questions/177544011/how-do-i-test-a-websites-performance-using-webpagetest

fractaltheory: @jeffkamo thanks for the link. WPT shows some wildly different results - one set of runs shows (First Interactive) 11.4s vs 12.8s - while another set shows 7.67s vs 7.74s :man_shrugging: As expected, though, this change does not introduce a noticeable performance change.

jeffkamo: This is why I like to run 9 tests rather than 3. Anyway, glad there are still results hitting good TTI!

jansepar:
👍

PR #867, opened 2017-07-10
Fix SFCC Connector

kschmidtdev:
Fixed up a bunch of issues with AMP and the SFCC Connector. The PR will remain configured for SFCC until it is approved, then I will switch it back to the Merlins one. 

 **JIRA**: https://mobify.atlassian.net/browse/AMP-76

## Changes
- Fixed a bunch of browser environment dependencies in the SFCC Connector

## Todos
- [x] Document how to configure for SFCC

## How to test-drive this PR
- Run `npm run dev`
- Check out http://localhost:3000/s/2017refresh/mens for the PLP, and http://localhost:3000/s/2017refresh/11736753.html for the PDP


web/app/connectors/_sfcc-connector/config.js
```diff
@@ -36,3 +38,8 @@ export const getPaymentURL = () => `${getBaseURL()}COBilling-Start`
 export const getConfirmationURL = () => `${getBaseURL()}COSummary-Submit`
 
 export const buildSearchURL = (query) => `${SEARCH_URL}${buildQueryString(query)}`
+
+export const configuredStorageAvailable = () => config.setItemInStorage && config.getItemInStorage && config.removeItemInStorage
```
jeremywiebe: I'm concerned about the growing list of things we need to "polyfill" here.  What do you think about pushing the storage things into the SDK as an abstraction that's usable for both browser and node environments.  The ultimate fallback for the node environment could be just the js dictionary you've used.  At the end of the day I don't think it matters if storage is available as we should treat it as a cache and not as durable storage anyways (check if we have a cached value, if so, use it, otherwise do the work to get the value). 

That would go for the `atob` stuff to.  We could have a `base64Encode()` and `base64Decode()` function in the SDK. Tree shaking should effectively eliminate that code in the app bundle.

amp/app/data-integration/connectedStore.js
```diff
@@ -27,12 +30,33 @@ import {PAGE_TITLE} from './constants'
...
+
 export const jsdomEnv = () => jsdom.envAsync('', ['http://code.jquery.com/jquery.js']) // TODO: Use local copy
 
-export const initializeStore = (fullUrl, container) => {
+export const initializeStore = (fullUrl, containers) => {
     return jsdomEnv().then((window) => {
         registerConnector(Connector({
-            jqueryResponse: jqueryResponse(window)
+            jqueryResponse: jqueryResponse(window),
+            setItemInStorage,
```
kschmidtdev: I don't know if I would call them polyfills, but I agree they should be grouped - I'll do this when I take another stab at this.

web/app/connectors/_sfcc-connector/utils.js
```diff
@@ -104,7 +110,7 @@ export const getAuthTokenPayload = (authToken) => {
     // The token consists of 3 parts: header, payload and signature
     // separated by a '.', each part is encoded
     // we only need the payload
-    return JSON.parse(window.atob(authToken.split('.')[1]))
+    return JSON.parse((atob()(authToken.split('.')[1])))
```
kschmidtdev: You're right! I think the parenthesis might have been required, but I'll check when I get back to this and make it work the same for node and browser.

web/app/connectors/_sfcc-connector/config.js
kschmidtdev: Great ideas, I'll gather up what utils could be offered via the SDK and get them in there.

amp/app/data-integration/connectedStore.js
jeremywiebe: Agreed.  Polyfills was the first thing that popped into my head and I didn't want to get sidetracked with NamingThings™️    

😄

kschmidtdev:
@jeremywiebe I know there is one conflict, but when you get a chance can you re-review the parts you gave feedback on? Thanks!

kschmidtdev:
@jeremywiebe when you get a chance can you take another look at this? Thanks!

web/app/connectors/_sfcc-connector/config.js
```diff
@@ -12,11 +18,16 @@ let config = {}
 
 export const registerConfig = (cfg) => {
     config = cfg
+    if (!config.storageType) {
```
jeremywiebe: We solved this similarly in the `hybris-connector` branch [here](https://github.com/mobify/platform-scaffold/pull/893/files#diff-6b264bf71de359a2d33fb5b042e5874fR22).  I think I like that solution a bit better as we let the library deal with the fallbacks and can easily add other storage engines without adjusting the fallback code.

kschmidtdev: Is there a reason why there is no cookie storage in there?

kschmidtdev:
@jeremywiebe Any other feedback prior to 👍 ? 😄

web/app/connectors/_sfcc-connector/config.js
jeremywiebe: Not really. I suspect it started out as direct use of `sessionStorage` and then was never expanded. I don't see a reason why we can't include cookie storage in between session and memory storage.

amp/app/data-integration/connectedStore.js
```diff
@@ -6,15 +6,11 @@ import {createStore, combineReducers, compose, applyMiddleware} from 'redux'
...
 import {fromJS} from 'immutable'
-import Promise from 'bluebird'
-import _jsdom from 'jsdom'
-import fs from 'fs'
 
 // DO NOT USE! Merlins Connector is an example connector that is for demo only
-import {Connector} from '../../../web/app/connectors/_merlins-connector'
-// import {Connector} from '../../../web/app/connectors/_sfcc-connector'
+// import {initConnector} from './connectors/merlins-connector'
+import {initConnector} from './connectors/sfcc-connector'
```
jeffkamo: Perhaps this shouldn't be committed.

web/package.json
```diff
@@ -43,6 +43,7 @@
     "sw-toolbox": ">=3.4.0"
   },
   "devDependencies": {
+    "atob": "^2.0.3",
```
jeffkamo: Should this `^` be removed?

amp/app/data-integration/connectedStore.js
jeffkamo: oh this is merging into `amp`. Maybe this is fine for now, sorry!

kschmidtdev: I will reverse this before committing 😄

web/package.json
kschmidtdev: Good catch, that was put there because it was added via `npm i --save-dev`

PR #868, opened 2017-07-10
Change where modify ga id is defined

wizardlyhel:
Make sure Mobify GA ID is lives in the `package.json` so that platform generator can update it

 **JIRA**: https://mobify.atlassian.net/browse/MTT-652

## Changes
- move where ga-id is defined

## How to test-drive this PR
- Run `npm run dev`

Expected: Mobify GA analytics are firing

### DIFF
--- a/web/app/store/index.js
+++ b/web/app/store/index.js
@@ -25,7 +25,7 @@ import analyticsManager from 'progressive-web-sdk/dist/analytics/analytics-manag
 
 analyticsManager.init({
     projectSlug: AJS_SLUG,              // eslint-disable-line no-undef
-    mobifyGAID: 'UA-53825302-1',
+    mobifyGAID: MOBIFY_GA_ID,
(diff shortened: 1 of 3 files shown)


web/app/store/index.js
```diff
@@ -25,7 +25,7 @@ import analyticsManager from 'progressive-web-sdk/dist/analytics/analytics-manag
 
 analyticsManager.init({
     projectSlug: AJS_SLUG,              // eslint-disable-line no-undef
-    mobifyGAID: 'UA-53825302-1',
+    mobifyGAID: MOBIFY_GA_ID,           // eslint-disable-line no-undef
```
jeremywiebe: I would really like to see all new webpack-created constants prefixed with `WEBPACK_`.  Tracing down where these constants come from can be time consuming.  Let's start with this one and we'll come back and fix the rest later. :)

PR #870, opened 2017-07-10
WEB-1427: view profile 

gsaray:
Account view and edit profile

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1427 and https://mobify.atlassian.net/browse/WEB-1428
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/756

## Changes
- Add account view and edit profile

## Todos

## How to test-drive this PR
- checkout `Account-Commands` in the sdk
- run `npm link` (in SDK)
- run `npm install` (is SDK)
- checkout this branch 
- run `npm run dev`
- [Preview](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2Fcustomer%2Faccount%2Flogin%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- login with mobifyqa@gmail.com to land on dashboard
- click on account information


Cdok:
Do not merge until https://github.com/mobify/platform-scaffold/pull/834 is merged

web/app/connectors/_sfcc-connector/account/commands.js
```diff
@@ -175,3 +179,57 @@ export const updateBillingAddress = (formValues) => (dispatch) => {
...
+        })
+}
+
+
+export const updateAccountInfo = ({names, email}) => (dispatch) => {
+    const {sub} = getAuthTokenPayload()
+    const customerId = JSON.parse(sub).customer_info.customer_id
+    const requestBody = {
+        first_name: names.split(' ')[0],
+        last_name: names.split(' ')[1],
```
jnurse: There should be a util function `splitFullName` that can split these for you.

web/app/connectors/_sfcc-connector/account/commands.js
```diff
@@ -175,3 +179,57 @@ export const updateBillingAddress = (formValues) => (dispatch) => {
...
+        email
+    }
+
+    return makeApiJsonRequest(`/customers/${customerId}`, requestBody, {method: 'PATCH'})
+        .then(({first_name, last_name, login}) => {
+            const result = {
+                accountFormInfo: {
+                    names: `${first_name} ${last_name}`,
+                    email: login
+                }
```
jnurse: This is pretty basic parsing, but since it's the same as the parsing we're doing in `initAccountInfoPage` could we make it into a parsing function that both commands call? 

We could even turn the then portion of this promise chain into a function that both commands use, something like `handleAccountInfoData` that parses the data then calls `dispatch(recieveAccountInfoUIData((result))) `.

web/app/connectors/_sfcc-connector/config.js
```diff
@@ -29,14 +29,15 @@ const getBaseURL = () => `/on/demandware.store/Sites-${getSiteID()}-Site/default
...
 export const getSignOutURL = () => `${getBaseURL()}Logout-Logout`
 export const getCheckoutShippingURL = () => `${getBaseURL()}COShipping-Start`
 export const getCartURL = () => `${getBaseURL()}Cart-Show`
 export const getWishlistURL = () => `${getBaseURL()}Wishlist-Show`
 export const getPaymentURL = () => `${getBaseURL()}COBilling-Start`
 export const getConfirmationURL = () => `${getBaseURL()}COSummary-Submit`
 
-export const getMyAccountURL = () => `${getBaseURL()}/customer/account/`
+export const getMyAccountURL = () => `/customer/account`
+export const getWishlistURL = () => `${getBaseURL()}wishlist/`
```
jnurse: Can we make sure all these URLs match the equivalent page on the desktop site? It might take a bit of additional logic but I think it's necessary. Otherwise we'll have trouble navigating directly to one of these pages.

web/app/containers/account-info/actions.js
```diff
@@ -0,0 +1,64 @@
...
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
+import {validateFullName} from '../../utils/utils'
+import {updateAccountInfo, updateAccountPassword} from 'progressive-web-sdk/dist/integration-manager/account/commands'
+import {addNotification} from 'progressive-web-sdk/dist/store/notifications/actions'
+import {SubmissionError} from 'redux-form'
+
+export const receiveData = createAction('Receive AccountInfo data')
```
jnurse: Is this action being used? If not can we remove it?

web/app/containers/account-info/actions.js
```diff
@@ -0,0 +1,64 @@
...
+
+// This action will change the `title` key in the local private state
+export const changeTitle = createAction('Change AccountInfo title', 'title')
```
jnurse: Same as above, is this action being used anywhere?

web/app/containers/account-info/container.jsx
```diff
@@ -0,0 +1,126 @@
...
+import Field from 'progressive-web-sdk/dist/components/field'
+import FieldRow from 'progressive-web-sdk/dist/components/field-row'
+import PasswordInput from 'progressive-web-sdk/dist/components/password-input'
+import Button from 'progressive-web-sdk/dist/components/button'
+import {UI_NAME} from 'progressive-web-sdk/dist/analytics/data-objects/'
+
+import {submitAccountInfoForm} from './actions'
+
+import {getAccountInfoInitialValues} from './selectors'
+// import * as accountInfoActions from './actions'
```
jnurse: 🔫

web/app/containers/account-info/container.jsx
```diff
@@ -0,0 +1,126 @@
...
+
+const AccountInfoForm = ({handleSubmit, onSubmit, error}) => {
+    return (
+        <div className="t-account-info">
+            <div className="t-account-info__headings u-padding-top-lg u-padding-bottom-lg u-padding-start-md u-padding-end-md">
+                <div className="t-account-info__breadcrumb">
+                    <Breadcrumbs items={[{text: 'Back to Dashboard', href: '/customer/account'}]} />
```
jnurse: Can we get this URL from the state instead? `/customer/account` isn't a valid URL for our salesforce instance.

web/app/containers/account-info/container.jsx
```diff
@@ -0,0 +1,126 @@
...
+                <div className="t-account-info__sub-heading">
+                    <h2 className="u-h5 u-text-family">Change name / email</h2>
+                </div>
+                <div className="t-account-info__content u-padding-md">
+                    <FieldRow>
+                        <ReduxForm.Field
+                            component={Field}
+                            label="First & Last Name"
+                            name="names"
+                            error={error && error.names}
```
jnurse: Is this line necessary? Redux form should be automatically sending the correct errors to it's field component (and the child components).

web/app/connectors/_merlins-connector/account/parsers.js
```diff
@@ -2,4 +2,22 @@
...
+
+export const parseAccountInfo = ($, $html) => {
+    const $mainContent = $html.find('.page-main')
+    const firstName = $mainContent.find('#firstname').val()
+    const lastName = $mainContent.find('#lastname').val()
+    const email = $mainContent.find('#email').val()
+
+    return {
+        title: getTextFrom($mainContent, '.page-title > span'),
+        accountFormInfo: {
```
jnurse: I wonder if the contents of accountFormInfo should go in the user branch of the store instead? So `names` and `email` would be found directly under the `user` branch, we could even store firstname and lastname separately. Then we can build out the initialValues object in the selector.

web/app/containers/account-info/actions.js
```diff
@@ -0,0 +1,64 @@
...
+    if (currentPassword && newPassword && currentPassword === newPassword) {
+        return Promise.reject(new SubmissionError({
+            _error: {
+                newPassword: 'Enter a different new password'
+            }
+        }))
+    }
+
+    return dispatch(updateAccountInfo(formValues))
+        .then(() => dispatch(updateAccountPassword(formValues)))
```
jnurse: If we're making the password fields optional, we should probably check to see if the user entered values for the password fields before attempting to update the password.

web/app/connectors/_sfcc-connector/account/commands.js
```diff
@@ -175,3 +179,57 @@ export const updateBillingAddress = (formValues) => (dispatch) => {
...
+            }
+            dispatch(recieveAccountInfoUIData((result)))
+        })
```
jnurse: We should probably check for a response fault here too.

web/app/connectors/_sfcc-connector/account/commands.js
```diff
@@ -175,3 +179,57 @@ export const updateBillingAddress = (formValues) => (dispatch) => {
...
+    const customerId = JSON.parse(sub).customer_info.customer_id
+    const requestBody = {
+        current_password: currentPassword,
+        password: newPassword
+    }
+
+    return makeApiJsonRequest(`/customers/${customerId}/password`, requestBody, {method: 'PUT'})
+        .then(checkForResponseFault)
+        .catch(() => {
+            throw new SubmissionError({_error: 'Password Change Failed'})
```
jnurse: I think we might want to look at what the fault is before just throwing an error. If it's an InvalidCustomerException then we should probably throw a different error, and handle that error differently in the action, possibly navigating the user to the login page since their login session has probably expired.

web/app/containers/account-info/container.jsx
Cdok: https://github.com/mobify/platform-scaffold/pull/891/files#diff-fefec2c8181e87339d6453c90a5bf823R18 Once the wishlist PR gets merged in we can use this selector to grab the account URL 👍

web/app/connectors/_sfcc-connector/account/commands.js
```diff
@@ -226,8 +234,28 @@ export const updateAccountPassword = ({currentPassword, newPassword}) => (dispat
         password: newPassword
     }
 
-    return makeApiJsonRequest(`/customers/${customerId}/password`, requestBody, {method: 'PUT'})
-        .then(checkForResponseFault)
+    return makeApiRequest(`/customers/${customerId}/password`, {method: 'PUT', body: JSON.stringify(requestBody)})
+        .then((res) => {
+            if (res.status >= 200 && res.status < 400) {
```
jnurse: I don't think there's a lot of support on mobile for response.status, https://developer.mozilla.org/en-US/docs/Web/API/Response/status

web/app/containers/account-info/selectors.js
```diff
@@ -3,8 +3,11 @@
...
-import {getUi} from '../../store/selectors'
+import {getUser} from '../../store/selectors'
 
-export const getAccountInfo = createSelector(getUi, ({accountInfo}) => accountInfo)
-export const getAccountInfoInitialValues = createGetSelector(getAccountInfo, 'accountFormInfo')
+export const getAccountInfoInitialValues = createSelector(getUser, (user) => {
+    return user.get('names') ? {
+        names: user.get('names'),
+        email: user.get('email')
+    } : null
```
jnurse: Can you return an empty object here instead of null?

Cdok: Discussed offline - passing an empty object makes Redux Form think that the initialValue for the form is `{}`, passing `null` ensures the initialValues are the correct `names` and `email` values

web/app/connectors/_sfcc-connector/account/commands.js
Cdok: Hmmm, what's the best way to handle this?

It's weird because `res.json()` on a successful response throws `Uncaught (in promise) SyntaxError: Unexpected end of JSON input`, so I needed to check for a successful response and resolve the Promise ahead of `res.json()`.

PR #872, opened 2017-07-10
remove this props, does not support on carousel

cole-sanderson:
Fix carousel props on PDP: `previousIcon` and `nextIcon` does not work with `<Carousel>`

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions](http://localhost:3000/potions/eye-of-newt.html)

### DIFF
--- a/amp/app/containers/product-details/partials/product-details-carousel.jsx
+++ b/amp/app/containers/product-details/partials/product-details-carousel.jsx
@@ -9,15 +9,11 @@ import Carousel from '../../../components/carousel'
 // Selectors
 import {getProductImages} from 'progressive-web-sdk/dist/store/products/selectors'
 
-import {staticURL} from '../../../utils'
-
 const ProductDetailsCarousel = ({images}) => {
 
     return (
         <Carousel
             id="product-details-carousel"
-            previousIcon={staticURL('chevron-left')}
-            nextIcon={staticURL('chevron-right')}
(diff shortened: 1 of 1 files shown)


amp/app/containers/product-details/partials/product-details-carousel.jsx
```diff
@@ -9,15 +9,11 @@ import Carousel from '../../../components/carousel'
...
 import {getProductImages} from 'progressive-web-sdk/dist/store/products/selectors'
 
-import {staticURL} from '../../../utils'
-
 const ProductDetailsCarousel = ({images}) => {
 
     return (
         <Carousel
             id="product-details-carousel"
-            previousIcon={staticURL('chevron-left')}
```
kschmidtdev: I see - but these assets are still loading (except they fail to load on https://amp.merlinspotions.com/eye-of-newt.html) I guess through css?

Also maybe take a look at `amp-carousel-button-prev` in this PR, I think it should be updated to `a-`?

cole-sanderson: We cannot add image to markup for AMP Carousel controls, it needs to be done with CSS. `amp-carousel-button-prev` is AMP built in class, it's something that we cannot control.

kschmidtdev: Ok, makes sense

PR #875, opened 2017-07-12
Assert subscribed only if push messaging is enabled

ellenmobify:
## Changes
- Wait for product listing items to be present on the page before trying to find the default ask
- Allow asynchronous scripts to run for 2 seconds before Selenium gives a timeout error http://nightwatchjs.org/api/timeoutsAsyncScript.html. If the test fails, you will see that `window.Progressive.MessagingClient.LocalStorage.get('mobifyMessagingClientSubscriptionStatus')` is `null` instead of `[object Object]`

## How to test-drive this PR
- CircleCI should not be affected https://circleci.com/gh/mobify/platform-scaffold/tree/e2e-subscribe-windows
- No failures on https://ci.appveyor.com/project/mobify/platform-scaffold/build/1.0.1311


web/tests/system/workflows/push-subscribe.js
```diff
@@ -43,6 +45,5 @@ export default {
         // by this point.
         pushMessaging
             .acceptDefaultAsk()
-            .assertSubscribed()
```
fractaltheory: Is there any way to detect whether we're in an AppVeyor instance? I'd really like to keep this assertion around for a CircleCI environment. Otherwise, we lose the purpose of the test, don't we?

ellenmobify: The assertion still runs on CircleCI, it just gets called inside `pushMessaging.acceptDefaultAsk` if it find that the default ask is present.

fractaltheory: Right but ... I think we want to test that the default ask is present too. Doesn't this change mean that if it wasn't present, it won't assert anything and thus will never show a failure?

ellenmobify: Yes, you are correct. It won't assert anything if the default ask is not present, as is the case with the AppVeyor environment right now. Should we spend more time investigating why the default ask is not present in AppVeyor?

fractaltheory: It would be nice to understand why it isn't working in the AppVeyor context, since it was working on our Windows machine just fine. If we don't have time to investigate that, then we need to disable it specifically for AppVeyor - because I'd still like to assert both the existence of the UI as well as successful subscription (at least in CircleCI)

ellenmobify:
@fractaltheory Looks like the issue here was just a case of the page loading a bit slower than on CircleCI, so the check for the default ask happened before the the default ask could load. The test 
will now wait until the PLP item has loaded before it looks for the default ask.

fractaltheory:
@ellenmobify awesome, thanks for looking into this. LGTM! 👍

PR #879, opened 2017-07-12
Skip the rest of the tests if a product is sold out

ellenmobify:
## Changes
- Keep track of whether the product is in stock or not
- Condition the rest of the tests (after adding item to cart) on whether the product is in stock

## How to test-drive this PR
- Rebuild https://circleci.com/gh/mobify/platform-scaffold/tree/e2e-sold-out
- If testing the sold out case, set `PRODUCT_LIST_INDEX` and `PRODUCT_INDEX` to 2

### DIFF
--- a/web/tests/system/workflows/guest-checkout.js
+++ b/web/tests/system/workflows/guest-checkout.js
@@ -18,7 +18,7 @@ let checkout
 let pushMessaging
 
 const PRODUCT_LIST_INDEX = process.env.PRODUCT_LIST_INDEX || 2
-const PRODUCT_INDEX = process.env.PRODUCT_INDEX || 1
+const PRODUCT_INDEX = process.env.PRODUCT_INDEX || 2
 const ENV = process.env.NODE_ENV || 'test'
 
 export default {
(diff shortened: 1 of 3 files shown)


web/tests/system/workflows/guest-checkout.js
```diff
@@ -69,47 +69,56 @@ export default {
...
-        browser
-            .waitForElementVisible(cart.selectors.cartTemplateIdentifier)
-            .assert.visible(cart.selectors.cartTemplateIdentifier)
+        if (productDetails.inStock) {
+            productDetails.navigateToCart()
+            browser
+                .waitForElementVisible(cart.selectors.cartTemplateIdentifier)
+                .assert.visible(cart.selectors.cartTemplateIdentifier)
+        } else {
+            browser.log('Item is out of stock.')
```
jasecode: Should we add an additional log something like (Please change PRODUCT_INDEX) number to promote that e2e tests get tested properly?
We can also radically Fail the test if the product is out of stock with better test verbose. That way we don't have to put the if statements on the other following commands.

web/tests/system/workflows/registered-checkout.js
```diff
@@ -70,52 +70,61 @@ export default {
...
-        productDetails.navigateToCart()
-        browser
-            .waitForElementVisible(cart.selectors.cartTemplateIdentifier)
-            .assert.visible(cart.selectors.cartTemplateIdentifier)
+        if (productDetails.inStock) {
+            productDetails.navigateToCart()
+            browser
+                .waitForElementVisible(cart.selectors.cartTemplateIdentifier)
+                .assert.visible(cart.selectors.cartTemplateIdentifier)
+        }
```
jasecode: Missing a browser.log Here - same as the one in the guest-checkout.js

jasecode:
👍

PR #886, opened 2017-07-13
add account address

gsaray:
Add account address page 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1430

## Changes
- (change1)

## Todos

## How to test-drive this PR
- Run `npm run dev`
- Open menu and sign in
- login with mobifyqa@gmail.com
- go to address book


web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,106 @@
...
+                        title="Change Address"
+                        className="u-width-full u-color-brand"
+                        icon="edit"
+                        showIconText={true}
+                        iconClassName="u-margin-end"
+                    />
+                }
+            />
+        </div>
+        <div className="t-account-address__section u-bg-color-neutral-10 u-padding-top-lg u-padding-bottom-md u-padding-start-md u-padding-end-md u-border-top u-border-bottom">
```
jeffkamo: Same here, the padding classes can be simplified to `u-padding-top-lg u-padding-md`

web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,108 @@
...
+            <h2 className="u-h5 u-text-family">Other saved addresses</h2>
+        </div>
+        <div className="t-account-address__content u-padding-md">
```
jeffkamo: If there is more than one "other address", does there need to be any change to how the spacing works?

web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,108 @@
...
+                }
+                children={
+                    <div className="u-padding-md">Address 1</div>
+                }
+                footer={
+                    <div className="u-flexbox">
+                        <div className="u-flex u-border-end">
+                            <Button
+                                type="button"
+                                title="Edit"
```
jeffkamo: Is there meant to be a way to assign an address as the _default_ address? Or is that part of this edit button?

web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,106 @@
...
+            </div>
+            <div className="u-margin-top-md">
+                <h1 className="t-account-info__title u-text-uppercase u-width-1of2">Address Book</h1>
+            </div>
+            <Button
+                text="Add new address"
+                className="pw--tertiary u-margin-top-lg u-width-full u-text-weight-medium"
+            />
+        </div>
+        <div className="t-account-address__section u-bg-color-neutral-10 u-padding-top-lg u-padding-bottom-md u-padding-start-md u-padding-end-md u-border-top u-border-bottom">
```
jeffkamo: Hmm. Or, arguably, this might be a good time to use the `t-account-address__section` class to make these styles more transportable. How many of these sections are there likely to be?

Also a minor thing: is `section` the best name here? Considering the semantic meaning of the `<section>` element, I'm not sure this element means the same thing. Looks more like a heading for a section or separator, rather than the section itself. Thoughts?

web/app/components/card/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+ * Card component is used to show content in a card with header and footer
+ */
+
+const Card = ({
+    className,
+    header,
+    children,
+    footer
+}) => {
+    const classes = classNames('c-card', className)
```
jeffkamo: This might be a nice time to trial a modifier pattern we could eventually use for Button. I propose something like this:

```
const classes = classNames('c-card', {
    'c--shadow': props.shadow,
    'c--border': props.border,
}, className)
```

And usage would look like this:

```
<Card shadow />
<Card border />
```

Thoughts?

web/app/components/card/_theme.scss
```diff
@@ -1,7 +1,14 @@
 // Card
 // ===
 
-.u-card {
+.c-card {
```
jeffkamo: wow, Git really figure this one out, huh? 👍

web/app/containers/account-address/container.jsx
gsaray: I removed utility classes and applied the style to `t-account-address__sub-heading`

web/app/containers/account-address/container.jsx
gsaray: This is now `t-account-address__sub-heading` .

web/app/containers/account-address/container.jsx
gsaray: there wouldn't be any change here. I'm adding a margin top to the second card when there is more than 1 card.

web/app/containers/account-address/container.jsx
gsaray: I don't think that's in scope.

web/app/connectors/_merlins-connector/account/commands.js
```diff
@@ -39,6 +40,13 @@ export const initAccountDashboardPage = (url) => (dispatch) => { // eslint-disab
     return Promise.resolve()
 }
 
+export const initAccountAddressPage = (url) => (dispatch) => { // eslint-disable-line
+    return dispatch(fetchPageData(url))
+        .then(([$, $response]) => {
+            dispatch(recieveAccountAddressUIData(accountAddressParser($, $response)))
```
jnurse: I don't think this address data should go into the UI branch of the store. It should either go into the checkout branch where we've already defined some keys like `storedAddresses` and `defaultShippingAddressId` or it should go into the user branch of the store (and we should look into moving the addresses in the checkout branch there too).

web/app/connectors/_merlins-connector/account/parsers.js
```diff
@@ -3,3 +3,36 @@
...
+    const firstname = text[0].split(' ')[0]
+    const lastname = text[0].split(' ')[1]
+    const city = text[2].split(',')[0] ? text[2].split(',')[0].trim() : ''
+    const regionId = text[2].split(',')[1] ? text[2].split(',')[1].trim() : ''
+    const postcode = text[2].split(',')[2] ? text[2].split(',')[2].trim() : ''
+
+    return {
+        firstname,
+        lastname,
+        addressLine1: text[1].trim(),
```
jnurse: What about addressLine2?

web/app/connectors/_sfcc-connector/account/commands.js
```diff
@@ -175,3 +175,45 @@ export const updateBillingAddress = (formValues) => (dispatch) => {
...
+                            preferred,
+                            country_code
+                        }) => {
+
+                            return {
+                                firstname: first_name,
+                                lastname: last_name,
+                                telephone: phone,
+                                postcode: postal_code,
+                                addressLine1: address1,
```
jnurse: Same as above.

web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,147 @@
...
+        {firstname ?
+            <p>{firstname} {lastname}</p>
+            :
+            <SkeletonText width="50%" style={{lineHeight: '20px'}} />
+        }
+        {addressLine1 ?
+            <p>{addressLine1}</p>
+            :
+            <SkeletonText width="60%" style={{lineHeight: '20px'}} />
+        }
```
jnurse: We're missing addressLine2 here too.

web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,147 @@
...
+            <p>{telephone}</p>
+            :
+            <SkeletonText width="50%" style={{lineHeight: '20px'}} />
+        }
+    </div>)
+    return (
+        <div className="t-account-address">
+            <div className="t-account-address__heading u-padding-top-lg u-padding-bottom-lg u-padding-start-md u-padding-end-md">
+                <div className="t-account-address__breadcrumb">
+                    <Breadcrumbs items={[{text: 'Back to Dashboard', href: '/customer/account'}]} />
```
jnurse: Can you get this URL for the store instead?

web/app/containers/account-address/container.jsx
```diff
@@ -0,0 +1,147 @@
...
+        lastname,
+        addressLine1,
+        city,
+        countryId,
+        postcode,
+        telephone,
+        regionId
+    } = defaultAddress
+
+    const addressBlock =
```
jnurse: Could we make this a component instead? Or a partial?

web/app/router.jsx
```diff
@@ -77,6 +79,7 @@ const Router = ({store}) => (
                 <Route component={Login} path="customer/account/create/" routeName="register" fetchAction={initPage(initRegisterPage)} />
 
                 <Route component={AccountDashboard} path="customer/account" routeName="account" fetchAction={initPage(initAccountDashboardPage)} />
+                <Route component={AccountAddress} path="customer/address" routeName="accountAddress" fetchAction={initPage(initAccountAddressPage)} />
```
jnurse: Can you add a route for the SFCC version of the page? We'd want the path prop to be something like this: `*/Address-List`

jnurse: Can you add a route for the SFCC version of the page? We'd want the path prop to be something like this: `*/Address-List`?

jnurse:
When I preview with the SFCC connector all my addresses have the same data as my default. 

screen shot 2017-07-18 at 4 50 38 pm

web/app/connectors/_merlins-connector/account/commands.js
gsaray: moved address data into the user branch

web/app/containers/account-address/container.jsx
gsaray: I'll fix this when view wishlist is merged so I can use getAccountURL

gsaray:
oh, we are going to only show one default address

web/app/store/user/reducer.js
```diff
@@ -6,13 +6,14 @@ import Immutable from 'immutable'
...
 import {setLoggedIn, receiveUserCustomContent} from 'progressive-web-sdk/dist/integration-manager/results'
-import {receiveAccountInfoData, receiveWishlistData} from 'progressive-web-sdk/dist/integration-manager/account/results'
+import {receiveAccountInfoData, receiveAccountAddressData, receiveWishlistData} from 'progressive-web-sdk/dist/integration-manager/account/results.js'
 
 const initialState = Immutable.Map()
 
 const userReducer = handleActions({
     [setLoggedIn]: mergePayload,
     [receiveUserCustomContent]: mergePayload,
+    [receiveAccountAddressData]: mergePayload,
```
jnurse: Discussed offline, we're going to look at moving all saved address data into the user branch of the redux store and remove any saved address data from the checkout branch in another PR. For now, we'll merge this work so we can pull it into the SFCC connector move branch, then we can look at making the schema changes after.

PR #896, opened 2017-07-20
Check TTI

ellenmobify:
## Changes
- Verify that Time to interactive/First interactive is under a threshold

## How to test-drive this PR
- `npm run dev`
- `npm run test:pwa-preview`

### DIFF
--- a/web/tests/performance/lighthouse/check-lighthouse-score.js
+++ b/web/tests/performance/lighthouse/check-lighthouse-score.js
@@ -41,12 +41,18 @@ const checkLighthouse = function(jsonResults) {
 }
 
 /**
-* Display some important performance metrics.
+* Verify time to interactive/first interactive
 */
 const checkTTI = function(jsonResults) {
-    console.log(`Time to first interactive: ${jsonResults.audits['first-interactive'].displayValue}`)
-    console.log(`Analyzing total bundle size...`)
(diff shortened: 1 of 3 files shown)


web/tests/performance/lighthouse/check-lighthouse-score.js
```diff
@@ -41,12 +41,18 @@ const checkLighthouse = function(jsonResults) {
...
-* Display some important performance metrics.
+* Verify time to interactive/first interactive
 */
 const checkTTI = function(jsonResults) {
-    console.log(`Time to first interactive: ${jsonResults.audits['first-interactive'].displayValue}`)
-    console.log(`Analyzing total bundle size...`)
-    console.log(`${jsonResults.audits['total-byte-weight'].displayValue}`)
+    const actualValue = jsonResults.audits['first-interactive'].rawValue
+
+    if (actualValue > 10000) {
```
jasecode: Thinking if whether we should have this configurable in the package.json as well along with the minimum lighthouse score. 🤔  During development for partners , I'd imagine this would fail quite frequently and it'd be a bottleneck.

ellenmobify: Ya totally. We should also figure out how to communicate the ideal threshold (eg 10s) for TTI and Lighthouse score so that these values improve over time.

PR #898, opened 2017-07-20
AMP-89: pagination

cole-sanderson:
**JIRA**: https://mobify.atlassian.net/browse/AMP-89

## Changes
- `<Pagination>` component 

## Todos
- [x] Waiting for PWA team to merge Pagination so can look if we can reuse any of parsers from PWA

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions PLP](http://localhost:3000/potions.html)
- Scroll down to bottom to check Pagination


web/app/store/categories/selectors.js
```diff
@@ -28,6 +28,7 @@ export const getCategoryProductIds = createGetSelector(getSelectedCategory, 'pro
 export const getCategoryItemCount = createGetSelector(getSelectedCategory, 'itemCount')
 export const getCategoryTitle = createGetSelector(getCategory, 'title')
 export const getCategoryParentID = createGetSelector(getCategory, 'parentId', null)
+export const getPagination = createGetSelector(getCategory, 'pagination')
```
kschmidtdev: Is this code from the web team? Did they not add something suitable? I'm concerned about deviating from the integration manager, because if we switch to the salesforce connector, this might break.

web/app/connectors/_merlins-connector/categories/commands.js
```diff
@@ -25,6 +25,7 @@ export const initProductListPage = (url) => (dispatch) => {
     const filter = decodeURIComponent(extractFilter(url))
     // Merlins uses 'product_list_order' as URL search key
     const rawURL = url
+
```
kschmidtdev: I think the linting is complaining about a trailing space in this file, that's why circle is failing

web/app/store/categories/selectors.js
cole-sanderson: Yeah, there was nothing that was suitable for what we need on AMP. We only need to access to pagination href and current page for pagination on AMP. If we make switch to salesforce connector and will need to add something in there?

web/app/connectors/_merlins-connector/categories/commands.js
cole-sanderson: I figured out why CircleCI was failing because the `pageToShow` is undefined. I fixed the issue but not sure if it's good idea, can you please have another look.

cole-sanderson: I have removed the empty space and still getting the same error, I am stuck with fixing the error. Do you have any idea how to fix it?

olibrook:
Code itself looks good, but I'm going to hold off a proper review until I can run the tests locally.

I think @kschmidtdev raises a good point - can we get the pagination data from the integration-manager somehow, instead of parsing it out separately?

amp/app/containers/product-list/partials/product-list-contents.jsx
```diff
@@ -135,6 +139,28 @@ const ProductListContents = (props) => {
                 :
                     <NoResultsList routeName={routeName} />
                 }
+
+                {pageCount > 1 &&
+                    <Pagination
+                        className="u-margin-top-lg"
+                        pageCount={pageCount}
+                        currentPage={pagination.current.number}
```
kschmidtdev: I looked at the PWA and it seems to just look at the `p` query object of the URL for what the current page is, and sets it on the outgoing URLs accordingly...I think we could do the same here and avoid adding to the merlins parser. I'm concerned about diverging too much from how the PWA is doing pagination, and if we proceed with this then we'd have to do extra work for the SFCC connector.

cole-sanderson:
:wooooo: thank you @kschmidtdev for your help 👍

PR #905, opened 2017-07-24
Run e2e on all connectors

jasecode:
Lets Run End to End tests on all Connectors: merlins, sfcc, stub.

## Changes
- ./scripts/change-connector.sh , script takes in two arguments and changes the first argument to the second argument
- refactored our e2e tests - tagging them and making them into groups according to connectors
- config.yml , adds additional jobs that runs e2e on separate connectors

## How to test-drive this PR
- Run `npm start`
- bash ./scripts/change-connector.sh merlins sfcc
- Run npm run test:e2e -- --tag sfcc --group workflows/sfcc
- - bash ./scripts/change-connector.sh sfcc stub
- Run npm run test:e2e -- --tag stub --group workflows/stub


web/scripts/change-connector.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Change connector from the first Argument to the second argument
+# eg: ./change-connector.sh merlins sfcc
```
jeremywiebe: It'd be cool if we didn't need to know the current connector to change.  We should be able to accomplish that by just commenting out all `initConnector` lines and then uncommenting the one we want.

jasecode: Having a bit of trouble doing that with the `sed` command. Could use a bit of a walk to the right direction in the logic for this when you have time. 🤔

jasecode: We're leaving it as it is now. Will address a better implementation of this in a future PR.

PR #912, opened 2017-07-27
[ACCOUNT] View Orders

gsaray:
# View Orders Scaffold Page

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1433
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/788

## Changes
- Implements the View Orders page on the scaffold account dashboard

## Todos

## How to test-drive this PR
- `npm link` to the linked SDK PR
- Test on both connectors
- Run `npm run dev`
- Login and enter the account page, checking the Order page to test functionality


web/app/connectors/_sfcc-connector/parsers.js
```diff
@@ -256,3 +256,26 @@ export const parseFilterOptions = (refinements) => {
...
+    data.forEach(({
+        order_no,
+        confirmation_status,
+        creation_date,
+        customer_info,
+        order_total
+    }) => {
+        ordersMap[order_no] = {
+            orderNumber: order_no,
+            date: creation_date,
```
jnurse: Can we convert this to a readable string? 

screen shot 2017-08-15 at 1 18 31 pm

web/app/connectors/_sfcc-connector/parsers.js
```diff
@@ -256,3 +256,26 @@ export const parseFilterOptions = (refinements) => {
...
+            shippingAddress: {
+                fullName: customer_info.customer_name
+            },
+            total: order_total.toString(),
```
jnurse: Can you use the formatPrice util here to add the `$` and turn it into a string, rather than using `.toString()`?

web/app/containers/account-order-list/container.jsx
```diff
@@ -0,0 +1,148 @@
...
+import Button from 'progressive-web-sdk/dist/components/button'
+import Icon from 'progressive-web-sdk/dist/components/icon'
+import {UI_NAME} from 'progressive-web-sdk/dist/analytics/data-objects/'
+import {getOrderList, getOrdersPage, getNumOrderPages} from './selectors'
+import {reorderItems, setOrderListPage} from './actions'
+import OrderBlock from './partials/account-order-block'
+import SkeletonText from 'progressive-web-sdk/dist/components/skeleton-text'
+import Pagination from 'progressive-web-sdk/dist/components/pagination'
+
+const NoOrder = ({dashboardURL}) => (
```
Cdok: Would there be a benefit to connecting the component vs. passing dashboard URL as a prop from the orderlist?

Cdok:
@jnurse Ah SFCC re-order was making the request but wasn't fetching the cart / basket data after. Good catch, should be fixed now :eyes:

web/app/containers/account-order-list/container.jsx
jnurse: I think the main benefit would be consistency and cleanliness. In this case it's just one prop, so passing it down isn't a big deal, but in general we've been connecting components when possible to avoid getting into situations were we have large mapStateToProps and mapDispatchToProps functions because we're passing everything down to other components. 

This is definitely a non-blocking suggestion, so if you'd like to leave it as is I'm fine with that.

Cdok: Yeah I definitely think if we were passing a lot of state or actions through it would clean things up. I was just looking at `NoAddress` and `MiniCart` and since they're mostly static components as well they aren't being connected.

I think since the `NoOrder` component is just extracting markup out of the parent component for readability, I'd be okay with leaving it un-connected

web/app/connectors/_sfcc-connector/parsers.js
```diff
@@ -256,3 +256,26 @@ export const parseFilterOptions = (refinements) => {
...
+    data.forEach(({
+        order_no,
+        confirmation_status,
+        creation_date,
+        customer_info,
+        order_total
+    }) => {
+        ordersMap[order_no] = {
+            orderNumber: order_no,
+            date: creation_date ? creation_date.split('T')[0] : '', // eslint-disable-line
```
jnurse: Could we do something like this instead:
```
new Date(creation_date).toLocaleDateString()
```

PR #925, opened 2017-08-03
Adds loader.js support for AMP pages for the new V8 tag

therafatm:
**JIRA**: https://mobify.atlassian.net/browse/AMP-103

## Reviewers
@kschmidtdev @olibrook 

## Changes
- Adds new AMP urls constant file for merlins amp routes
- Adds logic to loader.js to include amp links when appropriate

## Todos
- [ ] Get 👍 

## How to test-drive this PR
- Have the AMP-Validator chrome plugin installed
- `cd web/ && npm i && npm run dev`
- Set mobile emulation and preview Merlins and browse to the paths which have an amp page, i.e. `/potions, /books, etc`
- Verify that the validator plugin turns blue and that it reroutes you to the correct amp pages when you click on it


web/app/ampUrls.js
```diff
@@ -0,0 +1,11 @@
+export const baseAMPUrl = 'https://amp.merlinspotions.com'
+
+export const validAMPUrls = [
+    // PLPs
```
kschmidtdev: What about PDPs? 😄

web/app/loader.js
```diff
@@ -200,6 +202,17 @@ const waitForBody = () => {
     return waitForBodyPromise
 }
 
+const hasAMPPage = (validUrlList, path) => {
+    for (let x = 0; x < validUrlList.length; x++) {
```
kschmidtdev: In ES6 we tend to write this as `validUrlList.forEach((url) => {`

web/app/loader.js
```diff
@@ -200,6 +202,17 @@ const waitForBody = () => {
...
+        const url = validUrlList[x]
+        const exp = new RegExp(`.*${url}.*`)
+        if (exp.test(path)) {
```
kschmidtdev: You could probably flatten this into `/`.*${url}.*`/.test(path)`
But I think that regex is a little too loose. It could match things we don't intend to. Maybe just treat all the entries as regex data? 
Also currently if a url matches, then the pathname === the validUrlList entry. So there is no need to return it. This could boil down to:
`return new RegExp(url).test(path)`

web/app/loader.js
```diff
@@ -244,6 +257,15 @@ const loadPWA = () => {
         content: 'utf-8'
     })
 
+    // Only add AMP tag for specified URLs
+    const ampPath = hasAMPPage(validAMPUrls, window.location.pathname)
+    if (ampPath !== false) {
```
kschmidtdev: This could just be `if (ampPath) {`

web/app/loader.js
```diff
@@ -244,6 +257,15 @@ const loadPWA = () => {
...
+        loadAsset('link', {
+            rel: 'amphtml',
+            href: `${baseAMPUrl}${ampPath}`
```
kschmidtdev: If we use a wildcard regex to match for PDPs like the PWA and AMP does, then ${ampPath} won't be useful here - `window.location.pathname` should work?

web/app/ampUrls.js
therafatm: Did we not choose to leave out the PDP's as users wouldn't get a response for a url like `'/checkout/cart/configure/id/*/product_id/*/'` from a google search (according to our conversation yesterday), and therefore no AMP page? Also current requests to a PDP page from an amp page seems to be directed to the standard PWA page. Am I missing something here?

web/app/loader.js
therafatm: I intentionally kept this as a standard for loop in order to be able to return the first match, ideally we'd use `Array.some()` but that just returns true/false....

web/app/ampUrls.js
kschmidtdev: Those are just for PDPs served from within the checkout, but this is the rule for matching PDPs (like `eye-of-newt.html`, `unicorn-blood.html`, etc)
`app.get('*.html', handlePage(ProductDetails))` (it runs last, which is important)

The PLP links to PWA PDPs, but we definitely support AMP PDPs that users will find via Google Search. But they will be something like `https://www.merlinspotions.com/eye-of-newt.html`

web/app/loader.js
kschmidtdev: That makes sense! Based on my other feedback `Array.some()` should do the trick then.

web/app/loader.js
therafatm: I agree. I wasn't sure of what to accept and what not to accept, particularly with the possibility of query parameters and so on. Maybe `/{url}.html.*` would be more appropriate? Or if we are certain that we only want exact matches, we can just do `return new RegExp(url).test(path)` as you suggested 😄

web/app/loader.js
therafatm: This would again depend on what `hasAMPPage` returns ....

web/app/ampUrls.js
therafatm: OOOOOOOH. Cool I'll add that.

web/app/ampUrls.js
```diff
@@ -0,0 +1,30 @@
...
+    '/aging-potion-2.html',
+    '/potions/aging-potion-2.html',
+    '/beginners-guide-to-transfiguration.html',
+    '/books/beginners-guide-to-transfiguration.html',
+    '/dragon-breeding-for-pleasure-and-profit.html',
+    '/books/dragon-breeding-for-pleasure-and-profit.html',
+    '/beginner-s-guide-to-transfiguration-2.html',
+    '/books/beginner-s-guide-to-transfiguration-2.html',
+    '/books/necronomicon.html',
+    '/necronomicon.html'
```
kschmidtdev: Note for our future selves - eventually we'd like the PDP urls to be a * regex like the AMP and PWA routers use, but first the AMP service needs to learn how to deal with PDPs that are not valid, example:
Navigate to https://www.merlinspotions.com/doesntexist.html <-- loader.js would say this is a PDP, and the AMP service would think https://amp.merlinspotions.com/doesntexist.html is one too, but has an error in rendering. It needs to be updated to deal with this and serve an appropriate response.

web/app/loader.js
```diff
@@ -335,6 +339,15 @@ const loadPWA = () => {
         content: 'utf-8'
     })
 
+    // Only add AMP tag for specified URLs
+    const ampPath = hasAMPPage(validAMPUrls, window.location.pathname)
```
therafatm: Sure. That works.

kschmidtdev:
Oops, preview for desktop does work 😊

web/app/loader.js
```diff
@@ -287,10 +289,22 @@ const waitForBody = () => {
     return waitForBodyPromise
 }
 
-/**
```
kschmidtdev: Was this comment block removal intentional?

therafatm: Nope. Adding it back in. Good catch!

web/app/loader.js
```diff
@@ -287,10 +289,26 @@ const waitForBody = () => {
...
+    if (hasAMPPage(validAMPUrls, window.location.pathname)) {
+        loadAsset('link', {
+            rel: 'amphtml',
+            href: `${baseAMPUrl}${window.location.pathname}`
+        })
+    }
+}
+
 /**
- * Initialize the app. Assumes that all needed polyfills have been
```
kschmidtdev: Sorry to be so pedantic but can you re-add the spaces here? I want to minimize the diffs from `develop` to make merging as easy as possible

therafatm: Sure 😅

PR #926, opened 2017-08-03
Performance metrics

wizardlyhel:
Add performance metrics instrumentation

 **JIRA**: https://mobify.atlassian.net/browse/MTT-669
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/795

## Changes
- Add performance metric instrumentation

## Todos
- [x] Add a high-level description of your changes to the CHANGELOG.md, including a link to the PR with your changes

## How to test-drive this PR
```
npm link progressive-web-sdk
npm run dev
```
Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)


web/app/loader.js
```diff
@@ -34,13 +34,50 @@ const messagingEnabled = MESSAGING_ENABLED  // replaced at build time
...
         enabled: messagingEnabled
+    },
+    PerformanceTiming: {
+        pageStart: navigationStart,
+        mobifyStart,
+        timingStart
     }
 }
 
+// Track First Paint and First Contentful Paint
```
benlast: In nonPWA mode, the `loader.js` will be loaded on every page - will this screw up the results? Do you need PWA/nonPWA mode as a dimension maybe?

wizardlyhel: Okay. Although we are collecting, we are not sending in nonPWA mode

wizardlyhel: I'll look into if I can move the code to a point where we know we are going to load PWA. I know FP and FCP will not be tracked if it is initialized too late

wizardlyhel:
@benlast updated loader.js - We shouldn't be tracking these metrics in nonPWA mode

web/app/loader.js
```diff
@@ -46,9 +50,71 @@ window.Progressive = {
...
+                    category: 'timing'
+                },
+                dimensions: {
+                    page_start: navigationStart,
+                    mobify_start: mobifyStart,
+                    first_paint: timings.firstPaint,
+                    first_contentful_paint: timings.firstContentfulPaint,
+                    app_start: timings.appStart,
+                    timing_start: timingStart,
+                    page_contentful_paint: (navigationStart && performance.timing.domContentLoadedEventEnd) - timingStart,
```
sketchytechky: Just talking out loud - There's a chance that `timingStart` can be `undefined` (in strange case) - in that case these number will be `NaN` - data pipeline can handle that.

sketchytechky: We are also assuming `window.performance` is always there - which I think we need to be careful about?

web/app/loader.js
```diff
@@ -46,9 +50,71 @@ window.Progressive = {
...
+        }
+        ttiObserver.o = new window.PerformanceObserver((list) => {
+            ttiObserver.e = ttiObserver.e.concat(list.getEntries())
+        })
+        ttiObserver.o.observe({entryTypes: ['longtask']})
     }
 }
 
+const triggerNonPWAPerformanceEvent = (tracker) => {
+    window.addEventListener('load', () => {
```
sketchytechky: This will break on old IE - just saying 😛  http://caniuse.com/#search=addEventListener

web/app/loader.js
wizardlyhel: timingStart can only be undefined if mobifyStart is undefined. mobifyStart is instrumented in the tag itself so if that is missing, it would be a tag issue

I'll update the check for `window.performance`

web/app/template.jsx
```diff
@@ -34,10 +37,12 @@ const template = (WrappedComponent) => {
         }
 
         componentWillMount() {
+            trackPerformance(PERFORMANCE_METRICS.templateWillMount)
```
wizardlyhel: mmmm... let me think on this

wizardlyhel: Good catch tho

web/app/loader.js
```diff
@@ -34,6 +34,17 @@ const messagingEnabled = MESSAGING_ENABLED  // replaced at build time
 const CAPTURING_CDN = '//cdn.mobify.com/capturejs/capture-latest.min.js'
 const ASTRO_CLIENT_CDN = `//assets.mobify.com/astro/astro-client-${ASTRO_VERSION}.min.js`
 
+const getPerformanceTiming = (type, timeDiff = 0, defaultValue) => {
+    if (window.performance && performance.timing && performance.timing[type]) {
+        return window.performance.timing[type] - timeDiff
+    }
+    return defaultValue || undefined
```
sketchytechky: nitpick - I think the `|| undefined` is not necessary.
```
> testFunc = (a, b=0, c) => { return c; }
> testFunc(1)
undefined
```

In fact, this is harmful (e.g. if defaultValue is 0, this will return undefined).

wizardlyhel: right

sketchytechky: nitpick - I think the `|| undefined` is not necessary.
```
> testFunc = (a, b=0, c) => { return c; }
> testFunc(1)
undefined
```

In fact, this is harmful (e.g. if defaultValue is 0, this will return undefined).

wizardlyhel:
Added one more metrics: `is_saved_page`

This metric is a boolean and will indicate if the page was a saved page from the redux store. This is distinguish a user that sees the skeleton page verse a user that sees the existing data from store

web/app/loader.js
```diff
@@ -46,6 +57,78 @@ window.Progressive = {
...
+    },
+    PerformanceTiming: {
+        pageStart: navigationStart,
+        mobifyStart,
+        timingStart
+    }
+}
+
+// Track First Paint and First Contentful Paint for PWA and non-PWA
+if ('PerformanceObserver' in window) {
```
jeremywiebe: Will this work if we need to polyfill `PerformanceObserver`?

web/app/loader.js
```diff
@@ -46,6 +57,78 @@ window.Progressive = {
...
+                timing_start: timingStart,
+                page_contentful_paint: getPerformanceTiming('domContentLoadedEventEnd', timingStart, 'null'),
+                page_content_load: getPerformanceTiming('loadEventEnd', timingStart, 'null')
+            }
+        })
+    }, 0)
+}
+
+const triggerNonPWAPerformanceEvent = (tracker) => {
+    if (window.addEventListener) {
```
jeremywiebe: On line 69 we checked for the existence of an API using `if ('PerformanceObserver' in window)`... is this the same type of check? Just thinking about consistency.

web/app/router.jsx
```diff
@@ -68,10 +70,16 @@ if (isRunningInAstro) {
     }
 }
 
-const initPage = (initAction) => (url, routeName) => (dispatch) => {
+const initPage = (initAction) => (url, routeName) => (dispatch, getState) => {
     return dispatch(initAction(url, routeName))
-        .then(() => dispatch(setFetchedPage(url)))
-        .then(() => dispatch(onPageReady(routeName)))
+        .then(() => {
+            trackPerformance(PERFORMANCE_METRICS.isSavedPage, hasFetchedCurrentPath(getState()) ? 'true' : 'false')
```
jeremywiebe: Doesn't `hasFetchedCurrentPath()` return a boolean value?  The ternary made me double-check. Or are you catching an undefined return value if the path hasn't already been fetched?

web/app/loader.js
wizardlyhel: there's a polyfill for `PerformanceObserver`?

web/app/router.jsx
wizardlyhel: Ahh I wanted to be passing string 'true/false'

web/app/loader.js
jeremywiebe: :facepalm: I saw the `tti-polyfill` and thought it also polyfilled `PerformanceObserver`!

PR #928, opened 2017-08-03
Resolve account address country select not populating

Cdok:
"Also, on both Magento/SFCC the countries aren't loading. The drop down is empty and doesn't contain any options inside"

In SFCC we only support United States, Merlin's wasn't populating when there wasn't an item in the cart.

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1432
 **Linked PRs**: https://github.com/mobify/platform-scaffold/pull/892

## Changes
- Instead of fetching and parsing cart page, we are instead parsing a new set of data from the `/address/new` page

## Todos

## How to test-drive this PR
- Run `npm run dev`
- Go to account address page, ensuring that the Country Select is populating correctly when adding or editing a new address


web/app/connectors/_merlins-connector/account/parsers.js
```diff
@@ -47,3 +47,47 @@ export const parseWishlistProducts = ($, $response) => {
...
+    const optionalZipList = magentoResponse.getIn(['#country', 'regionUpdater', 'countriesWithOptionalZip'])
+    const regionRequiredMap = {}
+    const postCodeOptionalMap = {}
+
+    optionalZipList.toJS().forEach((countryId) => {
+        postCodeOptionalMap[countryId] = true
+    })
+
+    const regions = []
+    hasRegionDropdown.toJS().config.regions_required.forEach((country) => {
```
jnurse: Rather than converting this to JS twice, can you store the JS version in a variable instead of the map?

web/app/connectors/_merlins-connector/account/parsers.js
```diff
@@ -47,3 +47,47 @@ export const parseWishlistProducts = ($, $response) => {
...
+                id: region,
+                label: countryRegions[region].name
+            })
+        })
+    })
+
+    const countries = []
+
+    $response.find('select#country option').each((_, option) => {
+        const $option = $(option)
```
jnurse: We're not doing much with this option, could we skip converting it to a jQuery object and just get the value (option.value) and text (option.textContent) using plain javascript? 

It would mean we also don't need to pass `$` into this function.

PR #938, opened 2017-08-10
Force-load in preview, with PWA/non-PWA modes

benlast:
#The V8 tag changed Preview behaviour. Previously, previewing force-loaded the loader (bypassed the user-agent check). Now, it doesn’t.

You *must* use the V8.1 tag when testing this PR. The V8.0 tag will load `loader.js` async under certain conditions (preview on Chrome 60, with a desktop UA),  and the loader will attempt to `document.write` a `<plaintext>` tag, which will fail because the DOM is already constructed. To use the new tag (if it hasn't been deployed to Merlins), use Charles with a Rewrite rule that replaces the existing tag (`<!-- MOBIFY.*END MOBIFY -->` as a regex) with the V8 tag from the file `web/mobify-tag/v8.min.html` in this PR.

 **JIRA**: [WEBDATA-214](https://mobify.atlassian.net/browse/WEBDATA-214) which has been recreated as [TAG-1](https://mobify.atlassian.net/browse/TAG-1)

## Changes
- In `web/app/loader.js`:
  - Add `IS_PREVIEW` and `IS_PREVIEW_PWA_MODE` flags
  - Rename `IS_LOCAL_PREVIEW` to `IS_LOADED_LOCALLY` to better reflect the meaning of the flag
  - In `isSupportedPWABrowser`, return true if running in preview PWA mode, bypassing the UA check
  - Rename `isSupportedNonPWAMessagingBrowser` to `isSupportedNonPWABrowser` - nonPWA mode is not Messaging-specific
  - In `isSupportedNonPWABrowser`, if service workers are supported and running in preview nonPWA mode, return true, bypassing the UA check
- Update the V8 tag source files to V8.1

## Todos
- [x] Add a high-level description of your changes to the CHANGELOG.md, including a link to the PR with your changes
- [ ] Update Merlins to the V8.1 tag *before merging and deploying this fix*

## How to test-drive this PR
- Run `npm run dev`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fwww.merlinspotions.com%2F&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1) and verify that the loader is force-loaded in PWA mode, irrespective of the UA
- Add `non-pwa=1` to the hash part of the page URL (the preview parameters) and verify that the loader is force-loaded in non-PWA mode, irrespective of the UA


jnurse:
I'm still seeing the PWA when I add `non-pwa=1` to the URL hash and I have a mobile user agent. Should the user agent override the hash setting? Or vice versa?

fractaltheory:
Force previewing the PWA with a non-mobile UA doesn't work properly with a V8.0 tag, meaning it can't be tested against the live Merlin's site. That said, I don't think any customer sites are on a V8.0 tag, so maybe we just need to get Merlin's updated first.

I'm also seeing the same behavior as @jnurse - mobile UA still sees PWA despite presence of `non-pwa=1` hash.

benlast:
@jnurse @fractaltheory I updated the PR instructions to explain that this needs the V8.1 tag in order to work properly in preview, with a quick Charles-style explanation of how to do that.

fractaltheory:
**V8.1 Tag**
_Previewing_
- desktop UA -> PWA
- mobile UA -> PWA

_Not previewing_
- desktop UA -> Non-PWA (well, the normal desktop site anyway)
- mobile UA -> PWA

As we discussed offline - there isn't really support for Non-PWA in a V7 tag context - and indeed, it doesn't work. However, previewing the PWA still works fine with a V7 tag.

Looks good! 👍

jnurse:
If I'm previewing the PWA with a non-mobile UA and I navigate to a page in the app (so my hash disappears) then I refresh the page, I see the desktop page.

PR #950, opened 2017-08-16
Resolve Merlin's inconsistent LoggedIn state

Cdok:
# Resolve Merlin's inconsistent LoggedIn state

Previously, Merlin's connector was parsing the desktop site's navigation to check if a user was loggedIn or not.

When navigating back to the home page on desktop, this navigation didn't always have the container we were looking for. This was causing some weird issues with the `loggedIn` state being flipped to `false` when the user was still logged in.

This PR switches the Merlin's connector to instead set and clear `localStorage` or `cookie`s depending on the user's logged in state. The `localStorage` and `cookie` keys match the respective desktop site storage keys. i.e. a user can log in on desktop, and they will still be logged in on mobile (we aren't setting a mobile-unique cookie).

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1493

### Landing on a page
 - Merlin's connector will check local storage _or_ cookies to see if the user is logged in
### Logging in
 - Local storage or cookies will be set on login
### Logging out
 - Local storage or cookies will be cleared on logout

We will no longer check the nav or parse desktop on each `fetchPageData` call to check for login status, as this was resulting in inconsistent states, even though the user was logged in.

## Changes
- Remove merlin's connector reliance on parsing navigation for `loggedIn` state
- Add `localStorage` and `cookie` for loggedIn states on Merlin's

## How to test-drive this PR
- Use Merlin's connector
- Run `npm start`
- Attempt logging in and out, in both Chrome & Incognito Safari & Incognito Chrome
- Navigate around these pages, ensuring that the loggedIn state is consistent


web/app/connectors/_merlins-connector/account/utils.js
```diff
@@ -105,3 +105,18 @@ export const updateCustomerAddresses = () => (dispatch) => {
...
+    const useLocalStorage = isLocalStorageAvailable()
+
+    if (useLocalStorage) {
+        const mageStorage = localStorage.getItem('mage-cache-storage')
+        const {customer} = mageStorage ? JSON.parse(mageStorage) : {}
+        return !!(customer && customer.fullname)
+    }
+
+    const mageCookie = getCookieValue('ls_mage-cache-storage')
+    const {customer} = mageCookie ? JSON.parse(decodeURIComponent(mageCookie)) : {}
```
jnurse: If you stick this in an else block rather than returning from the if block then you can have just one return statement and not repeat the logic to check the customer object. 

Normally I would say return early from the function, but since the logic to check if logged in is the same for both local storage and cookie storage it might be better to just have the one return statement.

web/app/connectors/_merlins-connector/account/commands.js
```diff
@@ -146,7 +144,7 @@ const clearMessageCookie = () => {
 const DEFAULT_ERROR_TEXT = 'Username or password is incorrect'
 const EXISTING_ACCT_REGEX = /already an account/
 
-const submitForm = (href, formValues, formSelector, responseUrl) => {
+const submitForm = (href, formValues, formSelector) => (dispatch) => {
```
jnurse: Does this still need a dispatch?

PR #957, opened 2017-08-21
Updated nightwatch-config.js to hide chrome info bars

jimthemighty:
Updated nightwatch-config.js to hide chrome info bars.

I added a couple of arguments to the hide the yellow info bar that appears at the top of chromedriver.


jasecode:
Hey @jimthemighty Unfortunately our tests won't run on forked branches and thus I'm unable to merge in this PR. I will however address it in an additional one with your name as credit 😃 . Thanks for the contribution!

PR #963, opened 2017-08-22
Web 1422  share

gsaray:
Add share to PDP

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1422

## Changes
- Add share to PDP

## Todos

## How to test-drive this PR
- Checkout `develop` in SDK 
- Run `npm link`
- Go to scaffold and checkout this branch
- Run `npm link progressive-web-sdk`
- Run `npm start`
- Go to a PDP 
- Click on share

## Note:
- right now, if you click on share options, it's going to submit the form. I have fixed the issue in 
https://github.com/mobify/progressive-web-sdk/pull/825 so we need a SDK release.


web/app/containers/product-details/partials/product-details-add-to-cart.jsx
```diff
@@ -89,8 +126,18 @@ const ProductDetailsAddToCart = ({
                     }}
                     data-analytics-name={UI_NAME.wishlist}
                 />
+                <Share
+                    className="u-flex"
+                    onShow={() => setOpenShare(true)}
+                    onDismiss={() => setOpenShare(false)}
```
gsaray: hmm no, I get errors like `cannot read property 'content' of undefined`. I think I have to create a "modal" component and include it in modal manager. what I have seems to be simpler than using modal manager. what do you think?

jnurse: My branch for the more menu has a fix for that `cannot read property 'content' of undefined` error so that you can use the regular modal actions (https://github.com/mobify/platform-scaffold/pull/973/files#diff-bace2c625a3e42d7ef8bae074c8664bfL70). 

You're right we probably don't want to use the modal manager for the share component right now, mainly because the share component in it's current form doesn't really work when you separate the modal from the trigger element. So getting it working with the modal manager would be a challenge.

PR #971, opened 2017-08-23
Generating unique string IDs for SFCC adding address

Cdok:
# Generating unique string IDs for SFCC adding address

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1511

## Changes
- Merlin's connector does not support "address names"
- For this reason, the form field was taken out of the UI
- Since the SFCC connector requires an address name, our UI is now generating an ID which it passes to the connector

## Todos

## How to test-drive this PR
- Try SFCC & Merlin's connectors
- Run `npm start`
- Login to mobifyqa@gmail.com
- Navigate to the account address page
- Attempt to add addresses on both connectors

### DIFF
--- a/web/app/containers/account-address/actions.js
+++ b/web/app/containers/account-address/actions.js
@@ -11,7 +11,7 @@ import {UI_NAME} from 'progressive-web-sdk/dist/analytics/data-objects/'
 export const setAddressID = createAction('Set Address ID', ['addressID'])
 export const setIsEditing = createAction('Set isEdit', ['isEdit'])
 
-export const submitAddAddress = (formValues) => (dispatch, getState) => {
+export const submitAddAddress = (formValues) => (dispatch) => {
     const {firstname, lastname} = splitFullName(formValues.name)
     // Merlin's connector doens't support address names,
     // and SFCC requires an address name.
--- a/web/app/containers/account-address/actions.js
+++ b/web/app/containers/account-address/actions.js
@@ -3,7 +3,6 @@
 /* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
 
 import {createAction} from 'progressive-web-sdk/dist/utils/action-creation'
-import Immutable from 'immutable'
 import {closeModal} from 'progressive-web-sdk/dist/store/modals/actions'
 import {splitFullName} from '../../utils/utils'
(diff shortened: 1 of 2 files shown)


web/app/containers/account-address/actions.js
```diff
@@ -11,10 +12,16 @@ import {UI_NAME} from 'progressive-web-sdk/dist/analytics/data-objects/'
 export const setAddressID = createAction('Set Address ID', ['addressID'])
 export const setIsEditing = createAction('Set isEdit', ['isEdit'])
 
-export const submitAddAddress = (formValues) => (dispatch) => {
+export const submitAddAddress = (formValues) => (dispatch, getState) => {
```
jnurse: Are you using getState here?

web/app/containers/account-address/actions.js
```diff
@@ -11,10 +12,16 @@ import {UI_NAME} from 'progressive-web-sdk/dist/analytics/data-objects/'
...
     const {firstname, lastname} = splitFullName(formValues.name)
+    // Merlin's connector doens't support address names,
+    // and SFCC requires an address name.
+    // Since we're not showing the addressName field, we need
+    // to manually assign an addressName for SFCC to accept the address.
+    // Merlin's connector will ignore this value.
+    const addressName = Math.random().toString(36).slice(2) // eslint-disable-line
```
jnurse: What happens if we're editing an address? With SFCC we'll have an addressName and it will be in the state right? Maybe we should only generate one if we don't have a name in our state?

web/app/containers/account-address/actions.js
Cdok: Ah good catch, I was using it before to check for collisions (getting the state, checking all the addresses against the generated ID).

web/app/containers/account-address/actions.js
Cdok: This is only being generated in the `submitAddAddress` action, with `submitEditAddress` the original generated `addressName` will be maintained, because it's being passed through the `formValues` param

jnurse: Won't we override it in line 24?

Cdok: We always want to override it in this action, since a separate action is being dispatched when we're editing - this one is always net-new addresses

image

^ edit address is always handing off from the form values

PR #976, opened 2017-08-23
Disable non-pwa feature

fractaltheory:
We don't yet want to release the non-pwa feature until the September release. Add a package.json controlled flag (injected at build-time via webpack), and use that flag in loader.js to disable it.

## Changes
- Add `nonPwaEnabled` boolean to `web/package.json`, with value `false`
- Update the webpack `base.loader.js` config to add a `NON_PWA_ENABLED` value at build-time
- Update `web/app/loader.js` with a local variable named `nonPwaEnabled`
- Use the local `nonPwaEnabled` variable to guard against the entire NonPwa code branch

## Todos
- [x] +1

## How to test-drive this PR
- Run `npm start`
- Using Charles, Map Remote `https://cdn.mobify.com/sites/progressive-web-scaffold/production/loader.js` to `https://localhost:8443/loader.js`
- View the desktop site without preview (and without mobile emulation)
- There should be no network requests for a `non-pwa.js` file, and you should see `a.js` requested by the loader.js file

### DIFF
--- a/web/app/loader.js
+++ b/web/app/loader.js
@@ -1,4 +1,4 @@
-/* global AJS_SLUG, NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, SITE_NAME, MESSAGING_ENABLED, DEBUG */
+/* global AJS_SLUG, NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, SITE_NAME, MESSAGING_ENABLED, DEBUG, NON_PWA_ENABLED */
 import {getAssetUrl, getBuildOrigin, loadAsset, initCacheManifest} from 'progressive-web-sdk/dist/asset-utils'
 import {
     documentWriteSupported,
@@ -30,6 +30,7 @@ import preloadJS from 'raw-loader!./preloader/preload.js' // eslint-disable-line
 
 const ASTRO_VERSION = NATIVE_WEBPACK_ASTRO_VERSION // replaced at build time
 const messagingEnabled = MESSAGING_ENABLED  // replaced at build time
+const nonPwaEnabled = NON_PWA_ENABLED // replaced at build time
 
 const CAPTURING_CDN = '//cdn.mobify.com/capturejs/capture-latest.min.js'
 const ASTRO_CLIENT_CDN = `//assets.mobify.com/astro/astro-client-${ASTRO_VERSION}.min.js`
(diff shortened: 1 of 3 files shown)


web/app/loader.js
```diff
@@ -1,4 +1,4 @@
-/* global AJS_SLUG, NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, SITE_NAME, MESSAGING_ENABLED, DEBUG */
+/* global AJS_SLUG, NATIVE_WEBPACK_ASTRO_VERSION, MESSAGING_SITE_ID, SITE_NAME, MESSAGING_ENABLED, DEBUG, NON_PWA_ENABLED */
```
benlast: @jeremywiebe is on a mission to get us to prefix these with `WEBPACK_` to make it easier to see where they come from - could we maybe make this `WEBPACK_NON_PWA_ENABLED`?

web/app/loader.js
```diff
@@ -643,7 +644,7 @@ if (shouldPreview()) {
     ) {
         loaderLog('Starting in PWA mode')
         loadPWA()
-    } else if (isSupportedNonPWABrowser()) {
+    } else if (NON_PWA_ENABLED && isSupportedNonPWABrowser()) {
```
benlast: I'm confused why you set `nonPwaEnabled` above (line 33) and then use the webpack-constant here?

web/package.json
```diff
@@ -6,6 +6,7 @@
   "mobifyGAID": "UA-78284797-2",
   "messagingSiteId": "merlinspotions",
   "messagingEnabled": true,
+  "nonPwaEnabled": true,
```
benlast: We want to set this to `false` for tomorrow's release, right?

web/app/loader.js
```diff
@@ -643,7 +644,7 @@ if (shouldPreview()) {
     ) {
         loaderLog('Starting in PWA mode')
         loadPWA()
-    } else if (isSupportedNonPWABrowser()) {
+    } else if (nonPwaEnabled && isSupportedNonPWABrowser()) {
```
benlast: You could omit the `nonPwaEnabled` value altogether and just use the webpack constant?

fractaltheory: Yeah it's a bit of a convention that was used for the other injected variables - so I went with it.

web/app/loader.js
fractaltheory: Fixed: https://github.com/mobify/platform-scaffold/pull/976/commits/0ec5b00c3f6075a9e83757f2532c20bb65789a34

web/app/loader.js
fractaltheory: I definitely just forgot to use the new variable - fixed here: https://github.com/mobify/platform-scaffold/pull/976/commits/8ce1fa7990bac89fcfcab93756441262c34b7edb

web/package.json
fractaltheory: Definitely `false` - fixed: https://github.com/mobify/platform-scaffold/pull/976/commits/551d6d648e76cff8933b825d901869113a950bbf

PR #998, opened 2017-09-06
Messaging: Use inline ask instead of default ask

fractaltheory:
Modifies Merlin's Potions footer newsletter section to use an Inline Ask messaging component, in addition to the existing e-mail signup form - with two Tab sections.

 **JIRA**: [MSG-517](https://mobify.atlassian.net/browse/MSG-517)
 **Linked PRs**: https://github.com/mobify/progressive-web-sdk/pull/865

## Changes
- Removes usage of Messaging `DefaultAsk` component from `web/app/containers/app/container.jsx`
- Update footer newsletter section copy and styling as per [MSG-517](https://mobify.atlassian.net/browse/MSG-517)
- Add usage of Messaging `InlineAsk` component

## Todos

## How to test-drive this PR
- Check out the #develop branch of the SDK, and run `npm run prod:build` then `npm link`
- `npm link progressive-web-sdk`
- Run `npm start`
- Scroll to footer, and note the new section


web/app/components/messaging-inline-ask/_base.scss
```diff
@@ -0,0 +1,24 @@
+// MessagingInlineAsk
+// ===
+// [1] - A translucent light-green
+
+.c-messaging-inline-ask__success {
+    background: rgba(0, 167, 116, 0.1); // [1]
```
benlast: Should this be a reference to a standard colour?

web/app/components/messaging-inline-ask/actions.js
```diff
@@ -0,0 +1,13 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+import {addNotification} from 'progressive-web-sdk/dist/store/notifications/actions'
+
+export const messagingNotSupportedNotification = () => (dispatch) => {
+    dispatch(addNotification(
+        'messagingNotSupported',
+        'Notifications are currently blocked for this site. Open "Settings" to allow notifications.',
```
benlast: The `"Settings"` reference here - is that to a particular browser menu or option? Is that enough to let a shopper know what to do? (I'm not suggesting we should be explicit about how to fix blocked notifications - that'd be a lot of text over all our supported browsers...)

web/app/components/messaging-inline-ask/index.jsx
```diff
@@ -0,0 +1,79 @@
...
+            // app notification letting them know it's currently blocked
+            this.props.notSupportedNotification()
+        }
+    }
+
+    render() {
+        return (
+            <div>
+                <p className="u-margin-bottom-md">
+                    Get notified on all the latest deals, promotions and new products.
```
benlast: Do the docs explain where the various strings are located, for a dev who needs to change them? Or should they maybe be defined as `const` values at the top of the file? Just a non-blocking pondering comment, really - I know we're not trying to build a super-configurable setup here :)

web/app/styles/themes/_pw-components.scss
```diff
@@ -64,8 +64,8 @@
 @import 'node_modules/progressive-web-sdk/dist/components/swatch/base';
 @import 'node_modules/progressive-web-sdk/dist/components/stepper/base';
 @import 'node_modules/progressive-web-sdk/dist/components/tabs/base';
-@import 'node_modules/progressive-web-sdk/dist/components/default-ask/base';
-@import 'node_modules/progressive-web-sdk/dist/components/default-ask/theme';
+// @import 'node_modules/progressive-web-sdk/dist/components/default-ask/base'; // disabled since we're using inline ask instead
```
benlast: Should these be deleted rather than commented out, or is the idea that a dev can see how she could use either approach?

web/app/components/messaging-inline-ask/_base.scss
```diff
@@ -0,0 +1,24 @@
+// MessagingInlineAsk
+// ===
+// [1] - A translucent light-green
```
jeffkamo: Minor/non-blocking: Technically this is Markdown, so there should be a space above this, otherwise the above heading might not be interpreted as a heading correctly.

That said, we don't actually compile these comments into any documentation, so it's not a huge deal.

Also... we tend to use `1.` for numbered lists format more so than the `[1]` parenthetical format.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
```diff
@@ -3,34 +3,79 @@
...
+            return (
+                <div className="t-footer__newsletter-loading">
+                    <InlineLoader />
+                </div>
+            )
+        }
 
-                <NewsletterForm onSubmit={onSubmit} />
+        // If we're unable to subscribe, render the newsletter section open by default
+        const tabIndex = this.props.canSubscribe ? 0 : 1
```
jeffkamo: After the user has subscribed to push notifications, should the tabs start to default to the newsletter subscription form instead?

web/app/containers/footer/_footer.scss
```diff
@@ -1,6 +1,6 @@
 // Footer
 // ===
-
+// [1] - To get that placeholder height _just_ right, at least on a fairly wide device
```
jeffkamo: This is sitting far from the line it comments. Instead, can this be positioned under the closest heading to the line it is relevant for? For example the `Panel Newsletter` heading

web/app/containers/footer/_footer.scss
```diff
@@ -39,3 +39,10 @@
         @include visually-hidden();
     }
 }
+
+.t-footer__newsletter-loading {
+    display: flex;
+    justify-content: center;
+
+    margin: 78px 0; // [1]
```
jeffkamo: We generally use the a plain `1` at line comment itself, and then a `1.` numbered list format for the comment explanation above.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
```diff
@@ -3,34 +3,79 @@
...
-                </h2>
+export class FooterNewsletterSubscription extends React.Component {
+    // In the case that we know Messaging is at least supported on the browser,
+    // we render a tabbed navigation section
+    renderTabs() {
+        // Render a loader while the Messaging Client is initialized
+        if (!this.props.isMessagingReady) {
+            return (
+                <div className="t-footer__newsletter-loading">
+                    <InlineLoader />
```
jeffkamo: Instead of hiding both the push notification tab _and_ the newsletter tab, would it make more sense to only hide the push notification tab? The user could theoretically still sign up for newsletters while push notifications are still waiting.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
```diff
@@ -3,34 +3,79 @@
...
+            </Tabs>
+        )
+    }
+
+    render() {
+        return (
+            <div className="t-footer__newsletter u-padding-top-lg u-padding-bottom-lg">
+                <div className="u-margin-bottom-md u-margin-start-md u-flexbox u-align-center">
+                    <Icon name="alert" title="Subscribe to Notifications" />
+                    <h2 className="u-h4 u-margin-start-md">
```
jeffkamo: I would advise against the use of margins inside of flexbox context. They may cause issues on older devices. Using padding instead.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
```diff
@@ -3,34 +3,79 @@
...
+                    <NewsletterForm onSubmit={this.props.onSubmit} />
+                </TabsPanel>
+            </Tabs>
+        )
+    }
+
+    render() {
+        return (
+            <div className="t-footer__newsletter u-padding-top-lg u-padding-bottom-lg">
+                <div className="u-margin-bottom-md u-margin-start-md u-flexbox u-align-center">
```
jeffkamo: I would recommend future proofing this a bit by using `u-padding-start-md u-padding-end-md` instead of `u-margin-start-md`. It's slightly more verbose, but it's more resilient to future changes.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
```diff
@@ -3,34 +3,79 @@
...
+                    <Icon name="alert" title="Subscribe to Notifications" />
```
jeffkamo: It may be worthwhile adding `u-flex-none` to this icon, to prevent any odd distortions if the heading text changes.

web/app/components/messaging-inline-ask/index.jsx
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+            // We allow users to click the opt in button, but we'll show an in-
+            // app notification letting them know it's currently blocked
+            this.props.notSupportedNotification()
+        }
+    }
+
+    render() {
+        return (
+            <div>
```
jeffkamo: Can this div include the root `c-messaging-inline-ask`? Even if it isn't used for styling, at least it will show which element is the component root in the markup.

web/app/components/messaging-inline-ask/actions.js
fractaltheory: This is the copy I received as part of the design. In this case, this would be shown to a compatible browser - essentially Chrome on Android. So, this is meant to inform Chrome for Android users.

web/app/styles/themes/_pw-components.scss
fractaltheory: I think that was my idea, but I think I might just update the existing docs to show how you would import these SCSS files if you wanted to use the built-in DefaultAsk component, and delete these as suggested.

web/app/components/messaging-inline-ask/_base.scss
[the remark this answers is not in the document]
fractaltheory: Did not know I could do this! Awesome.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
fractaltheory: We wanna show push notifications active by default, but in the case that a user has blocked permissions, we show e-mail by default. There are then a couple constraints on this:
- Messaging has to load before we know if they've blocked notifications
- We don't want a jarring, automatic switch between one tab and the next

Because of these constraints, we went with a loading spinner. But, I hear ya, not ideal.

web/app/containers/footer/partials/footer-newsletter-subscription.jsx
fractaltheory: Yep, that's a good catch - will change that.

web/app/components/messaging-inline-ask/index.jsx
fractaltheory: Ya, they would need to edit the component .jsx file directly. I've made a small change to accept properties now, though - so it should be clear: https://github.com/mobify/platform-scaffold/pull/998/commits/cdf1548b53226af9816d8cb0efe28fe0952eb4ce

fractaltheory:
@benlast @jeffkamo thanks very much for the thorough review. I've addressed all feedback, however one piece resulted in a knock-on bug, which I fixed via a cache variable (`tabIndex`), see https://github.com/mobify/platform-scaffold/pull/998/commits/3ff18378d36b8f0b4bf62799a7ec284b42ac88c8

Would appreciate your thoughts on whether this is a decent solution. I was considering storing that in state but that just seems like overkill.

fractaltheory:
@jeffkamo here's the diff after moving the component to the SDK: https://github.com/mobify/platform-scaffold/compare/3ff18378d36b8f0b4bf62799a7ec284b42ac88c8...9ccd6185416ee60c4a32a81b722c67f5f85c8ddb

jeffkamo:
Still looks pretty good to me 👍

PR #1002, opened 2017-09-07
Add better Debugging for End to End Tests.

jasecode:
Tired to go through from homepage to checkout manually? Have a script but always having to comment out ''browser.end" is getting very annoying?
Yes. Even for QA. It's hard to debug via SSH as well because of this.  

Lets add a test command that DOESN'T close the browser at the end, even if it fails.

These tests will go from Homepage - Before you Place an Order on checkout.
We also open a new chrome session every-time so that's a plus, we start fresh.

## Changes
- package.json: add test:e2e-debug which sets a `debug ENV variable`
- Refactored Opening and Closing steps of tests into functions
- Add step numbers to the name of tests and console.log the ENV variable for easier debugging

## How to test-drive this PR
- Run `npm run dev`
- npm run test:e2e-debug --tests tests/e2e/workflows/guest-checkout.js
- npm run test:e2e-debug --tests tests/e2e/workflows/registered-checkout.js

Note that you should CTRL-Force Quit those browers you've opened when the tests end as they won't close themselves.

### DIFF
--- a/web/tests/e2e/page-objects/product-list.js
+++ b/web/tests/e2e/page-objects/product-list.js
@@ -17,7 +17,7 @@ const ProductList = function(browser) {
 ProductList.prototype.navigateToProductDetails = function(productIndex) {
     // Navigate from ProductList to ProductDetails
     this.browser
-        .log(`Navigating to ProductList number: ${productIndex}`)
+        .log(`Navigating to ProductDetails number: ${productIndex}`)
         .waitForElementVisible(selectors.productDetailsItem(productIndex))
         .click(selectors.productDetailsItem(productIndex))
         .waitUntilMobified()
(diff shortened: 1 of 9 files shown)


web/tests/e2e/page-objects/product-list.js
```diff
@@ -17,7 +17,7 @@ const ProductList = function(browser) {
 ProductList.prototype.navigateToProductDetails = function(productIndex) {
     // Navigate from ProductList to ProductDetails
     this.browser
-        .log('Navigating to ProductDetails')
+        .log(`Navigating to ProductList number: ${productIndex}`)
```
ellenmobify: Navigating to Product number here.

web/tests/e2e/workflows/guest-checkout.js
```diff
@@ -34,24 +33,16 @@ export default {
     },
 
     after: (browser) => {
-        browser.end()
+        Home.prototype.closeBrowser(browser)
```
ellenmobify: Can this be `home.closeBrowser()`?

PR #1003, opened 2017-09-07
WEB-1533: merlins alt tag audit

jnurse:
This PR adds missing alt tags to images on the site. It resolves images included directly in the project and images parsed from Merlins. The SFCC audit will be in a separate PR. 

 **JIRA**: https://mobify.atlassian.net/browse/WEB-1533
 **Linked PRs**: n/a

## Changes
- Added alt tags to banners on the homepage
- Added alt tags to merlins PDP parsing
- Added alt tag to image in cart remove item modal
- Added alt tag to empty account order list image
- Added alt tag to image in account remove address modal
- Added alt tag to image in wishlist remove item modal

## Todos

## How to test-drive this PR
- Run `npm start`
- navigate through the site and check that no images are included without alt tags


web/app/containers/account-order-list/container.jsx
```diff
@@ -29,11 +30,13 @@ const NoOrder = ({dashboardURL}) => (
             </div>
         </div>
         <div className="u-padding-md u-margin-top-lg u-flexbox u-direction-column u-align-center u-justify-center">
-            <Icon
-                name="empty"
-                className="u-color-brand"
-                size="huge"
-            />
+            <Image
```
jeffkamo: Minor/non-blocking: Is there a reason for this switch to the image version of this icon? I think the icon is still used elsewhere, i.e. empty "address book" page.

jnurse: No I just noticed while doing the audit that it wasn't an image, and on similar pages we used an image not an icon. Do you have any thoughts on which is better here? An image or an icon?

jeffkamo: I suppose I would say the icon is preferable, because it is more flexible, one less network call, and can still be provided accessible text. Again, not a huge deal, so I'd say it's an optional fix.

jeffkamo:
Looks like all that's left is just a broken test:

```
FAIL  app/connectors/_merlins-connector/home/parser.test.js
```

PR #1012, opened 2017-09-12
[MBFYCNCT-71] Host Configuration

juanjcampos:
API configuration moved to registration. Host set as relative by default.

 **JIRA**: [MBFYCNCT-71](https://thinkwrap.jira.com/browse/MBFYCNCT-71)

## Changes
- With the current default config, the OCC services should be exposed on the same domain that the website. If not, the OCC host must be set under api.host (init-hybris-connector.js)


web/app/init-hybris-connector.js
```diff
@@ -4,6 +4,11 @@ import {registerConnector} from 'progressive-web-sdk/dist/integration-manager'
 const initConnector = () => {
     console.log('[IM.hybris-connector] Registering Hybris connector')
     registerConnector(Connector({
+        api: {
+            host: '',
+            type: 'rest',
```
jeremywiebe: @juanjcampos Do you foresee this needing to change per project? To me this sounds like a parameter that wouldn't change unless we modified the connector.

juanjcampos: We have clients that have customised both type and/or version with something different to 'rest' or 'v2'

jeremywiebe: Ok, that's good to know.  Thanks.

PR #1016, opened 2017-09-12
Tweaks to support performance investigations

benlast:
Some changes made while investigating rumoured performance issues

 **JIRA**: [MSG-533](https://mobify.atlassian.net/browse/MSG-533)

## Changes
- Enable compression when serving images from the webpack server and dev server (to eliminate it as a source of Lighthouse audit issues)
- Reduce the Chrome window size for headless tests (X11 rendering can be slow and a larger window means that the "spinner" animation runs slower and affects timings)
- Reduce the Chrome window size for production tests
- Upgrade lighthouse to 2.4.0 - this removes an issue where lighthouse waited for the main thread to become idle, which caused it to fail to see TTFI or TTFMP on a headless (xvfb) Chrome, which is what we use in Circle tests
- Fix the production lighthouse test by skipping the `no-javascript` audit (which appears to be broken on production, but isn't needed).

## How to test-drive this PR
- With the dev server running (`npm start`), run `npm run test:pwa-preview` to verify that Lighthouse tests work for preview
- Run `npm run test:pwa-prod` to verify that Lighthouse tests work for production

## Results of the investigation
See this [Google Drive folder](https://drive.google.com/open?id=0B-3ZUbkxQtoqVXFISGI2Uloyejg). The Tag Performance Investigations document summarizes the findings.


web/tests/performance/lighthouse/prod-lighthouse-config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
johnboxall: Why would we want to skip this in Prod?

benlast: Because it doesn't work. It's an audit that's supposed to load the page without Javascript and verify something, but it doesn't open the page URL correctly and always times out, causing the lighthouse test to fail and produce no output. Suppressing it still gives us valid performance reports.

PR #1022, opened 2017-09-14
[MBFYCNCT-74] checkout inline login delivery methods issue

juanjcampos:
On new checkout inline login, if user logged in has a default address, retrieve shipping methods for this address

The problem here was that if the user logs in on checkout step 1 (new inline login feature) and it has a default address, this address is automatically added to the cart, but the shipping methods are not retrieved automatically. This PR fixes this


web/app/containers/checkout-shipping/actions.js
```diff
@@ -67,10 +67,13 @@ export const submitSignIn = () => (dispatch, getState) => {
     } = getShippingFormValues(getState())
 
     return dispatch(login(username, password, 'on'))
-        .then(({isCartMerged}) => {
+        .then(({defaultAddress, isCartMerged}) => {
             if (isCartMerged) {
                 dispatch((openModal(CART_MERGE_ITEMS_MODAL)))
             }
+            if (defaultAddress && defaultAddress.id) {
+                dispatch(fetchShippingMethodsEstimate(defaultAddress))
```
kevinxh: I think it would be better if we dispatch `onSavedShippingAddressChange` to set the fetching flag , so the button is disabled while fetching shipping methods, like

```js
dispatch(onSavedShippingAddressChange(defaultAddress.id, defaultAddress))
```

juanjcampos: Done and tested. Works fine :)

PR #1031, opened 2017-09-15
re-enable the step to remove items from cart after a test is finished

ellenmobify:
It was disabled because it was super flaky. Maybe it's not so flaky now?


web/tests/e2e/page-objects/cart.js
```diff
@@ -33,21 +33,23 @@ Cart.prototype.removeItems = function() {
     // Remove all items from the cart
     const self = this
     this.browser
-        .preview()
+        .url(process.env.npm_package_siteUrl)
         .log('Opening mini cart')
         .waitForElementVisible(selectors.miniCart)
         .click(selectors.miniCart)
         .waitForElementVisible(selectors.miniCartContent)
+        .pause(2000)
```
jasecode: I know I'm guilty of doing this too but is there any element we can have an implicit wait for so we don't hard-code any pause values?

web/tests/e2e/page-objects/cart.js
```diff
@@ -33,21 +33,23 @@ Cart.prototype.removeItems = function() {
...
         .element('css selector', selectors.viewCart, (result) => {
             if (result.value && result.value.ELEMENT) {
                 self.browser
                     .log('View cart')
                     .waitForElementVisible(selectors.viewCart)
                     .click(selectors.viewCart)
                     .log('Removing item from cart')
-                    .execute(`document.querySelector('${selectors.removeItem}').click()`)
+                    .waitForElementVisible(selectors.removeItem)
```
jasecode: I'm planning to make it so we actually submit orders in the future for our checkout scripts now that we actually have control on the inventory of items. Is it possible to make a condition if item in cart exists?

web/tests/e2e/workflows/registered-checkout.js
```diff
@@ -33,7 +33,7 @@ export default {
     },
 
     after: () => {
-        // cart.removeItems()
+        cart.removeItems()
```
jasecode: So there seems to be no option here for '-debug' and if this fails it'll be really annoying to go through it to debug because the test won't stop at the point where it fials, it'll go back to the removeItems in the cart.

We should ignore 'removeItem' altogether if the 'debug' flag is passed.

web/tests/e2e/page-objects/cart.js
ellenmobify: This is not super explicit but `selectors.viewCart`, which selects the View and Edit Cart button in the mini cart, is only present if the cart has an item. For Merlin's Potions it's sufficient to check the presence of the View and Edit Cart button. 

For generated projects, is it possible for a user to view their cart if they have no items in it? If so, then we will need to check that a cart item exists before attempting to remove it.

web/tests/e2e/page-objects/cart.js
```diff
@@ -42,7 +42,7 @@ Cart.prototype.removeItems = function() {
             .waitForElementVisible(selectors.miniCart)
             .click(selectors.miniCart)
             .waitForElementVisible(selectors.miniCartContent)
-            .pause(1000)
+            .pause(2000)
```
ellenmobify: So this may be the flaky part - the `.pause()` is required. We cannot determine that the cart data has finished loading/updating. We cannot wait for the View and Edit Cart button to be visible because it may not exist.

PR #1034, opened 2017-09-18
Sets up template for error page from lancome build.

therafatm:
Sets up template for error page from lancome build.

 **JIRA**: https://mobify.atlassian.net/browse/AMP-81
**Status**: Ready for Review

## Changes
- Adds error container and styles, i.e. 
screen shot 2017-09-19 at 2 07 35 pm

- Adds analytics for error page
- Incorporates canonicalURL utility function to anywhere that uses a custom URL building function
- Bumps `mobify-amp-sdk` version to 0.7.2

## Todos
- [x] make it work
- [x] Get 👍 

## How to test-drive this PR
- `cd amp && npm run dev`
- visit `localhost:3000/randomroute` to see the static error page
- visit `localhost:3000/potions.html`  || `http://localhost:3000/potions/eye-of-newt.html`
to ensure PLP/PDP works


amp/app/constants.js
```diff
@@ -0,0 +1,9 @@
+const DEVELOPMENT_BASE_URL = 'https://merlinspotions.com'
```
kschmidtdev: We can runs the merlins docker locally, so this would be 'http://merlinspotions.local' - but we wouldn't want that to be the default. Will have to think about this some more 🤔

therafatm: Yeah I just put that in there for now because I didn't want it to break.

therafatm:
I think this is good to go other than figuring out what to do with `DEVELOPMENT_BASE_URL` in `/amp/app/constants.js`

kschmidtdev:
@therafatm I think it's fine to leave the solution as is, maybe with a comment about how to modify it if you want to run against your local server (will be a rare occurrance)

For some reason when I click links from the 404 the `mobify_id` parameter isn't being set right...but it's working for Lancome? I haven't figured it out yet.

amp/app/containers/error/container.jsx
```diff
@@ -0,0 +1,72 @@
...
+    >
+
+        {children}
+
+        <div id="app-wrap" className="t-app__wrapper u-flexbox u-direction-column">
+            <div id="app-header" className="u-flex-none" role="banner">
+                <header className="t-header">
+                    <div className="t-header__bar">
+                        <HeaderBar>
+                            <HeaderBarTitle href="/">
```
kschmidtdev: Please wrap this link with `canoncialUrl`

kschmidtdev:
Can you bump the version of `mobify-amp-sdk` to the [latest](https://github.com/mobify/mobify-amp-sdk/releases/tag/v0.7.2) in this PR as well? Thanks!

eamonnburke1:
@therafatm Two things.
1. The image is a little distorted. In the code, the image is set to display as a square, but the source image isn't quite square.
2. Can you delete this sentence 'You can click the button below to go back to the home page!' The button itself is obvious enough.

PR #1038, opened 2017-09-19
Improve session bridging and management

jeremywiebe:
This PR improves the SFCC connector session management and ensures that we default to bridging the desktop SFCC session using cookies. If that fails, we hit the desktop page and retry. 

## Changes
- Use the session bridging API to get a JWT token (by default)
- Provide connector configuration option to allow getting a guest/anonymous session that is not linked to a desktop session
- Check if token is nearing expiry (< 5 min) and refresh it (we can't refresh it once it's expired because the token is invalid at that point)

## Todos

## How to test-drive this PR
This is a bit involved, but we basically want to simulate a desktop session not existing, or being timed out, when we start. Because a hard-refresh would restore the desktop cookies, we need to use breakpoints and some judicious deleting of cookies to simulate this. 

- Run `npm start`
- Preview [Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- Delete the `mob-auth` session token (Dev Tools -> Application tab -> Session Storage)
- Select the **Sources** tab in Dev Tools
    - Set a breakpoint in `_sfcc-connector/session.js` on the line on the that calls `requestAuthToken()` 
        image
- Hard reload the page
- When you hit the breakpoint go into the Dev Tools and delete all cookies that start with `dw`. (This will be the piece that simulates an expired/non-existant desktop session.)
- Go back to the **Sources** tab and hit **Run**
- You will hit the breakpoint again, just hit **Run** again
- Switch to the **Network** tab
    - Verify that the call to `customers/auth` returned a 401 (due to missing `dw*` cookies)
    - Verify that a subsequent `GET` was made to `.../Home-Show`
    - Verify that a subsequent `GET` was made to `customers/auth` with a body:
        ```js
        {type: "session"}
        ```
       The response should contain an **Authorization** header starting with `Bearer ....`
       For example:
       image


web/app/connectors/_sfcc-connector/config.js
```diff
@@ -11,7 +11,15 @@ export const SEARCH_URL = '/catalogsearch/result/'
 let config = {}
 
 export const registerConfig = (cfg) => {
-    config = cfg
+    config = Object.assign(config, cfg)
```
kevinxh: What's the reason behind this?

web/app/connectors/_sfcc-connector/session.js
```diff
@@ -0,0 +1,175 @@
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+/* Copyright (c) 2017 Mobify Research & Development Inc. All rights reserved. */
+/* * *  *  * *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  *  * */
+
+import {makeRequest} from 'progressive-web-sdk/dist/utils/fetch-utils'
+
+import {getApiEndPoint, getRequestHeaders, getHomeURL, getShouldBridgeDesktopSession} from './config'
+import {getAuthToken, getAuthTokenPayload, storeAuthToken, deleteAuthToken, deleteBasketID} from './utils'
+
+const FIVE_MIN_IN_SECONDS = 5 * 60 // seconds
```
kevinxh: OCAPI session time is fixed at 5 mins?

web/app/connectors/_sfcc-connector/session.js
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Bridges an OCAPI (JWT token) session to desktop session.
+ * @warn  This is almost certainly not the function you want.
+ * @param {string} authorization The JWT token (in the same form it's returned
+ *                               to by `POST /customers/auth`. ie. 'Bearer aec9293229...'
+ *
+ * {@link https://documentation.demandware.com/DOC1/topic/com.demandware.dochelp/OCAPI/17.7/shop/Resources/Sessions.html#id-2104283703__id-514053870|Sessions}
```
kevinxh: Very nice and clear documentation! 💯

web/app/connectors/_sfcc-connector/config.js
jeremywiebe: It might not be needed, but I wanted to avoid changing the instance that `config` pointed to. I'm going to revert it for now as it's unrelated to the changes this PR represents.

web/app/connectors/_sfcc-connector/config.js
```diff
@@ -12,6 +12,14 @@ let config = {}
...
 export const registerConfig = (cfg) => {
     config = cfg
+
+    // Provide defaults for feature flags
+    config.features = config.features || {}
+
+    // By default we always want to bridge the desktop session.
+    // In non-browser contexts (AMP, for example) we may want to turn this
+    // feature off and use an unbridged ('guest' session).
+    config.features.shouldBridgeDesktopSession = config.features.shouldBridgeDesktopSession || true
```
kevinxh: I wonder if we should document existence of the available flags to a connector in the config file, because I assume when a SFCC project starts, engineers will fill the necessary information first in the config file. As the number of flags grow, they need to look at different places to set all the flags.

for example in `init-sfcc-connector.js`:

```js
const initConnector = () => {
    registerConnector(Connector({
        siteID: '2017refresh',
        clientID: '5640cc6b-f5e9-466e-9134-9853e9f9db93',
        /*
        features: {
            // the flag that indicate if the project wants to bridge the desktop session
            shouldBridgeDesktopSession: false
        }
        */
    }))
}
```

jeremywiebe: Very good idea! 💯  

I've opted for having a `DefaultConfig` object exported from the connector. This shows all the config options and their default values. Callers to `registerConfig()` would be able to pass in any part of that config object to override defaults.

kevinxh: 👍 I'm considering update the Hybris connector to also include a `DefaultConfig` object.

web/app/connectors/_sfcc-connector/cart/utils.js
```diff
@@ -2,36 +2,85 @@
...
 
-    return makeApiRequest('/baskets', options)
+    const customerData = getCustomerData(authToken)
+    return makeApiRequest(`/customers/${customerData.customer_id}/baskets`)
         .then((response) => response.json())
-        .then((basket) => {
-            storeBasketID(basket.basket_id)
-            return basket
+        .then(({baskets = []}) => {
+            console.log(`  Found ${baskets.length} baskets.`)
```
kevinxh: Just curious, why leave two spaces in front of the log? 🤔

jeremywiebe: Mostly to "nest" this message below the one that get's written earlier in the function. I was trying to give them a sense of connection.
```js
console.log(`Checking customer's existing baskets.`)
...
console.log(`  Found ${baskets.length} baskets.`)
```

Renders
```
Checking customer's existing baskets.
  Found 1 baskets.
```

( 🤔  the fact that the pluralization is broken is dumb... but I didn't think it worth handling for a developer-focused log message).

jeremywiebe:
@kevinxh The multiple calls to the `/auth` endpoint are due to an unrelated race condition which will be solved when we work on providing an "orderly startup" to the app so that some actions can be performed and guaranteed to end before the rest of the app boots up.  Right now we do something that needs the cart, which the needs an auth token, so an async network request is dispatched.  While that's pending, other code also accesses the cart, and since the first `/auth` request hasn't returned we still don't have an auth token and so a second `/auth` call is dispatched. At this point I haven't seen this cause an issue but it's something we need to fix yet.

PR #1040, opened 2017-09-20
preview build saves files

benlast:
By default, the `webpack-dev-server` serves files from memory, so they're not visible in the file system. This PR adds the `write-file-webpack-plugin` to save the packed files in the `build` directory. It also adds `Cache-Control: no-cache` to all served files to avoid source maps being cached in Chrome.

 **JIRA**: N/A

## Changes
- Add `write-file-webpack-plugin` and configure it in `web/dev-server/index.js`
- Add cache control in `web/dev-server/index.js

## Todos

## How to test-drive this PR
- Run `npm start`
- When the bundle is valid, check the `build` directory - the packed files, plus gzipped versions, plus source maps, should all be present
- Use `wget -O /dev/null --server-response --no-check-certificate --header 'accept-encoding: gzip' https://localhost:8443/loader.js.map` to verify that the response contains `Cache-Control: no-cache`
- Use `npm run test:pwa-preview` to verify that the lighthouse score is unaffected
- Use preview to verify that it works


web/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Add code to set `window.Progressive.Messaging.supported`. [#1004](https://github.com/mobify/platform-scaffold/pull/1004)
 - Minor changes to support Lighthouse testing [#1016](https://github.com/mobify/platform-scaffold/pull/1016) 
 - Use UglifyJS to remove dead and unused code from builds, and use GZIP compression in dev server [#1032](https://github.com/mobify/platform-scaffold/pull/1032)
+- Save the output from webpack builds [#1040](https://github.com/mobify/platform-scaffold/pull/1040)
```
jeremywiebe: "... from webpack _dev server_ builds"

benlast:
We're going to leave this change for now - doesn't appear that it's needed.

PR #1048, opened 2017-09-22
Add fallback fetch for service-worker-loader

jeffkamo:
In SFCC, downloading the `service-worker-loader` is only possible _without_ the `.js` extension. Thus, we attempt to first get the file with the extension, but if that fails, we retry without the extension.

## Changes
- If fetching `service-worker-loader.js` fails, then try fetching it without the `.js` extension

## Todos
- ~~(if applicable) analytics events instrumented to measure impact of change~~
- ~~Add a high-level description of your changes to the CHANGELOG.md, including a link to the PR with your changes~~

## How to test-drive this PR
- In `web/app/main.jsx`, comment out the Merlin's `initConnector` import, and uncomment the SFCC import. It should look like...
    ```jsx
    // import initConnector from './init-merlins-connector'
    import initConnector from './init-sfcc-connector'
    ```
- Run `npm start`
- Preview [SFCC Merlin's Potions](https://preview.mobify.com/?url=https%3A%2F%2Fmobify-tech-prtnr-na03-dw.demandware.net%2Fon%2Fdemandware.store%2FSites-2017refresh-Site%2Fdefault%2FHome-Show&site_folder=https%3A%2F%2Flocalhost%3A8443%2Floader.js&disabled=0&domain=&scope=1)
- After the page loads, verify that the `service-worker-loader` file downloaded correctly. **Note** that you may see `service-worker-loader.js` (the one with the extension) will have failed, but the network request without the extension should succeed.


web/app/loader.js
```diff
@@ -300,6 +300,17 @@ const loadWorker = (pwaMode) => {
     // the worker loader is served at the root of the site, so that
     // it controls the entire site.
     return navigator.serviceWorker.register(url)
+        .catch((e) => {
```
benlast: Given that this would have to happen on every load, should we record which was the successful URL (e.g. in localStorage) and try that one first next time?

fractaltheory:
Is there a way we can detect if we're in a SFCC environment and switch between two URLs that way?

Or perhaps, something like the same thing we do when switching between merlinsConnector and SFCC - have two lines and just comment out one or the other?

jeffkamo:
@fractaltheory Jeremy and I discussed how we might do just what you described. We decided that the "good" solutions required more effort than we thought would be worth it for this release.

For example, having lines that we comment in/out is not ideal. That's adding code that might become obsolete at generation time. Instead, we'd rather explore a mixture of generator options, webpack configs, and refactoring how the application "selects" which connector to init. But again, not exactly trivial effort.

web/app/loader.js
```diff
@@ -300,6 +300,17 @@ const loadWorker = (pwaMode) => {
...
+            // This catch block can be removed for Salesforce Commerce Cloud
+            // implementations. In which case, the service worker file name
+            // defined in SW_LOADER_PATH should be modified to _not_ have a
+            // `.js` extension in the file path.
+            if (e.message.indexOf('410') >= 0) {
```
fractaltheory: Could we include a comment here, indicating what 410 means, at least in the context of the error message we get back from the registration failure?

web/app/loader.js
```diff
@@ -293,7 +296,12 @@ const getServiceWorkerURL = (pwaMode) => {
  * false if the worker fails to register, load or become ready.
  */
 const loadWorker = (pwaMode) => {
-    const url = getServiceWorkerURL(pwaMode)
+    let url = window.localStorage.getItem(STORED_SERVICE_WORKER_LOADER_URL)
```
fractaltheory: We have a `isLocalStorageAvailable` method that you might want to guard behind:
`import {isLocalStorageAvailable, loadScriptAsPromise} from 'progressive-web-sdk/dist/utils/utils'`

fractaltheory:
👍  from me on the service worker changes!

